PR #44, opened 2025-09-17
fix api.call_service target typing to also allow lists of ids

zlangbert:
When invoking an action / calling a service the target map can either be a single id per key or a list of ids per key. You can see this in the docs [here](https://www.home-assistant.io/docs/scripts/perform-actions/#targeting-areas-and-devices).

Great project, really enjoying it so far, thanks!

### DIFF
--- a/src/hassette/core/api.py
+++ b/src/hassette/core/api.py
@@ -327,7 +327,7 @@ class Api(Resource):
         self,
         domain: str,
         service: str,
-        target: dict[str, str] | None = None,
+        target: dict[str, str] | dict[str, list[str]] | None = None,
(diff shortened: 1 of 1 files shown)


NodeJSmith (approved):

NodeJSmith:
@zlangbert  As soon as I figure out how to get actions to run against the PR changes and they are green, I'll merge this in. Thanks for the PR!

NodeJSmith (approved):

NodeJSmith:
@zlangbert I'll figure out the actions/rules issues on my own time, approved and merged for now, thanks again!

PR #197, opened 2025-11-24
Wait until hitting user code to cast states to specific model

### LINKED ISSUE #195: Dispatch events as raw data, only convert at user code boundary
We currently convert eagerly, emitting events with typed data. Now that we have the dependency injection to handle conversion at the boundary and the states proxy to hold states, it is making less sense to do this eagerly. We should store everything raw and then convert at the last moment.

This also allows users to define their own types and/or add missing types prior to them being added to Hassette, as they can be converted based on the DI annotations or states proxy annotations

NodeJSmith:
## Description

Use BaseState everywhere until we hit user code, convert to specific model at that point. Add `raw_data` to BaseState to hold the data to be used for conversion.

Fix state proxy resource by listening to websocket events as well as HA events, as HA events are not always reliable (due to websocket losing connection when HA restarts).

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code quality improvement (refactoring, type hints, etc.)
- [ ] Test coverage improvement

## Related Issues

Fixes #195


NodeJSmith:
@codex review

PR #214, opened 2026-01-29
Improve documentation + other big improvements

NodeJSmith:
### Added
- Add --version/-v argument to Hassette to allow displaying the current version
- Add `__iter__`, `__contains__`, `keys`, `values`, and `items` methods to StateManager and StateRegistry
- Add functionality to route `state_change` events to more specific handlers based on domain and/or entity_id
  - This is done automatically by the `Bus` by adding the entity_id to the topic when creating the listener
  - Matched listeners are deduplicated to ensure delivery only happens one time
  - Events are dispatched to the most specific route if there are multiple matches
- Add `AnnotationConverter` class and `TypeMatcher` class for more robust validation/conversion during DI
- Add A, P, C, and D aliases to `hassette.__init__` for simpler imports
  - `A` = `hassette.event_handling.accessors`
  - `P` = `hassette.event_handling.predicates`
  - `C` = `hassette.event_handling.conditions`
  - `D` = `hassette.event_handling.dependencies`
- Add new `Comparison` condition for basic operators (e.g. `==`, `!=`, `<`, `>`, etc.) to compare values in state/attribute change listeners
- Add new accessors for getting multiple/all attributes at once from state change events
  - `get_attrs_<old|new|old_new>` - specify a list of attrs
  - `get_all_attrs_<old|new|old_new>` - get all attributes as a dict

### Fixed
- Fix AppHandler reporting failed apps as successful by using status attribute
  - This is due to some issues with how we're tracking apps, further fixes will need to happen in future releases
- Fix StateManager using `BaseState` when we do not find a class in the `StateRegistry`
  - This does not work because `BaseState` doesn't have a `domain`
  - Error is now raised instead
- Log level is now used by Apps if set directly in AppConfig in Python code (as opposed to config file)
- Fix HassPayload's context attribute not being a HassContext instance
- `MediaPlayerState` now has `attributes` using the correct type

### Changed
- BREAKING: Replaced `StateManager.get_states` with `__getitem__` that accepts a state class
  - The error raised in StateManager when a state class is not found in the `StateRegistry` now advises to use this method
- Renamed `LOG_LEVELS` to `LOG_LEVEL_TYPE`
- Renamed `get_default_dict` to `get_defaults_dict` to be more clear this is not referring to `defaultdict` class
- Use same validation for `AppConfig` log level as we do for `Hassette` config log level
- Extracted nested Attributes classes for each state out of class definition to make them first class citizens
  - e.g. `MediaPlayerState.Attributes` is now `MediaPlayerAttributes`

### Docs
- Remove `Why Hassette` page
- Remove docker networking page
- Very large cleanup/reorg/addition of docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,62 +23,6 @@ A simple, modern, async-first Python framework for building Home Assistant autom
 - **Simple & Focused**: Just Home Assistant automations - no complexity creep
 - **Developer Experience**: Clear error messages, proper logging, hot-reloading, and intuitive APIs
 
-## 🚀 Quick Start
-
-Install Hassette:
-
-```bash
-pip install hassette
-```
-
-Create a simple app (`apps/hello.py`):
-
-```python
-from typing import Annotated
-
-from hassette import App
-from hassette import accessors as A
-from hassette import dependencies as D
-
-
-class HelloWorld(App):
-    async def on_initialize(self):
-        self.bus.on_state_change("binary_sensor.front_door", handler=self.on_door_open, changed_to="on")
-
-        self.scheduler.run_minutely(self.check_front_door)
-
-    async def on_door_open(
-        self,
-        entity_id: D.EntityId,
-        friendly_name: Annotated[str | None, A.get_attr_new("friendly_name")],
-    ):
-        self.logger.info("%s opened!", entity_id)
-        name = friendly_name or entity_id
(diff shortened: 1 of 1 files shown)


NodeJSmith:
TODO: replace bad gif on landing page with 1 - 3 better gifs

PR #1205, opened 2026-07-06
fix: entity turn_on/off/toggle use homeassistant domain

### LINKED ISSUE #1021: Fix base entity turn_on/turn_off/toggle dispatching nonexistent domain services
## Description

`BaseEntity.turn_on`/`turn_off`/`toggle` (async) and `BaseEntitySyncFacade`'s matching methods (sync) delegate via `api.turn_on(self.entity_id, self.domain)`, which calls `call_service(domain=<entity.domain>, service="turn_on")`. Passing the entity's own domain means a `lock` entity issues `lock.turn_on` — a service that does not exist. Only the generic `homeassistant.turn_on` works across domains.

Domains whose Home Assistant service registry includes `turn_on`/`turn_off`/`toggle` (light, switch, fan, …) get generated typed facade overrides that dispatch the correct domain service, so those work. Domains without those services (lock, cover, …) inherit the base methods, so `lock_entity.turn_on()` / `lock_entity.sync.turn_on()` fail at runtime with a service-not-found error.

Pre-existing on the async path; made more prominent by #1020, which exports 21+ typed domain facades. Surfaced during the #1020 adversarial review (finding F13 — its "uses deprecated `homeassistant.turn_on`" framing was inaccurate; the real dispatch is `{domain}.turn_on`).

## Design decision needed

Pick one and document it:

- **(a)** Base methods pass `domain="homeassistant"` (generic, cross-domain, deprecated in HA 2024.x but functional).
- **(b)** Don't expose `turn_on`/`turn_off`/`toggle` on entities whose domain lacks those services.
- **(c)** Other.

## Acceptance Criteria

- [ ] Calling `turn_on`/`turn_off`/`toggle` (async and `.sync`) on a serviceless-domain entity (e.g. `lock`) dispatches a service Home Assistant accepts — no service-not-found error.
- [ ] The chosen approach is documented in the affected method docstrings.
- [ ] `BaseEntitySyncFacade.turn_on`/`turn_off`/`toggle` carry explicit `-> None` annotations, consistent with the generated facade methods.
- [ ] A regression test exercises a serviceless-domain entity through these methods and asserts the dispatched `(domain, service)`.
- [ ] Unit + integration tests pass; pyright clean.

## Affected Areas

- `src/hassette/models/entities/base.py` — `BaseEntity.turn_on`/`turn_off`/`toggle` (async) and `BaseEntitySyncFacade.turn_on`/`turn_off`/`toggle` (sync): dispatch + `-> None` annotations.
- `src/hassette/api/api.py` — `Api.turn_on`/`turn_off`/`toggle_service`: dispatch reference; no change expected unless approach (a) is implemented at this layer.
- `tests/unit/test_sync_entity_facade.py` (or a new test) — serviceless-domain regression.

## Context

Pre-existing behavior; not introduced by #1020. Service-bearing domains are unaffected (they override these methods via codegen). The fix is localized but gated on the design decision above.

NodeJSmith:
### Bug Fixes

- **`turn_on`/`turn_off`/`toggle` failed on entities without domain-native services** — `BaseEntity` and `BaseEntitySyncFacade` passed the entity's own domain (e.g. `lock`, `cover`) to the underlying API calls. Domains without a `<domain>.turn_on`/`turn_off`/`toggle` service (like `lock`) raised service-not-found errors. Domains with generated overrides (`light`, `switch`) were unaffected since they don't go through this base path.
- The fix drops the explicit domain argument so the API's default (`homeassistant`) generic service is used instead, which every entity domain supports.
- Added regression tests using `LockEntity` as a representative serviceless domain, covering both the async and sync facade paths.

Closes #1021

### DIFF
--- a/src/hassette/models/entities/base.py
+++ b/src/hassette/models/entities/base.py
@@ -86,23 +86,23 @@ class BaseEntity(BaseModel, Generic[StateT, StateValueT]):
         """Turn off the entity."""
         # Shape B delegate — returns the callee's handle directly (no await, no second guard_await).
         # The single guard_await lives at api.call_service (the true primary). See design/071.
-        return self.api.turn_off(self.entity_id, self.domain)
+        return self.api.turn_off(self.entity_id)
 
     def turn_on(self, **data: Any) -> Coroutine[Any, Any, None]:
         """Turn on the entity.
 
         Args:
-            **data: Service data fields forwarded to the domain's ``turn_on`` service.
+            **data: Service data fields forwarded to the ``homeassistant.turn_on`` service.
         """
         # Shape B delegate — returns the callee's handle directly (no await, no second guard_await).
         # The single guard_await lives at api.call_service (the true primary). See design/071.
-        return self.api.turn_on(self.entity_id, self.domain, **data)
+        return self.api.turn_on(self.entity_id, **data)
 
     def toggle(self) -> Coroutine[Any, Any, None]:
         """Toggle the entity."""
         # Shape B delegate — returns the callee's handle directly (no await, no second guard_await).
         # The single guard_await lives at api.call_service (the true primary). See design/071.
-        return self.api.toggle_service(self.entity_id, self.domain)
+        return self.api.toggle_service(self.entity_id)
 
 
 class BaseEntitySyncFacade(Generic[StateT, StateValueT]):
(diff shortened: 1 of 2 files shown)


NodeJSmith:
@coderabbitai full review

PR #1364, opened 2026-07-17
fix(testing): Harden app testing helpers

### LINKED ISSUE #1332: Preserve concrete app types through AppTestHarness
## Description

Make `AppTestHarness` generic over the app class supplied to its constructor so `harness.app` retains the concrete app subtype.

The constructor currently accepts `type[App]`, stores `_app` as `App | None`, and exposes `app -> App`. An app author therefore loses editor and Pyright knowledge of app-specific methods and attributes after passing `MyApp` into the harness, despite the runtime object being `MyApp`.

```python
async with AppTestHarness(MyApp, config={}) as harness:
    harness.app.my_app_method()  # should be known as MyApp
```

This is a static typing improvement only; runtime behavior should not change.

## Acceptance Criteria

- [ ] Passing `type[MyApp]` infers an `AppTestHarness` whose app type is `MyApp`.
- [ ] `harness.app` is statically typed as `MyApp` without a cast.
- [ ] App-specific methods and properties are visible to Pyright and editors.
- [ ] The type parameter works for `App`, `AppSync`, and generic app-config subclasses.
- [ ] Simulation and time-control mixins retain correct host typing.
- [ ] Runtime behavior and constructor ergonomics remain unchanged.
- [ ] A static type-check fixture prevents regression.

## Affected Areas

- `src/hassette/test_utils/app_harness.py`
- `src/hassette/test_utils/simulation.py`
- `src/hassette/test_utils/time_control.py`
- Type-check fixtures under `tests/`

### LINKED ISSUE #1337: Default AppTestHarness config to an empty mapping
## Description

Make the `AppTestHarness` configuration argument optional for apps whose config has no required fields.

The constructor currently requires `config`, so tests and docs repeat `config={}` even when it carries no information. This is minor ceremony, but it appears throughout the public examples and integration suite.

## Acceptance Criteria

- [ ] `AppTestHarness(MyApp)` is equivalent to `AppTestHarness(MyApp, config={})`.
- [ ] The accepted type is a read-only mapping or `None`, normalized without a shared mutable default.
- [ ] Apps with required configuration still fail through `AppConfigurationError` with the same diagnostics.
- [ ] Type inference and staged setup from #1330 continue to work.
- [ ] Public examples omit the empty argument where doing so improves clarity.

## Affected Areas

- `src/hassette/test_utils/app_harness.py`
- `docs/pages/testing/`
- Public namespace after #1333

### LINKED ISSUE #1343: Validate AppTestHarness time and drain inputs strictly
## Steps to Reproduce

Pass NaN, infinity, or negative values to public time-control or drain timeout parameters, or make the canonical `date_utils.now` patch target unavailable.

## Expected Behavior

Invalid numerical boundaries fail immediately with parameter-specific errors, and `freeze_time()` never reports success unless its clock override was installed.

## Actual Behavior

NaN can make deadline comparisons permanently false, infinity can disable intended boundaries, and negative advances contradict scheduler ordering assumptions. `freeze_time()` suppresses `AttributeError` while installing its canonical patch, allowing a refactor to make a test continue against real time while appearing frozen.

## Version

Current `main` behavior in `time_control.py` and `simulation.py`.

## Acceptance Criteria

- [ ] NaN and infinite durations/timeouts are rejected at every public boundary.
- [ ] Negative drain timeouts are rejected.
- [ ] Forward-only helpers reject negative aggregate advances; deliberate clock rollback, if supported, has a separately named API.
- [ ] `freeze_time()` fails closed when its clock override cannot be installed.
- [ ] Partial patch setup rolls back and releases ownership cleanly.
- [ ] Errors identify the invalid parameter and value.
- [ ] Regression tests prove invalid values cannot hang the drain loop or silently use wall time.

### LINKED ISSUE #1354: Reject invalid test API assertions and improve mismatch diagnostics
## Steps to Reproduce

Call `assert_not_called("trun_on")` or `assert_call_count("trun_on", 0)` with a misspelled method name. Both can pass because no such calls exist. For a real method with many recorded calls, provide mismatched kwargs and inspect the unstructured dump.

## Expected Behavior

Malformed assertions fail closed, and valid assertion failures explain the closest relevant differences.

## Actual Behavior

Negative and zero-count assertions accept arbitrary method strings, so typos produce false-green tests. Positive mismatch output dumps all same-method calls without identifying missing or differing fields.

## Version

Current `main` test API assertion behavior.

## Acceptance Criteria

- [ ] Assertion helpers validate method names against one canonical recordable-method contract.
- [ ] Misspellings fail with close-match suggestions.
- [ ] Negative call counts are rejected.
- [ ] Zero-count assertions remain supported for valid methods.
- [ ] Expected keyword names are validated where the production signature is closed.
- [ ] Mismatch diagnostics identify missing and differing keys for relevant same-method calls.
- [ ] Partial versus exact matching semantics remain explicit.
- [ ] Async and sync method names share the same validation source.
- [ ] Regression tests prove misspelled negative assertions cannot pass.

## Non-Goals

- A fluent assertion DSL
- A general query language

NodeJSmith:
### Harness ergonomics and typing
- Preserve the concrete app subtype through `AppTestHarness` so editors and Pyright expose app-specific members without casts while keeping runtime behavior unchanged.
- Accept an optional read-only config mapping and normalize it internally, removing repeated `config={}` ceremony for apps with no required settings without introducing a shared mutable default.

### Time and simulation boundaries
- Reject non-finite and negative simulation timeouts before event dispatch so invalid deadlines cannot silently disable drain bounds or hang tests.
- Keep clock advancement forward-only and fail closed when a time patch cannot be installed, including rolling back partial setup and releasing only the lock owned by that harness.

### Recording assertions
- Generate one canonical set of recordable API methods and use it for both async and sync recording paths, preventing assertion behavior from drifting between facades.
- Reject misspelled method names, invalid closed-signature keywords, and negative call counts so malformed negative assertions cannot produce false-green tests.
- Report the closest recorded call with missing, mismatched, and exact-match-only extra values, making failures actionable without changing partial-versus-exact matching semantics.

### Verification
- `prek -a`
- `uv run nox -s dev` (7,946 passed, 1 skipped, 2 xfailed)
- Focused generator and test-utils suite (196 passed)
- Code, integration, readability, and clean-code gates

### Follow-up
- Further structural decomposition of the test orchestration paths remains tracked in #1363 and is intentionally outside this quick-win PR.

Closes #1332

Closes #1337

Closes #1343

Closes #1354

### DIFF
--- a/codegen/src/hassette_codegen/sync_facade/recording.py
+++ b/codegen/src/hassette_codegen/sync_facade/recording.py
@@ -81,6 +81,12 @@ from typing import Any
 if typing.TYPE_CHECKING:
     from hassette.test_utils.recording_api import RecordingApi
 {type_checking_imports}\
+RECORDED_API_METHODS = frozenset(
+    {{
+{recorded_api_methods}
+    }}
+)
+
 # Stub message templates — imported by tests to avoid brittle substring matches.
 # Must stay byte-identical with the constants in codegen/src/hassette_codegen/sync_facade/.
 STUB_MSG_STATE_CONVERSION = (
@@ -94,6 +100,16 @@ STUB_MSG_GENERIC = (
 )
 
 '''
+
+_RECORDING_HELPERS = frozenset({"_record_call", "_create_helper", "_update_helper", "_delete_helper"})
+
+
+def _records_api_call(func: ast.AsyncFunctionDef | ast.FunctionDef) -> bool:
+    """Return whether a method records through the canonical helper path."""
+    return any(
+        isinstance(node, ast.Call) and isinstance(node.func, ast.Attribute) and node.func.attr in _RECORDING_HELPERS
+        for node in ast.walk(func)
+    )
 
 
 def _find_class(module: ast.Module, name: str, source_label: str) -> ast.ClassDef:
@@ -169,6 +185,7 @@ def generate_sync_recording(api_path: Path, recording_api_path: Path) -> str:
 
     # Generate each method; track body nodes for import derivation.
     generated_methods: list[str] = []
+    recorded_api_methods: set[str] = set()
     all_body_nodes: list[ast.stmt] = []
     all_runtime_annotation_symbols: set[str] = set()
     all_string_ref_symbols: set[str] = set()
@@ -183,6 +200,8 @@ def generate_sync_recording(api_path: Path, recording_api_path: Path) -> str:
 
         if use_body_copy:
             assert rec_func is not None
+            if _records_api_call(rec_func):
+                recorded_api_methods.add(method_name)
             # Collect annotation symbols from the recording_api signature
             runtime_syms, string_syms = collect_annotation_symbols(rec_func)
             all_runtime_annotation_symbols |= runtime_syms
@@ -250,6 +269,7 @@ def generate_sync_recording(api_path: Path, recording_api_path: Path) -> str:
         _RECORDING_MODULE_HEADER_TEMPLATE.format(
             imports=import_block,
             type_checking_imports=type_checking_imports,
+            recorded_api_methods="\n".join(f'        "{name}",' for name in sorted(recorded_api_methods)),
         )
         + _RECORDING_CLASS_HEADER
         + "\n"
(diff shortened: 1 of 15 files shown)


NodeJSmith:
@coderabbitai full review

NodeJSmith:
Added focused regression coverage for all timeout-validating simulation entry points, partial freeze-time patcher cleanup, and valid closed-signature assertion keywords.

PR #1395, opened 2026-07-21
refactor(frontend): centralize route metadata and path builders

### LINKED ISSUE #1386: Centralize frontend route metadata and path builders
## Description
Routes, sidebar entries, command-palette pages, app links, and deep handler/execution links independently construct paths. The `/design` route already exists in the router but not in navigation/palette, showing drift.

## Acceptance Criteria
- Top-level route/page metadata is defined in one place.
- App, instance, handler, execution, and log URLs are built through shared helpers.
- Sidebar and command palette consume the shared metadata where appropriate.
- Tests cover at least one deep execution route and query preservation.

## Source
Audit: `design/audits/2026-07-19-frontend-quality-audit/audit.md` M7.

NodeJSmith:
### Route Registry (`app-routes.ts`)

- Added `NAV_PAGES` — a single metadata array (`path`, `label`, `testId`) that now backs both the sidebar nav list and the command palette's static page items. Previously these were two separately hand-maintained arrays (`NAV_ITEMS` in `sidebar.tsx`, an inline array in `palette-items.ts`) that had to be kept in sync by hand.
- Added `HOME_PATH`, `handlerPath()`, `executionPath()` builders, and moved `logEntryExecutionHref()` (previously `executionDetailHref()` + `logEntryExecutionHref()` in `format.ts`) into `app-routes.ts` so all route construction lives in one file instead of being split across `format.ts` and ad hoc template strings.
- Renamed the internal query type from `AppRouteQuery` to `RouteQuery` and dropped the now-unneeded `appendQuery()` helper — `buildQuery()` already returns an empty string when there's nothing to append, so string concatenation is simpler than a conditional wrapper.
- Added `app-routes.test.ts` covering every builder, including the `handler`→`listener` execution-kind-to-URL-segment mapping that was previously undocumented and untested.

### Typed handler references replace `execLinkPrefix`

- `ExecutionTable`, `HandlerDetailLayout`, and `AppLink` took a pre-built path-prefix string (`execLinkPrefix`, or a bare `handlerId` string) and concatenated route segments onto it at the call site. Replaced with typed `appKey` / `handlerKind` / `handlerId` props, so the component itself builds the URL via `executionPath()`/`handlerPath()` instead of every caller re-deriving the prefix string.
- Added `handlerId` to `UnifiedRow` (`handler-rows.ts`) so decomposed handler rows carry a real ID instead of only the string segment they used for navigation before.

### Migrated call sites

- Sidebar, palette items, error spotlight, execution/job/listener detail pages, handlers tab, overview tab helpers, log table row/drawer, and `not-found.tsx` now call the shared builders instead of hardcoding `/apps/...`, `/handlers/...` template strings. ~86 hardcoded route occurrences across 26 files.

Closes #1386

### DIFF
--- a/frontend/src/app.tsx
+++ b/frontend/src/app.tsx
@@ -22,6 +22,7 @@ import { LogsPage } from "./pages/logs";
 import { NotFoundPage } from "./pages/not-found";
 import { AppStateContext } from "./state/context";
 import { createAppState, RELATIVE_TIME_TICK_MS } from "./state/create-app-state";
+import { HOME_PATH } from "./utils/app-routes";
 
 export function App() {
   const queryClient = useMemo(() => createQueryClient(), []);
@@ -119,7 +120,7 @@ export function App() {
               <ErrorBoundary resetKey={location}>
                 <Switch>
                   <Route path="/">
-                    <Redirect to="/apps" />
+                    <Redirect to={HOME_PATH} />
                   </Route>
                   {(["listener", "job"] as const).map((kind) => [
                     <Route key={`${kind}-exec`} path={`/apps/:key/handlers/${kind}/:id/exec/:execId`}>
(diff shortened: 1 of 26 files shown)


NodeJSmith:
@coderabbitai full review

PR #1437, opened 2026-07-26
feat: add deterministic DB seed script for dashboard QA

NodeJSmith:
### Seed script

- Add `scripts/seed_db.py`, a deterministic SQLite telemetry seeder covering 7 named scenarios (`healthy`, `empty`, `degraded`, `error`, `large-volume`, `lifecycle`, `adversarial`). Lets frontend QA, CLI doc generation, visual regression screenshots, and demos put the monitoring dashboard into a specific state without waiting on a live Home Assistant instance or a slow, non-deterministic demo-app run.
- All scenario timestamps are fixed offsets from a hardcoded reference instant rather than wall-clock reads, so running the script twice for the same scenario produces byte-identical database content.
- The seeder applies the real migration runner and imports the production INSERT-param builders (see below) instead of hand-rolling SQL, so schema drift between the seeder and the real write path is caught at generation time.
- See `design/specs/017-seed-db/design.md` for the full design and `design/research/2026-07-25-deterministic-db-seeding/research.md` for the feasibility research behind it.

### Publicize shared INSERT-param builders

- Rename `_execution_insert_params` → `execution_insert_params` and `_listener_insert_params` → `listener_insert_params` in `telemetry/repository.py`, and extract a new `job_insert_params()` out of the inline dict that `register_job()` used to build by hand. The seed script needs the exact same param-building logic the real write path uses — publicizing these instead of duplicating them means the seeder and the production path can never drift apart on column names or value coercion.
- Rename `database_service._LOG_COLUMNS` → `LOG_RECORD_COLUMNS` for the same reason on the log-records insert path.

### Test infrastructure

- Add `make_execution_record()` to the shared test factory registry (`src/hassette/test_utils/factories.py`) and register it with `tools/check_test_factories.py` so future local reinventions get flagged.
- Migrate `tests/unit/core/test_command_executor_pipeline.py`'s local `make_execution_record()` duplicate onto the shared factory.
- Fix `check_test_factories.py`'s AST visitor: it was flagging class methods and closures nested inside another function as top-level factory shadows. Only `def`s reachable from module scope (including ones nested inside `if`/`try`/`with`) count as shadows now — a method or closure sharing a factory's name isn't importable as a drop-in replacement, so it carries none of the duplication risk the guard exists to catch.
- Add schema-drift round-trip tests for the hand-typed `sessions` and `blocking_events` insert columns, matching the existing pattern already in place for the other four telemetry tables.
- Add `tests/integration/test_seed_db.py` and `tests/unit/core/test_param_builders.py` covering all 7 scenario generators and the param builders directly.

### Notable Changes

- `main()` in `seed_db.py` now gives a clear error message when two runs target the same `--output` path concurrently, instead of leaking a raw SQLite traceback.

### Housekeeping

- `CLAUDE.md`: document the seed script usage and scenario list under Common Commands.
- Delete the now-obsolete `design/specs/017-seed-db/tasks/` directory (design/plan artifacts archived after implementation landed).


NodeJSmith:
@coderabbitai review

PR #1466, opened 2026-07-27
refactor: swap frontend from Preact to React 19, Zustand, Tailwind v4

NodeJSmith:
### Frontend foundation migration (spec 020)

Migrates the web dashboard frontend from Preact + `@preact/signals` + CSS Modules to React 19 + Zustand + Tailwind CSS v4, and initializes shadcn/ui as the component foundation for spec 2 (component-by-component replacement, not part of this PR). See `design/specs/020-frontend-foundation-migration/design.md` for the full design rationale, including the four-phase sequencing (state → framework → styling → test infra) and why Zustand had to land before the React swap (signals can't coexist with the React renderer).

The migration is a pinned refactor — existing component behavior is unchanged, verified by the existing test suite plus the new `typecheck` CI gate. No end-user-facing API surface changed.

### Review fixes (post-migration)

An adversarial review (3 critics: Web Platform, Skeptical Senior Engineer, Data Integrity) of the initial migration surfaced 6 findings, all resolved:

- `handleWsConnected` now performs one atomic `set()` call instead of three, avoiding a torn-state render if a future edit inserts an `await` between the connection write and the reconnect-only clears.
- shadcn's dark-mode CSS selector changed from `.dark` to `[data-theme="dark"]` to match how this app actually toggles themes — the `.dark` block was dead code that would have silently broken dark mode for spec 2's first shadcn component.
- Theme/sidebar-collapsed/time-preset side effects (DOM attribute write, `localStorage` persistence) are now owned by the corresponding Zustand store actions instead of being duplicated at every call site.
- Added a `typecheck` script (`tsc --noEmit`) and a CI step for it — the migration rewrote the type surface of nearly every frontend file with no type-check gate anywhere in CI before this.
- Test `afterEach` now clears `localStorage` alongside the existing store reset, so preference-field defaults can't leak between tests in the same file.
- Filed #1465 to track fleet-wide re-renders on `AppDetailPage` from coarse-grained store selectors — inherited from the pre-migration architecture, not a regression from this branch, and out of scope for this fix pass.

A follow-up clean-code pass deduplicated the Shiki bootstrap, focus-trap logic, icon wrappers, error-display branches, and health-kind conditional; removed dead code (empty CSS rules, a stale lint suppression); and wired `clearServiceStatus`/`clearLogs` into the reconnect path in `store.ts`.

### Housekeeping

- Documented the Tailwind/shadcn coexistence model in `CLAUDE.md` (CSS Architecture section) and fixed a stray Preact JSX `class=` attribute in the same doc's example.


NodeJSmith:
@coderabbitai full review

NodeJSmith:
@coderabbitai full review

PR #1471, opened 2026-07-28
feat: add EntityTime trigger for schedules driven by an entity's time

### LINKED ISSUE #969: Add entity-driven time trigger for dynamic schedule from HA entity state
## Description

Add a trigger type that fires based on a time value stored in an HA entity's state or attributes, automatically rescheduling when the entity changes.

Use case: "run a handler when my phone's alarm goes off" — the alarm time lives in `sensor.phone_next_alarm` and changes daily. Or "trigger 30 minutes before a user-configured time" stored in an `input_datetime` helper.

Currently users would need to combine a state change listener (to detect when the entity updates) with a scheduler job (to fire at the right time), manually cancelling and rescheduling whenever the entity changes. This is a pattern that deserves first-class support.

## Prior Art

Node-RED's Time node triggers based on an HA entity's time attribute. It supports:
- Entity ID selection (any entity with a time-like state or attribute)
- Offset (positive or negative) from the entity's time value
- Random offset within a range
- Daily repeat mode (extract just the time component, fire daily at that time)
- Automatic rescheduling when the entity state changes

## Proposed API

```python
# Fire when the alarm goes off
await self.scheduler.schedule(
    self.on_alarm,
    EntityTime(
        entity_id="sensor.phone_next_alarm",
        offset=TimeDelta(minutes=0),
    ),
)

# Fire 30 minutes before a user-configured time, re-evaluated daily
await self.scheduler.schedule(
    self.pre_routine,
    EntityTime(
        entity_id="input_datetime.morning_routine",
        offset=TimeDelta(minutes=-30),
        daily=True,  # extract time component, repeat daily
    ),
)
```

## Design Considerations

- **State parsing**: The entity's state or attribute needs to be parsed as a time/datetime. `input_datetime` entities store ISO format strings. Phone alarm sensors vary by integration. Should support configurable attribute selection.
- **Rescheduling**: When the entity state changes, the pending trigger must be cancelled and rescheduled. This is a bus + scheduler coordination pattern.
- **Daily mode**: When `daily=True`, only the time component matters — the trigger fires daily at that time, rescheduling if the entity changes.
- **Missing/unavailable state**: Handle cases where the entity is unavailable or has no valid time value (e.g., no alarm set).
- **Relationship to #966**: Solar triggers (#966) are a specific case of time-from-entity (the sun entity). An `EntityTime` trigger could potentially subsume sunrise/sunset triggers if designed generically enough, though a dedicated `Sun` trigger with named events (`"sunrise"`, `"sunset"`) is more ergonomic for that common case.

## Acceptance Criteria

- [ ] New trigger type that reads a time value from an HA entity
- [ ] Automatic rescheduling when the entity state changes
- [ ] Configurable offset (positive or negative)
- [ ] Daily repeat mode (extract time component, fire daily)
- [ ] Graceful handling of unavailable/invalid entity state
- [ ] Unit tests including entity change mid-schedule and daily rollover

NodeJSmith:
## Summary

Adds `EntityTime` to `hassette.scheduler` — a trigger that reads its fire time from a Home Assistant entity and moves the job whenever that entity changes.

This covers the "run when my phone alarm goes off" and "fire 30 minutes before a user-configured time" cases. Today those require hand-wiring a state listener to a scheduler job and manually cancelling/rescheduling on every entity change; this makes the pattern first-class.

```python
# Fire when the alarm goes off
await self.scheduler.schedule(self.on_alarm, EntityTime("sensor.phone_next_alarm"))

# Fire 30 minutes before a user-configured time, every day
await self.scheduler.schedule(
    self.pre_routine,
    EntityTime("input_datetime.morning_routine", offset=TimeDelta(minutes=-30), daily=True),
)

# Read from an attribute rather than the state
await self.scheduler.schedule(self.at_dawn, EntityTime("sun.sun", attribute="next_dawn"))
```

## What changed

**`src/hassette/scheduler/triggers.py`** — the `EntityTime` trigger.

- `entity_id`, plus `attribute=` to read a time out of an attribute instead of the state, `offset=` (a `whenever.TimeDelta`, negative fires early), and `daily=` to keep only the time of day and repeat.
- `daily=True` goes through the same cron machinery `Daily` uses, so the repeat is DST-safe.
- Parses the value shapes HA actually emits for times: offset-aware ISO strings, naive ISO date-times, time-only strings, and unix timestamps.
- An entity with no usable time (unavailable, `unknown`, no alarm set) parks the job at `NO_OCCURRENCE` rather than removing it, so the job stays registered and resumes on the entity's next change.
- An absolute time at or before now also resolves to no occurrence — that keeps a restart from re-firing an alarm that already went off.
- The trigger resolves through a bound state reader; unbound (used outside the scheduler) it resolves to no occurrence instead of raising.

**`src/hassette/scheduler/scheduler.py`** — `schedule()` registers a framework-bus state listener per entity-driven job, tracked in `_entity_time_subs` and cancelled along with the job. The listener reads the new state off the event rather than from the `StateProxy` cache, whose update is dispatched concurrently and so may not be visible yet.

**`src/hassette/core/scheduler_service.py`** — `reschedule_job()` moves an already-queued job to a new time without deregistering it. It no-ops for a job that isn't on the heap (cancelled, or mid-dispatch — in that case `dispatch_and_log` consults the trigger itself).

**Docs** — `EntityTime` section in `docs/pages/core-concepts/scheduler/triggers.md` plus a runnable snippet at `docs/pages/core-concepts/scheduler/snippets/scheduler_entity_time.py`.

## Testing

- `tests/unit/scheduler/test_entity_time_trigger.py` (new) — resolution across the supported value shapes, offset and `daily=` behavior, past/boundary times, unparsable and unavailable states, unbound triggers, `entity_id` validation.
- `tests/integration/test_scheduler_entity_time.py` (new) — end-to-end through a real scheduler: the job moves when the entity changes, parks and recovers across an unavailable state, and the listener is torn down when the job is cancelled.
- `uv run nox -s dev` — 8402 passed, 1 skipped, 2 xfailed.
- `prek -a` and `prek run -a --hook-stage pre-push` (ruff, pyright, schema freshness, all repo linters) — clean.

No frontend files touched.

Closes #969

### DIFF
--- a/docs/pages/core-concepts/scheduler/snippets/scheduler_entity_time.py
+++ b/docs/pages/core-concepts/scheduler/snippets/scheduler_entity_time.py
@@ -1,0 +1,41 @@
+from whenever import TimeDelta
+
+from hassette import App, AppConfig
+from hassette.scheduler import EntityTime
+
+
+class WakeUpApp(App[AppConfig]):
+    async def on_initialize(self) -> None:
+        # --8<-- [start:alarm]
+        await self.scheduler.schedule(
+            self.start_the_day,
+            EntityTime("sensor.phone_next_alarm"),
+            name="phone_alarm",
+        )
+        # --8<-- [end:alarm]
+
+        # --8<-- [start:offset_daily]
+        await self.scheduler.schedule(
+            self.preheat_coffee,
+            EntityTime(
+                "input_datetime.morning_routine",
+                offset=TimeDelta(minutes=-30),
+                daily=True,
+            ),
+            name="preheat_coffee",
+        )
+        # --8<-- [end:offset_daily]
+
+        # --8<-- [start:attribute]
+        await self.scheduler.schedule(
+            self.open_blinds,
+            EntityTime("sun.sun", attribute="next_dawn"),
+            name="open_blinds_at_dawn",
+        )
+        # --8<-- [end:attribute]
+
+    async def start_the_day(self) -> None: ...
+
+    async def preheat_coffee(self) -> None: ...
+
+    async def open_blinds(self) -> None: ...
--- a/docs/pages/core-concepts/scheduler/snippets/scheduler_trigger_imports.py
+++ b/docs/pages/core-concepts/scheduler/snippets/scheduler_trigger_imports.py
@@ -1,0 +1,1 @@
+from hassette.scheduler import After, Once, Every, Daily, Cron, EntityTime, TriggerProtocol
--- a/docs/pages/core-concepts/scheduler/triggers.md
+++ b/docs/pages/core-concepts/scheduler/triggers.md
@@ -2,10 +2,10 @@
 
 A trigger determines when a scheduled job fires. Each built-in scheduling method creates a trigger internally. `schedule()` accepts a trigger directly for patterns the convenience methods do not cover.
 
-All five built-in trigger types and [`TriggerProtocol`][hassette.types.types.TriggerProtocol] are importable from `hassette.scheduler`:
+All six built-in trigger types and [`TriggerProtocol`][hassette.types.types.TriggerProtocol] are importable from `hassette.scheduler`:
 
 ```python
-from hassette.scheduler import After, Once, Every, Daily, Cron, TriggerProtocol
+from hassette.scheduler import After, Once, Every, Daily, Cron, EntityTime, TriggerProtocol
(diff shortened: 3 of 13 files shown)


NodeJSmith:
@coderabbitai review

Issue #72, opened 2025-09-26
Add backpressure and queue overflow handling for event bus


NodeJSmith:
## Description

The event bus currently has no mechanism to handle situations where events are produced faster than handlers can consume them. This can lead to unbounded memory growth and cascading delays.

## Proposed Solution

- Add a configurable max queue depth per listener
- When the queue is full, apply a backpressure strategy (drop oldest, drop newest, or block producer)
- Emit a warning/metric when backpressure is applied
- Optionally allow per-handler configuration

## Acceptance Criteria

- [ ] Event bus enforces a max queue depth (configurable, sensible default)
- [ ] Overflow strategy is configurable (drop oldest, drop newest, block)
- [ ] Warning logged when events are dropped
- [ ] Metrics for queue depth and dropped events available for monitoring

NodeJSmith:
## Prior Art Context

### The problem (Ronacher, 2020)
Armin Ronacher's ["I'm not feeling the async pressure"](https://lucumr.pocoo.org/2020/1/1/async-pressure/) identifies unbounded buffering as the root cause of async system failures. Async frameworks default to accepting unlimited work and buffering silently until OOM. A sensor polling at 1Hz can generate events faster than a slow handler processes them. The fix must be architectural (bounded queues, demand-driven flow, or load shedding), not tactical.

### Approaches from prior art

1. **Demand-driven backpressure** (Elixir GenStage/Broadway) — consumers tell producers how many events they can handle. Producers only emit that many. Gold standard but requires a fundamentally different API than push-based `on_state_change`. Not directly applicable.

2. **Bounded queues with load shedding** — set a max queue depth per listener. When full, either drop oldest events, drop newest, or signal the producer. Most practical for hassette's push-based model.

3. **Service readiness checks** (Ronacher) — a `is_ready` flag on handlers/services. The bus checks readiness before dispatching. If not ready, events are queued (bounded) or dropped with a warning.

4. **`partition_by` ordering** (Broadway) — events with the same key (e.g., entity_id) are routed to the same processor, guaranteeing per-entity ordering while allowing parallelism across entities. Relevant if hassette ever processes events concurrently.

### Recommendation

Bounded queues with configurable overflow policy (drop-oldest, drop-newest, or log-and-drop) is the most practical approach for hassette. Per-listener queue depth gives apps control over their own backpressure behavior without changing the push-based API.

Prior art brief: `design/research/2026-04-17-event-bus-api/research.md`

[...]

PR #1487, opened 2026-07-30
feat: add event priority tiers for bus dispatch under load

### LINKED ISSUE #671: Add event priority tiers for dispatch under load
## Description

Classify events into priority tiers (error > user_action > state_change > sensor_update). Under load, low-priority events can be sampled or deferred first while high-priority events always dispatch immediately. Complements the dispatch semaphore (#72).

## Prior Art

design/research/2026-05-02-bus-backpressure/research.md — Pattern 6 (Per-Subscription Policy). Different consumers/events get different treatment based on priority classification.

## Acceptance Criteria

- [ ] Event priority enum or classification mechanism
- [ ] Priority considered during dispatch scheduling (high-priority first)
- [ ] Under overload: low-priority events sampled/deferred before high-priority
- [ ] Configuration: per-listener or per-topic priority assignment

NodeJSmith:
## Summary

Every bus listener now carries a priority tier — `critical`, `high`, `normal`, or `low` — resolved once at registration from an explicit `event_priority=` or, when omitted, classified from the listener's topic. The tier decides two things once the global dispatch semaphore saturates: which listeners are handed a slot first, and which ones lose the event instead of waiting for one.

This is Layer 3 of the #72 backpressure plan, sitting on the Layer 1 dispatch semaphore (#678) and the Layer 2 per-listener backpressure policy (#1076).

## What changed

**Classification** (`src/hassette/bus/priority.py`) is a small explicit table, not a pattern language. Connectivity and service-status topics are `critical`; service calls, automations, and scripts are `high`; `sensor.*` state changes and handler-execution telemetry are `low`; everything else — including any topic hassette does not publish itself — is `normal`, so an unknown topic is never shed.

`sensor` is the only entity domain classified `low`: it is the highest-volume domain in a typical install and its readings are cumulative, so the next reading supersedes a shed one. `binary_sensor` is deliberately excluded — motion and door contacts are edge-triggered, where a shed event is information that never comes back.

**At the saturation gate** (`BusService.should_shed`) the two extreme tiers override the backpressure policy and the middle two defer to it: `critical` waits for a slot even under `drop_newest`, `low` sheds even under `block`, and `high`/`normal` behave exactly as they did before tiers existed. A shed event increments the existing `backpressure_dropped` counter, so it surfaces in the UI and web API with no new column.

**Dispatch ordering** sorts the fan-out by tier, keeping route specificity and the router's ordering priority as stable tie-breakers.

**Persistence and UI:** the resolved tier is written to `listeners.event_priority` (migration 012, backfilling `'normal'`) and rendered as a `tier` chip on the listener detail panel for anything other than `normal`.

## Behavior change

A listener on a `sensor.*` entity that relied on the default lossless `block` policy can now lose events while the bus is at its concurrency ceiling. Passing `event_priority="normal"` opts out, and the docs page leads with that case. `StateProxy`'s state-cache listener is pinned to `critical` so the cache still updates ahead of app handlers under the new ordering.

## Testing

- `nox -s dev` equivalent (unit + integration, `-n 4`): **8496 passed, 1 skipped, 2 xfailed**.
- `prek -a`: all hooks pass. `prek -a --hook-stage pre-push` also passes, including pyright, schema freshness, CLI-surface drift, and the sync-facade codegen checks.
- New coverage: `tests/unit/bus/test_priority.py` (classification table), `tests/unit/core/test_bus_dispatch_semaphore.py` (shed/wait decisions per tier), `tests/unit/bus/test_listeners.py` (option coercion and validation), `tests/unit/test_schema_migration.py` + `tests/integration/database/` (migration 012), and `tests/integration/bus/test_execution_modes.py`.
- Verified on the running demo stack rather than only in tests: `climate_controller`'s `sensor.*temperature*` listeners resolve to `low` and `security_monitor`'s `hass.event.call_service` listener to `high` via `GET /api/bus/listeners`, and the `tier` chip renders in the UI.
- Both doc reviews were run on the touched pages. `priority.md`: 33 claims checked, no findings. `backpressure.md`: 27 claims checked, one confirmed finding, fixed here (see below).

## Screenshots

The `tier` chip on a `low`-tier listener, generated through `scripts/capture_screenshots.py` from a new `docs/screenshots.yml` entry:

Listener detail panel showing the tier chip on a sensor listener classified as low

## Follow-up commit

The second commit fixes a pre-existing sentence in `backpressure.md` that this change made wrong: a zero `backpressure_dropped_count` was described as evidence the bus stayed below saturation, but a `block` listener at `normal` tier never records a drop at all, so its zero carries no saturation signal either way.

## Note

`mkdocs build --strict` currently aborts on two external intersphinx inventory fetches returning HTTP 429 (`whenever.readthedocs.io`, `docs.aiohttp.org`). That is an upstream rate limit, not a content error — the non-strict build is clean and reports no warnings for the changed pages.

Closes #671


NodeJSmith:
@coderabbitai review

NodeJSmith:
Closing — this issue is too complex for the autofix pipeline. 36 files changed, new migration, new enum, codegen, frontend, and docs for what was labeled size:small. Will be handled manually with proper design work.

PR #1496, opened 2026-07-31
fix: key the state model cache on last_updated, not the context id

thiesgerken:
Hi! Thanks for the project, I've been using it for a few months now. I encountered a problem with the caching, and me and my agent came up with the following solution:

## Problem

`DomainStates` memoises the validated state model per entity and uses the **context id** as its primary invalidation key ([`state_manager.py`](https://github.com/NodeJSmith/hassette/blob/main/src/hassette/state_manager/state_manager.py)):

```python
if cached is not None and context_id is not None and cached.context_id == context_id:
    return cached.model
```

A context id identifies the *cause* of a change, not its content. From the Home Assistant context documentation (<https://data.home-assistant.io/docs/context/>):

> Whenever anything (e.g. an automation or user interaction) triggers a new change, a new context is assigned. This context will be attached to **all events and states** that happen as result of the change.

So "one context, several states" is the documented contract. Two everyday patterns hit it:

- **A light with a transition.** The ramp and the settled value are written under the originating `light.turn_on` context. On my instance a lamp was switched on at 03:00:00 and went off at 03:00:02, and the logbook entry for the *off* state still reads `context_domain: light, context_service: turn_on` — it inherited the turn-on context.
- **An automation run.** Every action inherits the automation's context, so a `repeat` toggling an `input_boolean` ten times produces ten states under one id.

When that happens the first state is cached and answers for every later one, and the entry is only replaced once a state with a *different* context id arrives. For an entity that is not touched again, that is never. `last_updated` was not consulted, so a strictly newer state did not invalidate.

`StateProxy.load_cache()` does not help: it replaces `StateProxy.states` wholesale, but the fresh dict carries the same context id, so the short-circuit still fires. A watchdog of mine called `load_cache()` every five minutes and logged the divergence eight times in a row without fixing it; only a process restart helped.

## Fix

Key the fast path on `last_updated`, which moves whenever the state or its attributes change and therefore identifies the content:

```python
if cached is not None and last_updated is not None and cached.last_updated == last_updated:
    return cached.model
```

The context id is dropped rather than combined: if `last_updated` matches, the content is identical and the context adds nothing; if it differs, a context match must *not* return the cached model. The `deepfreeze` comparison stays as the fallback for states that carry no `last_updated`.

The cache stays just as effective — a repeated read of an unchanged state still returns the identical object and still calls `coerce_and_construct` only once.

## Tests

- `test_shared_context_id_does_not_mask_a_newer_state` — new regression test, fails on `main` (`assert True is False`) and passes with the fix.
- `test_context_id_match_returns_cached_object` → renamed to `test_last_updated_match_returns_cached_object`; the states in it are identical, so it passes either way, but the name now says what it checks.
- `test_frozen_state_match_returns_cached_object` — the two states carried the same `last_updated`, so after this change the new fast path would answer and the `deepfreeze` branch would no longer be exercised. The states now omit `last_updated` so the test keeps covering the fallback.

`tests/unit/test_state_manager.py` and `tests/unit/state_manager/` pass (51 tests); ruff check and format are clean. I did not run the full suite locally.

## How I hit this

A light control app of mine switched on a lamp in an empty room. Its rules are configured so that they may adjust that lamp but never switch it on, and the guard is `current > 0 or rule.can_turn_on` — the app believed the lamp had been at brightness 15 since 03:00 while HA had it off, so the "adjustment" became a `light.turn_on`. Same root cause on two other occasions with an `input_boolean` toggled by an automation.

### DIFF
--- a/src/hassette/state_manager/state_manager.py
+++ b/src/hassette/state_manager/state_manager.py
@@ -88,13 +88,13 @@ class DomainStates(Mapping[str, StateT]):
         if cached is not None and last_updated is not None and cached.last_updated == last_updated:
             return cached.model
 
-        # no last_updated to go by, so compare the frozen states
+        # fast path didn't match — compare full content
         frozen_state = deepfreeze(state)
         if cached is not None and cached.frozen_state == frozen_state:
             return cached.model
 
         validated = STATE_REGISTRY.coerce_and_construct(self._model, state, entity_id)
-        self._cache[entity_id] = CacheValue(last_updated, frozen_state, validated)
+        self._cache[entity_id] = CacheValue(last_updated=last_updated, frozen_state=frozen_state, model=validated)
         return validated
 
     def to_dict(self) -> dict[str, StateT]:
--- a/src/hassette/state_manager/state_manager.py
+++ b/src/hassette/state_manager/state_manager.py
@@ -27,7 +27,7 @@ HOME_STATE = "home"
 
 
 class CacheValue(Generic[StateT], NamedTuple):
-    context_id: str | None
+    last_updated: str | None
     frozen_state: frozendict
     model: StateT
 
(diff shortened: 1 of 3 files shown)


NodeJSmith:
Thanks! Can't believe I hadn't hit this myself yet, appreciate the clear write up and PR.

PR #1518, opened 2026-08-04
docs: replace stale pyright-ignore note in cache patterns

### LINKED ISSUE #1360: Clean up stale pyright-ignore admonition in cache patterns docs
## Description

Surfaced by the integration reviewer during T04 (cache docs async migration, PR for design/specs/013-resource-cache-redesign). Predates this change and was not touched by its diff.

## Key Items

- `docs/pages/core-concepts/cache/patterns.md` (API Response Caching section, around line 39) carries an admonition titled `Why the \`# pyright: ignore\` comments?`, explaining that the example snippet suppresses type-checker warnings.
- No version of `docs/pages/core-concepts/cache/snippets/cache_api_response.py` (before or after the T04 rewrite) actually contains a `# pyright: ignore` comment — verified via `git show` against the pre-T04 revision and the current file.
- The admonition describes something that isn't in the code it refers to, which will confuse a reader trying to find the comments being described.

## Acceptance Criteria

- [ ] Either add the described `# pyright: ignore` usage to `cache_api_response.py` (if the untyped-return problem is real and worth demonstrating), or rewrite/remove the admonition in `patterns.md` to match what the snippet actually does.

NodeJSmith:
## Summary

The **API Response Caching** section of the cache docs carried an admonition titled *"Why the `# pyright: ignore` comments?"* — but the snippet it pointed at (`cache_api_response.py`) has never contained a single `# pyright: ignore` comment, in any revision. A reader following the note would go looking for comments that aren't there.

Rather than adding suppressions just to justify the note, this fixes it in the direction that matches how the cache API actually works: the typing problem is real, but the fix is annotation, not suppression.

## What changed

- **`docs/pages/core-concepts/cache/snippets/cache_api_response.py`** — introduces a `WeatherEntry = tuple[ZonedDateTime, dict]` alias and annotates the cache read as `entry: WeatherEntry | None`, so the snippet demonstrates the shape it later unpacks into `cached_time, data`.
- **`docs/pages/core-concepts/cache/patterns.md`** — retitles the admonition to "Typing cached values" and rewrites it to explain the actual mechanism: `Cache.get()` is generic over `T` with `default: T | None = None`, so an unannotated read leaves `T` unsolved and resolves to `None` for the type checker. Annotating the target solves `T`; passing `default=` does too (`await self.cache.get("count", default=0)` reads as `int | None`).

## Testing

- `uv run nox -s dev` — 8604 passed, 1 skipped, 2 xfailed
- `prek -a` — all hooks pass (ruff, prettier, eslint, stylelint, tsc, and the repo's custom checkers)
- Verified the claims in the new prose against the real signature: `async def get(self, key: str, default: T | None = None) -> T | None` in `src/hassette/cache/protocol.py` and `wrapper.py`.

Docs-only change; no runtime or frontend behavior is affected.

Closes #1360


NodeJSmith:
@coderabbitai review

PR #1587, opened 2026-08-09
fix: harden apiFetch error-path tests against silent passes

### LINKED ISSUE #1511: Harden apiFetch error tests against silent passes and extract mock-response factory
Surfaced by a code-quality scan of `frontend/src/api/client.test.ts`. All findings predate the current work.

## Finding 1: Two error tests can pass without ever running their assertions

`extracts message field when detail is absent` (lines ~38-46) and `falls back to status text when body is not JSON` (lines ~53-62) wrap the call in a bare `try { await apiFetch(...) } catch (err) { ...assertions... }` and never assert that a throw actually occurred.

If `apiFetch` regresses so that it resolves instead of rejecting — for example, a future change to the `!response.ok` branch in `client.ts:26-36` — the `catch` block simply never executes, none of its `expect` calls run, and both tests report green while verifying nothing. The first test in the file avoids this by independently proving rejection via `rejects.toThrow(ApiError)`; these two do not.

**Fix:** standardize on one throw-and-inspect pattern for all three error tests. Either capture the rejection once (`const err = await apiFetch(...).catch((e) => e); expect(err).toBeInstanceOf(ApiError);`) or add an explicit `expect.fail("expected apiFetch to throw")` immediately after the `await` inside each `try`.

## Finding 2: First test invokes `apiFetch` twice for one assertion set

`extracts detail from JSON error response` calls `apiFetch("/apps/bad")` once through `expect(...).rejects.toThrow(ApiError)` and again inside a `try/catch` purely to read `status` and `message`. The two invocations are separate executions, so the properties being asserted are not provably from the run that was checked for the throw, and the mocked fetch is exercised redundantly.

**Fix:** collapse to a single invocation whose rejection is captured once, then assert instance type, `status`, and `message` against that one error.

## Finding 3: Mock response shape rebuilt inline four times

The `{ ok, status, statusText, json }` mock object is hand-written in all four tests with only the values changing (lines ~17-22, ~34-39, ~50-55, ~66-69).

**Fix:** add two small factories at the top of the file — `mockErrorResponse(status, statusText, body)` and `mockOkResponse(body)` — and have each test call one. This also gives one place to keep the mock in sync if `apiFetch` starts reading additional response fields.

## Acceptance Criteria

- [ ] Every error-path test in `client.test.ts` fails if `apiFetch` stops throwing (verify by temporarily making `apiFetch` resolve on a non-ok response and confirming the tests go red)
- [ ] No test calls `apiFetch` more than once for a single scenario
- [ ] Mock `fetch` responses are constructed through a shared factory rather than four inline literals
- [ ] `cd frontend && npm run test` passes

## Reviewed and not filed

Three findings were assessed and dropped as invalid: bare `422`/`500`/`502` literals and their `statusText` strings (each appears once per test as its arrange/assert pair — a named constant would add indirection without a second consumer), and the generic `result` name on line ~70 (the project style rule explicitly endorses short contextually-obvious names in local scope).

NodeJSmith:
## Summary

Fixes three test-quality gaps in `frontend/src/api/client.test.ts` flagged by a code-quality scan of the file (#1511):

- Two error-path tests (`extracts message field when detail is absent`, `falls back to status text when body is not JSON`) wrapped `apiFetch` in a bare `try { ... } catch (err) { ...assertions... }` with no assertion that a throw actually occurred. If `apiFetch` ever regressed to resolving instead of rejecting on a non-ok response, these tests would pass vacuously — their assertions live entirely inside the `catch` block. All three error tests are now standardized on a single throw-and-inspect pattern (`await apiFetch(...).catch((e) => e)` followed by explicit `instanceof`/property assertions), so each one provably fails if `apiFetch` stops throwing.
- The first test (`extracts detail from JSON error response`) invoked `apiFetch` twice for one assertion set — once via `expect(...).rejects.toThrow(ApiError)` and again in a separate `try/catch` to read `status`/`message`. Collapsed to a single invocation whose rejection is captured once.
- The `{ ok, status, statusText, json }` mock response object was hand-written inline in all four tests. Extracted into `mockErrorResponse()`, `mockNonJsonErrorResponse()`, and `mockOkResponse()` factories at the top of the file.

## Testing

- `cd frontend && npx vitest run src/api/client.test.ts` — 10 passed
- `uv run nox -s dev` (full backend suite via `ptest`) — 9203 passed, 1 pre-existing unrelated failure isolated and confirmed to be caused by an ambient `HASSETTE__CLI__AUTH_TOKEN` env var on this machine, not this change (passes cleanly with the var unset)
- `prek -a` — all hooks pass (ruff, eslint, tsc, stylelint, etc.)

Closes #1511

### DIFF
--- a/frontend/src/api/client.test.ts
+++ b/frontend/src/api/client.test.ts
@@ -2,6 +2,21 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { ApiError, apiFetch, postSession } from "./client";
 
+/** Builds a mock non-ok `fetch` response whose `json()` resolves to `body`. */
+function mockErrorResponse(status: number, statusText: string, body: unknown) {
+  return { ok: false, status, statusText, json: () => Promise.resolve(body) };
+}
+
+/** Builds a mock non-ok `fetch` response whose `json()` rejects, as if the body were not JSON. */
+function mockNonJsonErrorResponse(status: number, statusText: string) {
+  return { ok: false, status, statusText, json: () => Promise.reject(new Error("not json")) };
+}
+
+/** Builds a mock ok `fetch` response whose `json()` resolves to `body`. */
+function mockOkResponse(body: unknown) {
+  return { ok: true, json: () => Promise.resolve(body) };
+}
+
 describe("apiFetch", () => {
   const originalFetch = globalThis.fetch;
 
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #1607, opened 2026-08-13
perf(ui): scope execution-completion subscriptions in app-detail children

### LINKED ISSUE #1542: Extend fleet-wide re-render scoping to RecentActivitySection, job-detail, and listener-detail
## Description

Surfaced by an integration-reviewer pass on PR #1528 (`autofix/issue-1465`, which fixed `AppDetailPage`'s own selector scoping per issue #1465). `AppDetailPage` now selects only its own app's matching execution-completion record out of the WS `executionCompleted` batch instead of subscribing to the whole fleet-wide array, so it no longer re-renders on unrelated apps' activity. Three sibling components mounted inside that same page still use the old broad-subscribe-then-filter pattern, so the page's own re-render-scoping fix doesn't fully hold at the page level — a child still re-renders on unrelated apps' completions.

## Key Items

- `frontend/src/components/app-detail/recent-activity-section.tsx:127-136` — subscribes to the full `useAppStore((s) => s.executionCompleted)` batch and filters only inside the `useQueryInvalidator` callback (`events?.some((e) => e.app_key === appKey)`). Rendered directly inside `AppDetailPage`'s overview tab, so it re-renders on every other app's execution completion.
- `frontend/src/components/app-detail/job-detail.tsx:67-77` and `frontend/src/components/app-detail/listener-detail.tsx:83-90` — same broad-subscribe-then-filter pattern, predating PR #1528.
- These three implementations of "find this app/job/listener's matching execution-completion record" have drifted independently from the scoped `findExecution` approach now used in `app-detail.tsx`. A future change to matching semantics (e.g. adding instance-index matching) applied to one won't propagate to the others.

## Acceptance Criteria

- [ ] `RecentActivitySection`, `job-detail.tsx`, and `listener-detail.tsx` scope their `executionCompleted` store subscriptions the same way `app-detail.tsx`'s `findExecution` does, so none of them re-render on unrelated apps' activity.
- [ ] Consider extracting the scoped-lookup logic into one shared hook/selector reused by all four consumers instead of four independent implementations.
- [ ] Existing tests for these three components continue to pass; add render-count regression tests analogous to `app-detail.rerender.test.tsx` where a reasonable seam exists.

NodeJSmith:
## Summary

`AppDetailPage` already scoped its own `executionCompleted` store subscription (PR #1528, issue #1465) so it selects only its app's matching record out of the fleet-wide WS batch instead of subscribing to the whole array. Three components mounted *inside* that same page still used the old broad-subscribe-then-filter pattern, so the page-level fix didn't actually hold — a child re-rendered on every unrelated app's execution completion regardless.

This scopes those children and consolidates the matching logic in one place.

## What changed

- **New `frontend/src/hooks/use-scoped-execution.ts`** — holds the previously-inline `findExecution` logic from `app-detail.tsx`, exposed as `useAppExecution(appKey, kind?)`, `useJobExecution(jobId)`, and `useListenerExecution(listenerId)`. Selecting the *matching record* rather than the batch is what keeps a component off the fleet-wide render path: a batch of only other apps' executions selects to `undefined`, which is `Object.is`-equal to the previous `undefined`, so the store never notifies.
- **`recent-activity-section.tsx`, `job-detail.tsx`, `listener-detail.tsx`** — replaced `useAppStore((s) => s.executionCompleted)` + `events?.some(...)` inside the `useQueryInvalidator` callback with the corresponding scoped hook. The invalidator filter is now just a definedness check, since scoping already happened in the selector.
- **`app-detail.tsx`** — dropped its local `findExecution` and routes through the shared hook, so all four consumers share one implementation of the matching semantics. A future change (e.g. adding instance-index matching) now lands in one place instead of four.

The one remaining render cost is documented on the hook: the batch immediately following a matching completion selects back to `undefined`, which is not `Object.is`-equal to the previously matched record, so exactly one extra render fires per matching completion. Steady-state unrelated activity stays undefined-to-undefined.

## Testing

- New `frontend/src/hooks/use-scoped-execution.test.ts` — render-count assertions per hook: no re-render on another app's/job's/listener's completions, re-render plus correct record on its own, kind narrowing, and consecutive own completions each registering.
- New `frontend/src/components/app-detail/recent-activity-section.rerender.test.tsx` — `Profiler`-based commit-count regression test analogous to the existing `app-detail.rerender.test.tsx`, covering both the unrelated-app and own-app cases.
- Full frontend suite green: 1499 tests across 103 files.
- `prek -a` and `prek pyright -a --stage pre-push` clean; backend suite (`nox -s dev`) green apart from one pre-existing, console-width-dependent CLI table failure (`test_commands_app.py::TestCmdAppActivity::test_human_mode_renders_table`) that reproduces identically on a clean `main` and is untouched by this branch.

No rendered output changes — this only affects when components re-render, hence the `no-visual-change` label.

Closes #1542

### DIFF
--- a/frontend/src/components/app-detail/job-detail.tsx
+++ b/frontend/src/components/app-detail/job-detail.tsx
@@ -9,7 +9,7 @@ import { getJobExecutions, triggerJob } from "../../api/endpoints";
 import { useAsyncAction } from "../../hooks/use-async-action";
 import { useQueryInvalidator } from "../../hooks/use-query-invalidator";
 import { useRelativeTime } from "../../hooks/use-relative-time";
-import { useJobExecution } from "../../hooks/use-scoped-execution";
+import { isExecutionDefined, useJobExecution } from "../../hooks/use-scoped-execution";
 import { useScopedQuery } from "../../hooks/use-scoped-query";
 import { queryKeys } from "../../lib/query-keys";
 import { DETAIL_FETCH_LIMIT } from "../../utils/constants";
@@ -195,7 +195,7 @@ export function JobDetail({ job, appKey, instanceQs, onSwitchToCode }: Props) {
   const nextRunLabel = useRelativeTime(job.next_run ?? null);
   const fireAtLabel = useRelativeTime(job.fire_at ?? null);
 
-  useQueryInvalidator(execution, (exec) => exec !== undefined, queryKeys.jobExecutions(job.job_id));
+  useQueryInvalidator(execution, isExecutionDefined, queryKeys.jobExecutions(job.job_id));
 
   const kindLabel = handlerKindLabel("job", null, job.trigger_type);
   const jobKind = jobHealthKind(job);
--- a/frontend/src/components/app-detail/job-detail.tsx
+++ b/frontend/src/components/app-detail/job-detail.tsx
@@ -9,9 +9,9 @@ import { getJobExecutions, triggerJob } from "../../api/endpoints";
 import { useAsyncAction } from "../../hooks/use-async-action";
 import { useQueryInvalidator } from "../../hooks/use-query-invalidator";
 import { useRelativeTime } from "../../hooks/use-relative-time";
+import { useJobExecution } from "../../hooks/use-scoped-execution";
 import { useScopedQuery } from "../../hooks/use-scoped-query";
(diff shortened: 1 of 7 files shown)


NodeJSmith:
@coderabbitai review

PR #1612, opened 2026-08-13
test: add regression test for concurrent reload_app locking

### LINKED ISSUE #1227: Serialize app start/stop/reload to prevent orphaned instances on concurrent lifecycle operations
## Description

App lifecycle operations (start/stop/reload) are unserialized across every entry point that can trigger them. Nothing in the call chain — the file-watcher handler, the REST routes, or the registry they share — holds a lock across a start/stop/reload sequence for a given `app_key`. Two concurrent reloads for the same app key can each create a fresh set of instances; the second `create_instances` call overwrites the first in the registry, orphaning the first set of live, already-initialized instances (their handlers keep running, but no lifecycle operation can reach them again) while both sets fire handlers simultaneously.

## Reproduction / Trigger

Any of these can race against each other for the same `app_key`, since none of them synchronize:

1. **File-watcher path** — `AppLifecycleService.handle_change_event` (`src/hassette/core/app_lifecycle_service.py:397-438`) is registered directly as the bus handler for `HASSETTE_EVENT_FILE_WATCHER` (`src/hassette/core/app_handler.py:100-104`) and dispatched task-per-event. Two rapid file saves produce two concurrently-running tasks, each calling `refresh_config()` (`app_lifecycle_service.py:440-455`), which mutates shared state (`self.hassette.config.reload()` and `self.registry.set_manifests(...)`) with no lock, racing the other task's reads.
2. **REST path** — `POST /api/apps/{key}/start|stop|reload` (`src/hassette/web/routes/apps.py:105-139`) each call straight into `AppHandler.start_app` / `stop_app` / `reload_app` (`src/hassette/core/app_handler.py:125-135`), which delegate 1:1 to `AppLifecycleService.reload_app` (`app_lifecycle_service.py:321-333`) with no lock.
3. **Bootstrap path** — `AppLifecycleService.bootstrap_apps` (`app_lifecycle_service.py:242-272`) can overlap with either of the above during startup.

Concrete failure interleaving: a user clicks "reload" in the web UI for app `x` at the same moment a file save fires the watcher for `x`. Both tasks call `reload_app("x")`:
- Both run `stop_app("x")` (`app_lifecycle_service.py:305-319`) — the second one finds nothing in the registry (`unregister_app` already returned the instances to the first caller) and just logs a warning.
- Both run `start_app("x")` (`app_lifecycle_service.py:274-303`) → `factory.create_instances(...)` runs twice → the second `create_instances` call overwrites the app's entry in `self.registry.apps[app_key]`, discarding the reference to the first set of instances.
- The first set of instances is now orphaned: fully initialized, still running (bus listeners registered, scheduler jobs live), but unreachable by any future `stop_app`/`reload_app` call for that key — a permanent duplicate that only goes away on full framework restart.

There is no existing lock anywhere in this chain — confirmed by grepping `Lock`/`_lock` in `app_lifecycle_service.py`, `app_handler.py`, and `app_registry.py` (zero matches). The codebase already has a precedent for this exact shape of fix: `SessionManager` uses a single `self._session_lock = asyncio.Lock()` to coordinate concurrent session finalization and crash handling (`src/hassette/core/session_manager.py:48`, used at lines 100 and 132).

## Expected Behavior

Only one start/stop/reload operation can be in flight for a given `app_key` at a time. A second concurrent request for the same key either waits for the first to finish or is otherwise serialized — it must never proceed to `create_instances` while a prior instance set for that key is still live and untracked.

## Actual Behavior

Concurrent lifecycle operations for the same `app_key` race freely. The registry silently loses the reference to the previously-created instance set, orphaning live app instances (with their bus listeners and scheduled jobs still active) and causing duplicate handler execution until the entire framework restarts.

## Acceptance Criteria

- [ ] A per-app-key (or single global) lock in `AppLifecycleService` serializes `start_app`, `stop_app`, and `reload_app` for the same `app_key` across all three entry points (file watcher, REST routes, bootstrap).
- [ ] A regression test drives two concurrent `reload_app("x")` calls (or `asyncio.gather` of `start_app`/`reload_app` for the same key) and asserts exactly one live instance set remains in the registry afterward, with no orphaned listeners or duplicate handler registrations.
- [ ] `refresh_config()`'s mutation of `self.registry.manifests` / `self.hassette.config` is covered by the same serialization (or documented as safe under the chosen lock granularity).
- [ ] Existing app lifecycle tests (`tests/unit/core/test_app_lifecycle_service*.py`, `tests/system/test_app_lifecycle.py`) still pass.

## Affected Areas

- `src/hassette/core/app_lifecycle_service.py` — `start_app` (274-303), `stop_app` (305-319), `reload_app` (321-333), `handle_change_event` (397-438), `refresh_config` (440-455); needs the new lock plumbed through all four.
- `src/hassette/core/app_handler.py` — `start_app`/`stop_app`/`reload_app` (125-135) and the bus registration of `handle_change_event` (100-104); may need no change if the lock lives entirely in the lifecycle service.
- `src/hassette/web/routes/apps.py` — `start_app`/`stop_app`/`reload_app` routes (105-139); no change expected beyond relying on the now-serialized lifecycle service.
- `tests/unit/core/test_app_lifecycle_service.py`, `tests/unit/core/test_app_lifecycle_service_operations.py`, `tests/unit/core/test_app_lifecycle_service_coverage.py` — add the concurrency regression test here.
- `tests/system/test_app_lifecycle.py` — extend if the fix needs end-to-end coverage of the race.

## Context

Verified against the current `fable-review` worktree — all cited line numbers match. Existing lock precedent: `src/hassette/core/session_manager.py:48` (`self._session_lock = asyncio.Lock()`), a single lock guarding two competing lifecycle paths (`finalize_session()` and `on_service_crashed()`), the same shape of problem this issue describes for apps. No lock currently exists anywhere in the app start/stop/reload chain (confirmed via grep across `app_lifecycle_service.py`, `app_handler.py`, `app_registry.py`).

NodeJSmith:
### Regression test for issue #1227

- Adds `test_concurrent_reload_app_calls_serialize_the_whole_stop_and_start_pair` to `TestReloadAppLocking`, proving two concurrent `reload_app("x")` calls — e.g. a UI reload racing a file-watcher reload for the same `app_key` — fully serialize their stop+start sequence as one atomic unit. This is the last open acceptance criterion on #1227: the locking mechanism itself already shipped in #1605 (the app-key lock is now acquired once in `reload_app` and held across `_stop_app_unlocked` + `_start_app_unlocked`), but nothing proved the lock actually covered the whole pair rather than just `start_app`'s own internals. Verified RED against the old partial-lock behavior before confirming GREEN against the current fix.
- The test lives next to its sibling, `test_reload_app_acquires_app_key_lock_once`, in `test_app_lifecycle_service_operations.py` rather than `test_app_lifecycle_service_coverage.py` (which is scoped to branch-coverage gap-filling, a different concern).
- The mocked start-phase callback includes a yield point (`await asyncio.sleep(0)`) — without it, the event loop never gets a chance to interleave a second `reload_app` call during the first's start phase, so the test would pass even against a reintroduced bug that narrows the lock scope back to only the stop phase. This closes a false-pass gap that existed in an earlier draft of the test.

Closes #1227

### DIFF
--- a/tests/unit/core/test_app_lifecycle_service_coverage.py
+++ b/tests/unit/core/test_app_lifecycle_service_coverage.py
@@ -406,6 +406,67 @@ class TestReplayPreReleaseReconciliationSerialization:
         assert call_count == 2
 
 
+class TestReloadAppFullSerialization:
+    async def test_concurrent_reload_app_calls_serialize_the_whole_stop_and_start_pair(
+        self,
+        lifecycle_service: AppLifecycleService,
+        mock_registry: MagicMock,
+        mock_manifest: MagicMock,
+    ) -> None:
+        """Two concurrent reload_app("x") calls -- e.g. a UI reload racing a file-watcher
+        reload for the same app_key (issue #1227) -- must run each call's full stop+start
+        sequence as one atomic unit. `test_reload_app_acquires_app_key_lock_once` proves a
+        single reload_app acquires the lock once; this proves a *second*, concurrent
+        reload_app for the same key is genuinely blocked until the first's entire
+        stop+start pair has completed -- not just serialized within start_app's own
+        internals, which is the gap #1227 originally described (the second racer's
+        create_instances() overwriting the registry entry the first racer just created,
+        orphaning a live, already-initialized instance set).
+        """
+        mock_registry.get_manifest = Mock(return_value=mock_manifest)
+
+        gate = asyncio.Event()
+        first_entered = asyncio.Event()
+        call_order: list[str] = []
+
+        async def gated_stop_unlocked(_app_key: str) -> None:
+            call_order.append("stop_start")
+            if call_order.count("stop_start") == 1:
+                first_entered.set()
+                await gate.wait()
+            call_order.append("stop_end")
+
+        async def recording_start_unlocked(_app_key: str, _app_manifest: MagicMock, _force_reload: bool) -> None:
+            call_order.append("start")
+
+        lifecycle_service._admit_start = AsyncMock()  # pyright: ignore[reportAttributeAccessIssue]
+        lifecycle_service._stop_app_unlocked = gated_stop_unlocked  # pyright: ignore[reportAttributeAccessIssue]
+        lifecycle_service._start_app_unlocked = recording_start_unlocked  # pyright: ignore[reportAttributeAccessIssue]
+
+        task1 = asyncio.create_task(lifecycle_service.reload_app("test_app"))
+        await asyncio.wait_for(first_entered.wait(), timeout=1.0)
+
+        task2 = asyncio.create_task(lifecycle_service.reload_app("test_app"))
+        lock = lifecycle_service._get_app_key_lock("test_app")
+        await wait_for(lambda: bool(lock._waiters), desc="task2 queued on the app-key lock")
+
+        # task2 must be blocked acquiring the lock -- it must not have entered its own
+        # stop phase while task1's stop+start pair is still mid-flight.
+        assert lock.locked()
+        assert call_order.count("stop_start") == 1
+        assert not task2.done()
+
+        gate.set()
+        await asyncio.wait_for(task1, timeout=1.0)
+        await asyncio.wait_for(task2, timeout=1.0)
+
+        # Each reload's stop and start run back-to-back, and the second reload's stop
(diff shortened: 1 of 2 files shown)


NodeJSmith:
@coderabbitai review

PR #1615, opened 2026-08-14
test: name the repeated timestamp fixture literal in ws-validator tests

### LINKED ISSUE #1551: Extract repeated timestamp fixture literal in ws-validator tests
Surfaced by an automated code-quality scan of `frontend/src/api/ws-validator.test.ts`. Two of the three reviewers (LLM training-bias, deferred-debt) returned clean; one finding survived validity assessment.

## Finding: repeated `1700000000` fixture literal

**What is wrong**

The arbitrary epoch value `1700000000` is hardcoded five times across four separate test cases:

- `ws-validator.test.ts:10` — `connected` message `timestamp`
- `ws-validator.test.ts:29` — `execution_completed` message `timestamp`
- `ws-validator.test.ts:39` — `log` record inner `timestamp`
- `ws-validator.test.ts:52` — `log` message outer `timestamp`
- `ws-validator.test.ts:69` — `unknown_type` message `timestamp`

Unlike the `type` discriminators (which are wire-protocol values that should stay literal so a backend rename fails the test), this timestamp carries no contract meaning — it is pure fixture noise. Repeating it invites drift: a future edit that tweaks one occurrence leaves the others silently inconsistent, and the reader has no signal about whether the two timestamps inside the `log` case (lines 39 and 52) are intentionally equal or coincidentally so.

**Proposed fix**

Declare a single module-level constant above the `describe` block and reference it at all five sites:

```ts
const FIXTURE_TIMESTAMP = 1700000000;
```

This also makes the line 39 / line 52 relationship explicit — both reference the same named value rather than looking like two independent magic numbers.

## Not filed (assessed as invalid)

Three further nitpicks were dropped after checking them against the code: a suggested shared message factory (the three fixtures are distinct union members with no shared `data` shape, so a factory removes nothing), extracting the `"connected"` discriminator to a constant (hardcoding wire values is deliberate — a constant would silently track a backend rename and void the assertion), and renaming `err` to a full word (the file already uses the abbreviation `msg` throughout, so `err` is consistent).

NodeJSmith:
## What changed

`frontend/src/api/ws-validator.test.ts` hardcoded the arbitrary epoch value `1700000000` at five sites across four test cases. This replaces all five with a single module-level `FIXTURE_TIMESTAMP` constant.

## Why

The timestamp carries no wire-protocol meaning, unlike the `type` discriminators in the same file — those stay literal on purpose, so a backend rename fails the test rather than silently tracking along. The timestamp is pure fixture noise, and repeating it invited drift: editing one occurrence would leave the others quietly inconsistent.

It also makes one relationship explicit. The `log` test case carries two timestamps — the inner log-record field and the outer envelope field — and as two independent magic numbers a reader had no way to tell whether they were intentionally equal or coincidentally so. Both now reference the same named value.

Scope is deliberately narrow: no test logic, assertions, or fixture shapes changed, and the substituted value is identical at every site.

## Testing

- `npm run test -- --run src/api/ws-validator.test.ts` — 7 passed
- `uv run nox -s dev` — 9269 passed, 1 skipped, 2 xfailed
- `prek -a` and the `pre-push` hook stage (including `pyright`) — all hooks pass

Closes #1551

### DIFF
--- a/frontend/src/api/ws-validator.test.ts
+++ b/frontend/src/api/ws-validator.test.ts
@@ -2,12 +2,14 @@ import { describe, expect, it } from "vitest";
 
 import { validateWsMessage, WsValidationError } from "./ws-validator";
 
+const FIXTURE_TIMESTAMP = 1700000000;
+
 describe("validateWsMessage", () => {
   it("validates a connected message", () => {
     const msg = {
       type: "connected",
       data: { uptime_seconds: 123.4, entity_count: 50, app_count: 3, version: "0.25.0" },
-      timestamp: 1700000000,
+      timestamp: FIXTURE_TIMESTAMP,
     };
     expect(validateWsMessage(msg)).toEqual(msg);
   });
@@ -26,7 +28,7 @@ describe("validateWsMessage", () => {
           error_type: null,
         },
       ],
-      timestamp: 1700000000,
+      timestamp: FIXTURE_TIMESTAMP,
     };
     expect(validateWsMessage(msg)).toEqual(msg);
   });
@@ -36,7 +38,7 @@ describe("validateWsMessage", () => {
       type: "log",
       data: {
         seq: 1,
-        timestamp: 1700000000,
+        timestamp: FIXTURE_TIMESTAMP,
         level: "INFO",
         logger_name: "hassette.test",
         func_name: null,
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #1673, opened 2026-08-22
refactor: extract shared web route params and CLI query-param construction

### LINKED ISSUE #1570: Extract web route param and validation patterns
## Description

PMD CPD duplicate-code analysis flagged ~8 clusters of source-code duplication (not test code) across web routes and CLI commands:

- Telemetry routes repeat the same `app_key + instance_index + since + source_tier` parameter declarations
- App action routes (`start`/`stop`/`reload`) share `_validate_app_key` + `_require_known_app` + try/except + log + return pattern
- CLI commands share `make_client(ctx)` + params dict + optional param assignment
- Service `config_log_level` properties repeat across `bus_service`, `command_executor`, `service_watcher`

## Key Items

- `src/hassette/web/routes/telemetry.py` — repeated route parameter declarations
- `src/hassette/web/routes/apps.py` — start/stop/reload share validation + error handling + logging pattern
- `src/hassette/cli/commands/` — app/job/listener/log share client + param construction
- `src/hassette/core/` — config_log_level properties on 3 services

## Acceptance Criteria

- [ ] Telemetry route parameter declarations consolidated (e.g. shared dependency or param model)
- [ ] App action route pattern extracted or simplified
- [ ] CLI param construction deduplicated
- [ ] config_log_level property pattern evaluated (may be dup-ignore if deemed intentional)

NodeJSmith:
## Summary

PMD CPD flagged several source-code duplication clusters across the web routes and CLI commands. This refactor expresses each of them once. No behavior change — the HTTP API contract and CLI flag surface are identical.

## What changed

**Shared route parameter annotations** — `web/dependencies.py` now owns `AppKeyPath`, `InstanceIndexQuery`, `SinceQuery`, `SourceTierQuery`, and `LimitQuery`. Route signatures annotate with these instead of repeating a `Query(...)`/`Path(...)` call each time, which also removes the `reportCallInDefaultInitializer` suppressions those calls needed.

**Telemetry filter dependency** — per-app telemetry routes take the `instance_index`/`since`/`source_tier` set as one injected `TelemetryFiltersDep` and splat it via `query_kwargs`. FastAPI flattens a dependency's parameters into the operation, so the generated schema is unchanged: regenerating `frontend/openapi.json` produces a byte-identical file, which is why no schema or frontend type files appear in this diff.

**App action routes** — `apps.py`'s `start`/`stop`/`reload` endpoints share `_run_app_action`, which holds the validation, error mapping, logging, and response construction all three had repeated. The 409 branch is unreachable from `stop` (only `_admit_start` raises `AppBootstrapNotReleasedError`); its docstring says so.

**CLI query params** — `cli/client.py` gains `query_params()`, replacing the per-flag `if x is not None` blocks in the app, job, listener, and log commands.

**Docs** — `src/hassette/web/CLAUDE.md` documents the shared-parameter and telemetry-filter patterns so new routes follow them.

## Deliberately not deduplicated

- **Cyclopts command signatures** stay duplicated behind `dup-ignore` markers. They are the CLI's declared flag surface, not copy-pasted logic.
- **`config_log_level` overrides** stay as-is, with the rationale recorded on the base class property. Collapsing them to a `getattr` on a string-named class attribute would trade a checked attribute access for a string across 30 files, three of which are generated sync facades. They can't carry `dup-ignore` markers either: each flagged fragment swallows an unrelated statement above the override, so a marker there would mislabel that statement.

## Testing

- `uv run nox -s dev` — 6927 passed, 1 skipped, 2 xfailed
- `prek -a` — all hooks pass

Closes #1570

### DIFF
--- a/src/hassette/cli/client.py
+++ b/src/hassette/cli/client.py
@@ -228,6 +228,14 @@ class HassetteCLIClient:
         self.error_usage(f"Instance {instance!r} not found for app {app_key!r}. Available instances: {names}")
         raise AssertionError("unreachable")
 
+    def resolve_instance_or_none(self, app_key: str, instance: str | None) -> int | None:
+        """Resolve an instance selector, passing ``None`` through unchanged.
+
+        Convenience wrapper for CLI commands where an unset ``--instance`` flag means
+        "all instances" and should stay ``None`` rather than resolve to an index.
+        """
+        return None if instance is None else self.resolve_instance(app_key, instance)
+
     def _echo_success_target_and_warnings(self) -> None:
         """Surface the resolved non-loopback target and any TLS-verification warning once per
         invocation on the success path.
--- a/src/hassette/cli/client.py
+++ b/src/hassette/cli/client.py
@@ -28,6 +28,17 @@ DEFAULT_TIMEOUT = 10.0
 T = TypeVar("T")
 
 
+def query_params(**values: Any) -> dict[str, Any]:
+    """Build a query-parameter dict from CLI flags, dropping every flag left unset.
+
+    Commands share one convention: an unset flag is ``None`` and must not reach the server,
+    so the endpoint applies its own default. Writing that as ``query_params(since=since,
+    limit=limit)`` keeps the mapping from flag to query key on one line per command instead
+    of an ``if x is not None`` block per flag.
+    """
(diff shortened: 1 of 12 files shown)


NodeJSmith:
@coderabbitai review

PR #1675, opened 2026-08-23
refactor(bus): split the generated sync facade into core and named-event modules

### LINKED ISSUE #1573: Generated sync.py facade exceeds file size threshold
## Description

`src/hassette/bus/sync.py` is 888 lines (threshold: 800). Surfaced by `tools/check_file_size.py`, now a visible non-blocking CI job.

This file is auto-generated (header: "Auto-generated synchronous facade for `Bus`. Do not edit this file directly. It is generated from `bus.Bus` by `codegen/src/hassette_codegen/sync_facade/`."). Any fix belongs in the generator (`codegen/src/hassette_codegen/sync_facade/`), not in this file directly — e.g. splitting the generated output across multiple modules, or reducing per-method boilerplate in the generated template.

Note: issue #938 tracks consolidating the *generator's own* code duplication (table-driven generation) but does not address the *generated output's* line count — the two are separate concerns and #938 is not expected to resolve this on its own.

## Acceptance Criteria

- [ ] Generated `sync.py` brought under 800 lines (via generator template changes, e.g. splitting output or trimming boilerplate)
- [ ] No behavior changes
- [ ] Tests pass

NodeJSmith:
## Summary

`src/hassette/bus/sync.py` is generated, and it had grown to 888 lines — past the repo's 800-line
`HSL102` threshold. Since the file carries a "do not edit this file directly" header, the fix
belongs in the generator, not the output.

The named-event shortcuts (`on_homeassistant_start`, `on_app_running`, and the twelve others that
wrap a fixed topic) are more than half the generated facade and form a natural seam: they're thin
conveniences over `Bus.on()` with a fixed topic, distinct from the core registration API. The
generator now emits them into a second module.

- `bus/sync.py` — the core registration API, **426 lines**
- `bus/sync_events.py` — `BusSyncEventShortcuts`, the named-event shortcuts, **496 lines**

`BusSyncFacade` inherits from `BusSyncEventShortcuts`, so `bus.sync` exposes exactly the same
surface it did before. All 27 facade methods are byte-identical to their previous versions.

## What changed in the generator

- `_generate_facade` takes an optional `keep` filter, so one source class can feed two generated
  modules. It defaults to emitting everything, leaving the other four facades untouched.
- `BUS_EVENT_SHORTCUTS` names the split. Generation **fails loudly** (`SystemExit`) if that set
  lists a method `Bus` no longer defines — otherwise a renamed shortcut would silently drift back
  into `sync.py` with nothing to flag it.
- The codegen drift gate (`--target all --check`) now checks both output files.
- Writing any facade over 800 lines prints a warning at generation time, where the person who just
  grew the source class can act on it.

### Why inheritance rather than composition

`ApiSyncFacade` and `RecordingSyncFacade` use `.helpers`-style composition, but they wrap a real
async-side sub-resource (`Api.helpers`), so `api.sync.helpers.x()` reads the same on both sides.
`Bus` has no sub-resource behind its named-event shortcuts, so composition would invent a
`bus.sync.events.x()` surface with no async counterpart — a breaking API change in service of a
file-size fix. The rationale is recorded as a comment next to
`BUS_EVENT_SHORTCUTS_CLASS_HEADER`.

`BusSyncEventShortcuts` is exported from `hassette.bus` so the mkdocstrings reference keeps
documenting those methods, which inheritance would otherwise hide.

## Testing

- `uv run nox -s dev` — **6930 passed, 1 skipped, 2 xfailed** (2:44)
- `prek -a` — all 28 hooks pass, including `house-lint`
- `uv run python codegen/src/hassette_codegen/sync_facade/ --target all --check` — no drift, so the
  committed output matches what the generator produces
- `uv run house-lint check` — `src/hassette/bus/sync.py` no longer appears in the HSL102 findings
  (it was 888 lines on `main`). The 15 remaining findings are the pre-existing backlog that
  `file-sizes` already tracks on `main`.

New test `tests/unit/test_bus_sync_facade_surface.py` pins the merged surface: every public `Bus`
method still has a facade counterpart, and the shortcuts really do arrive by inheritance rather
than being duplicated into `sync.py`.

No behavior change, so no docs-site or frontend updates apply.

Closes #1573


NodeJSmith:
@coderabbitai review

Issue #1583, opened 2026-08-09
Decompose test_command_executor_pipeline.py


NodeJSmith:
## Description

`tests/unit/core/test_command_executor_pipeline.py` is 855 lines (threshold: 800). Surfaced by `tools/check_file_size.py`, now a visible non-blocking CI job.

## Acceptance Criteria

- [ ] File brought under 800 lines via extraction or decomposition
- [ ] No behavior changes
- [ ] Tests pass

Issue #1578, opened 2026-08-09
Decompose test_websocket_service.py


NodeJSmith:
## Description

`tests/integration/test_websocket_service.py` is 1200 lines (threshold: 800). Surfaced by `tools/check_file_size.py`, now a visible non-blocking CI job.

## Acceptance Criteria

- [ ] File brought under 800 lines via extraction or decomposition
- [ ] No behavior changes
- [ ] Tests pass

PR #1711, opened 2026-08-28
test: name the retry-count constant and share the change-event lock race scaffold

### LINKED ISSUE #1653: Clean up magic number and duplicated race scaffold in tests/unit/core/
## Description

Surfaced by `/mine-clean-code` (nitpicker) during the decompose-oversized-test-files batch (#1578-#1583). Both findings are pre-existing, not introduced by that branch — the branch only moved this content verbatim into new sibling files (or, for the coverage file, only touched a docstring).

## Key Items

- `tests/unit/core/test_command_executor_pipeline_queue.py:211,217` — `retry_count=3` hardcoded as a bare literal (twice) instead of referencing `_MAX_RETRY_COUNT` from `src/hassette/core/command_executor.py:45`.
- `tests/unit/core/test_app_lifecycle_service_coverage.py:301-349` and `:352-407` — `test_concurrent_invocations_are_serialized_by_the_change_event_lock` and `test_replay_and_handle_change_event_do_not_race` duplicate the same ~25-line gate/first_entered/call_count race-scaffold almost verbatim (event setup, `asyncio.create_task`, `wait_for(... _waiters ...)`, lock-held assertions, `gate.set()` + double `wait_for`), differing only in which coroutine is gated. No `dup-ignore` annotation, unlike similar intentional duplication elsewhere in the same suite.

## Acceptance Criteria

- [ ] `test_command_executor_pipeline_queue.py` references `_MAX_RETRY_COUNT` instead of the literal `3`.
- [ ] The two race-condition tests in `test_app_lifecycle_service_coverage.py` share a common helper (e.g. `assert_serializes_on_change_event_lock(...)`) or are explicitly annotated `# dup-ignore-start`/`# dup-ignore-end` if extraction isn't worth it.

NodeJSmith:
## Summary

Clears the two pre-existing code-quality findings `/mine-clean-code` (nitpicker) surfaced in `tests/unit/core/` during the decompose-oversized-test-files batch (#1578-#1583). Test-only changes — no production code or behavior is touched.

## What changed

**`test_command_executor_pipeline_queue.py` — name the retry ceiling.**
`test_max_retries_drops_batch` hardcoded the exhausted retry count as a bare `3` in two places. Tuning `_MAX_RETRY_COUNT` in `command_executor.py` would have left the test silently asserting a stale threshold while still passing. It now imports and references the constant, so the test tracks the production value.

**`test_app_lifecycle_service_coverage.py` — share the lock-race scaffold.**
The two "must serialize on `_change_event_lock`" tests each carried their own near-verbatim copy of the same ~25-line scaffold: the `gate`/`first_entered` events, the `call_count` bookkeeping, the `wait_for(... _waiters ...)` deterministic queue check, the lock-held assertions, and the `gate.set()` + double `wait_for` teardown. They differed only in which collaborator is gated and which caller holds the lock.

That scaffold now lives in a `ChangeEventLockRace` helper exposing three methods — `gated_call()` (the body the test swaps into its chosen collaborator), `start_first()` (start the lock-holding caller and wait until it parks), and `run_concurrent_call_and_assert_serialized()` (drive the second concurrent call and prove the lock queued it). Each test is left with just the parts that actually differ. The deterministic-waiter approach and the `CLAUDE.md` "Deterministic Async Race Gate" rationale are preserved in the helper's comments rather than duplicated per-test.

Extraction was chosen over the `# dup-ignore-start`/`# dup-ignore-end` escape hatch the issue offered as an alternative, since the duplication was incidental rather than intentional.

## Testing

- `uv run nox -s dev` — **7034 passed**, 1 skipped, 2 xfailed (2:54). Both refactored tests included and passing.
- `prek -a` — all 28 hooks pass (ruff check/format, flake8-async, house-lint, eslint/stylelint/tsc, factory-shadowing check, and the rest).

The refactor is behavior-preserving: the helper keeps the same event ordering, the same deterministic `_waiters` gate, and the same assertions (lock held, `call_count == 1` mid-race, second task not done, `call_count == 2` after release) that the inline versions made.

Closes #1653

### DIFF
--- a/tests/unit/core/test_app_lifecycle_service_coverage.py
+++ b/tests/unit/core/test_app_lifecycle_service_coverage.py
@@ -36,6 +36,10 @@ class ChangeEventLockRace:
     race through the methods below rather than touching them from a test body.
     """
 
+    # Generous upper bound on a race that resolves in microseconds: every await below is already
+    # gated on a deterministic signal, so this only bounds a hang, it never paces the test.
+    WAIT_TIMEOUT_SECONDS = 1.0
+
     def __init__(self, lifecycle_service: AppLifecycleService) -> None:
         self.lifecycle_service = lifecycle_service
         self.gate = asyncio.Event()
@@ -52,7 +56,7 @@ class ChangeEventLockRace:
     async def start_first(self, coro: Coroutine[Any, Any, Any]) -> asyncio.Task[Any]:
         """Start the lock-holding caller and wait until it is parked inside the gated collaborator."""
         task = asyncio.create_task(coro)
-        await asyncio.wait_for(self.first_entered.wait(), timeout=1.0)
+        await asyncio.wait_for(self.first_entered.wait(), timeout=self.WAIT_TIMEOUT_SECONDS)
         return task
 
     async def run_concurrent_call_and_assert_serialized(self, first_task: asyncio.Task[Any]) -> None:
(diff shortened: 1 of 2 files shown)


NodeJSmith:
@coderabbitai review

PR #1712, opened 2026-08-28
test: unprefix CLI client test helpers and parametrize credential-attachment tests

### LINKED ISSUE #1654: Fix naming drift and duplicated test shape in tests/unit/cli/test_client*.py
## Description

Surfaced by `/mine-clean-code` (nitpicker) during the decompose-oversized-test-files batch (#1578-#1583). All findings are pre-existing, not introduced by that branch — the branch only moved `TestCredentialAttachment` and its siblings verbatim into `test_client_credentials.py`.

## Key Items

- `tests/unit/cli/test_client.py:56,66` — `_make_host_port_config` and `_make_manifest_list` are the only two module-level test helpers in this file with a leading underscore; sibling helpers of identical scope (`make_transport`, `url_capturing_client`, `route_listeners`, `get_expecting_exit`, `get_json_error`, `stderr_for_successful_get`) have none. Per this repo's own "No Default Underscore Prefixes" convention, and the fact that `test_client_credentials.py:23` already imports and uses `_make_host_port_config` across files (so the "module-private" signal doesn't even hold), these should be unprefixed.
- `tests/unit/cli/test_client.py:141,305` — `_make_host_port_config("127.0.0.1", 8126)` redundantly passes the function's own default values instead of calling `_make_host_port_config()`.
- `tests/unit/cli/test_client_credentials.py:45-182` (`TestCredentialAttachment`) — 8 of 12 test methods repeat the identical 3-line shape (`factory = CLIClientFactory(make_cli_config(...)); client, captured_headers = factory.build_capturing_headers(); client.get(HEALTH_ENDPOINT, dict)`), varying only the `make_cli_config` kwargs and final assertion. This directory's own `CLAUDE.md` notes `tools/check_duplicate_code.py` flags boilerplate once it appears in three tests; this clears that bar with no `dup-ignore` annotation.

## Acceptance Criteria

- [ ] `_make_host_port_config`/`_make_manifest_list` renamed without leading underscore (update both call sites across `test_client.py` and `test_client_credentials.py`).
- [ ] Redundant explicit-default calls simplified to `_make_host_port_config()` (post-rename: `make_host_port_config()`).
- [ ] `TestCredentialAttachment`'s 8 near-identical tests collapsed into a `pytest.mark.parametrize` table (config kwargs → expected header), or explicitly annotated as intentional if kept separate.

NodeJSmith:
## Summary

Test-only cleanup of `tests/unit/cli/test_client.py` and `tests/unit/cli/test_client_credentials.py`, addressing the three nitpicker findings in #1654. All of these predate the decompose-oversized-test-files batch (#1578–#1583) — that branch only moved the affected classes verbatim.

## What changed

**Naming drift.** `_make_host_port_config` and `_make_manifest_list` were the only underscore-prefixed module-level helpers in `test_client.py`; every sibling of identical scope (`make_transport`, `url_capturing_client`, `get_expecting_exit`, …) has no prefix. The module-private signal never held anyway — `test_client_credentials.py` imports `_make_host_port_config` across files. Both are renamed without the prefix, and `make_manifest_list` picked up the docstring it was missing.

**Redundant explicit defaults.** Two `make_host_port_config("127.0.0.1", 8126)` call sites passed the function's own default values; they now call `make_host_port_config()`.

**Duplicated test shape.** Eight of `TestCredentialAttachment`'s twelve methods repeated the same three-line body, varying only the `make_cli_config` kwargs and the final header assertion. They collapse into one parametrized `(config_token, token_file, expected_header)` table. The no-credential-anywhere case folds into `test_missing_token_never_calls_generating_resolver`, which already asserted the same header absence plus the stronger no-minted-token invariant. The two blank-token rows carry a comment explaining why they are load-bearing rather than filler — they pin the CLI's empty/whitespace handling to `resolve_auth_token()`'s server-side behavior, so the CLI can't resolve a different credential than the service validates against.

**One follow-on extraction.** Simplifying the redundant-default call sites left three connect-error test bodies token-identical, which `tools/check_duplicate_code.py` would then flag as a new cluster. Those are extracted into `stderr_for_connect_error()`, the sibling of the existing `stderr_for_successful_get()`. This also made `test_client_credentials.py`'s `httpx` import unused, so it's dropped.

No production code, no behavior change, no assertion weakened — every previously-asserted condition still has a test asserting it.

## Testing

- `uv run nox -s dev` — 7033 passed, 1 skipped, 2 xfailed
- `prek -a` — all hooks pass (including `check_duplicate_code`, `check_test_factories`, ruff, eslint, tsc)

Closes #1654

### DIFF
--- a/tests/unit/cli/test_client.py
+++ b/tests/unit/cli/test_client.py
@@ -59,7 +59,7 @@ def make_transport(
     return httpx.MockTransport(_fixed_response)
 
 
-def _make_host_port_config(host: str = "127.0.0.1", port: int = 8126) -> HassetteConfig:
+def make_host_port_config(host: str = "127.0.0.1", port: int = 8126) -> HassetteConfig:
     """Bare host/port config for base-URL construction tests.
 
     Narrower than the shared ``make_cli_config()`` (from ``conftest.py``) — no ``cli``/``data_dir``
@@ -69,7 +69,8 @@ def _make_host_port_config(host: str = "127.0.0.1", port: int = 8126) -> Hassett
     return HassetteConfig(token=None, web_api=WebApiConfig(host=host, port=port))
 
 
-def _make_manifest_list(instances: list[AppInstanceResponse], app_key: str = "my_app"):
+def make_manifest_list(instances: list[AppInstanceResponse], app_key: str = "my_app"):
+    """Wrap ``instances`` in a single-app manifest list, as ``/api/apps/manifests`` returns it."""
     manifest = make_manifest_response(app_key=app_key, instance_count=len(instances), instances=instances)
     return make_manifest_list_response(manifests=[manifest])
 
@@ -83,7 +84,7 @@ def url_capturing_client() -> tuple[HassetteCLIClient, list[str]]:
         return httpx.Response(200, content=b"[]", headers={"content-type": "application/json"})
 
     transport = httpx.MockTransport(handler)
-    return HassetteCLIClient(_make_host_port_config(), json_mode=False, transport=transport), captured_urls
+    return HassetteCLIClient(make_host_port_config(), json_mode=False, transport=transport), captured_urls
 
 
 def route_listeners(client: HassetteCLIClient, **kwargs: Any) -> Any:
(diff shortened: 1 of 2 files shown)


NodeJSmith:
@coderabbitai review

PR #1713, opened 2026-08-28
test: reference BATCH_SIZE instead of literals in test_logging_persistence

### LINKED ISSUE #1655: Clean up magic number and helper-style inconsistency in test_logging_persistence.py
## Description

Surfaced by `/mine-clean-code` (nitpicker) during the decompose-oversized-test-files batch (#1578-#1583). Pre-existing findings, not introduced by that branch — the branch only moved this content verbatim out of the former `test_logging.py`.

## Key Items

- `tests/unit/test_logging_persistence.py:39,58,171,189` — `range(50)` / `range(49)` hardcodes `LogPersistenceHandler.BATCH_SIZE` (defined as `50` in `src/hassette/logging_.py:265`) as a bare literal in four places instead of referencing the source constant. If the constant ever changes, these tests silently stop testing the actual threshold boundary instead of failing loudly.
- `tests/unit/test_logging_persistence.py:153-161` — `TestLogPersistenceDropCountWithDB.enqueue_returning_false` / `.enqueue_raising_runtime_error` are `@staticmethod`s invoked via `self.<name>`, while the rest of the same file uses plain module-level helper functions (e.g. `_make_dropping_db_service`). Two different helper-declaration conventions for the same kind of thing in one file.

## Acceptance Criteria

- [ ] The four `range(50)`/`range(49)` call sites reference `LogPersistenceHandler.BATCH_SIZE` instead of a literal.
- [ ] The two staticmethod helpers are converted to plain module-level functions (or the file settles on one convention and documents why the other case differs).

NodeJSmith:
## Summary

Removes the hardcoded `50` / `49` literals from `tests/unit/test_logging_persistence.py` and settles the file on a single helper-declaration convention. Test-only change — no production code or behavior is touched.

## What changed

**Magic number → source constant.** Four call sites used bare `range(50)` / `range(49)` (and matching assertions) to exercise the `LogPersistenceHandler` batch-flush threshold, duplicating `LogPersistenceHandler.BATCH_SIZE` from `src/hassette/logging_.py`. They now reference the constant directly, including the off-by-one "just below threshold" case as `BATCH_SIZE - 1`.

Why this matters: with the literal, changing `BATCH_SIZE` would leave these tests passing while quietly no longer testing the threshold boundary at all — the 49-record test would stop being "one below the flush point" and become an arbitrary count. Referencing the constant makes the tests track the boundary, and fail loudly if the relationship ever breaks.

`test_batch_flushes_at_50_records` is renamed to `test_batch_flushes_at_batch_size_records` so the test name doesn't re-encode the same literal it just stopped hardcoding.

**One helper convention.** `TestLogPersistenceDropCountWithDB` declared two `@staticmethod` helpers invoked via `self.<name>`, while the rest of the file used plain module-level functions (`_make_dropping_db_service`). Both are now module-level `_enqueue_returning_false` / `_enqueue_raising_runtime_error`, and `_make_dropping_db_service` reuses `_enqueue_returning_false` instead of redefining the same logic as a local closure. Both helpers gained docstrings explaining which failure mode they simulate (full queue vs. shut-down service).

## Testing

- `prek -a` — all hooks pass

Closes #1655

### DIFF
--- a/tests/unit/test_logging_persistence.py
+++ b/tests/unit/test_logging_persistence.py
@@ -9,11 +9,11 @@ import asyncio
 import logging
 import queue
 import time
-from collections.abc import Coroutine
+from collections.abc import Callable, Coroutine
 from typing import Any
 from unittest.mock import MagicMock
 
-from hassette.logging_ import HassetteQueueListener, LogPersistenceHandler
+from hassette.logging_ import DEQUEUE_TIMEOUT_SECONDS, HassetteQueueListener, LogPersistenceHandler
 
 
 def _enqueue_returning_false(coro: Coroutine[Any, Any, Any]) -> bool:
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #1726, opened 2026-08-29
test: exercise real app discovery in autodetect and validate-apps tests

### LINKED ISSUE #1658: Improve test rigor in autodetect and validate-apps tests
## Description

CodeRabbit review on PR #1656 (test-file decomposition) surfaced two pre-existing test-quality gaps in `tests/unit/test_autodetect_apps_integration.py` and `tests/unit/test_validate_apps.py` — both predate this PR (relocated verbatim from the former `test_autodetect_apps.py`), so they're tracked here instead of being fixed in that decompose-only PR.

## Key Items

- `test_autodetect_apps_integration.py:53-77` (`test_hassette_config_autodetect_can_be_disabled`): the test never calls `config.set_validated_app_manifests()`, which is where `autodetect` is actually evaluated (`HassetteConfig.set_validated_app_manifests()`). It asserts on `config.apps.apps` (never populated regardless), so the test would still pass even if validation incorrectly discovered applications.
- `test_validate_apps.py:123-168,170-215` (`test_validate_apps_calls_autodetect`, `test_validate_apps_skips_conflicting_autodetected`): both tests patch the internal `hassette.config.config.autodetect_apps` collaborator directly rather than exercising real app discovery through temp files, bypassing the actual discovery contract that `set_validated_app_manifests()` must preserve.

## Acceptance Criteria

- [ ] `test_hassette_config_autodetect_can_be_disabled` calls `config.set_validated_app_manifests()` before asserting, and asserts `config.apps.manifests` is empty.
- [ ] `test_validate_apps_calls_autodetect` and `test_validate_apps_skips_conflicting_autodetected` create real temporary app files under the test app directory instead of patching `autodetect_apps`, and assert on the resulting manifests/conflict behavior.
- [ ] Any call-argument coverage currently obtained by mocking `autodetect_apps` is preserved only in integration tests that exercise the real discovery path (not unit tests).

NodeJSmith:
## Summary

Three tests in the app auto-detection suite asserted on things that could not fail. This
replaces the mocks and the wrong assertion target with real app discovery, so the tests now
exercise the contract `HassetteConfig.set_validated_app_manifests()` actually has to preserve.

## What changed

**`test_hassette_config_autodetect_can_be_disabled`** (`tests/unit/test_autodetect_apps_integration.py`)

The test never called `set_validated_app_manifests()` — the one place `autodetect` is
evaluated — and asserted on `config.apps.apps`, which is never populated by discovery either
way. It passed whether or not discovery ran. It now calls that method and asserts
`config.apps.manifests` is empty.

**`test_validate_apps_calls_autodetect`** and **`test_validate_apps_skips_conflicting_autodetected`**
(`tests/unit/test_validate_apps.py`)

Both patched `hassette.config.config.autodetect_apps` and asserted against the mock's return
value, so the real discovery path was never touched. Both now write real app modules under the
configured app directory and assert on the resulting manifests.

The call-argument coverage the mocks provided is preserved behaviorally rather than by
inspecting call args:

- `test_validate_apps_calls_autodetect` asserts the manually configured file is *not* detected
  a second time under its derived key — which is only true if its resolved path was passed to
  discovery in `known_paths`.
- `test_validate_apps_skips_conflicting_autodetected` points its manual entry at a *different*
  file than the one discovery finds. Because a manually configured file is excluded from
  discovery via `known_paths`, aiming both at the same file could never reach the key-conflict
  branch; pointing them at different files is the only way to hit it through real discovery.

**`write_app` hoisted into `hassette.test_utils.helpers`**

Both test modules needed it, so the local copy in `tests/unit/test_autodetect_apps.py` moved to
the shared helpers module per the repo's test-factory convention.

## Testing

- `prek -a` — all hooks pass

No source behavior changed; this is test-only.

Closes #1658

### DIFF
--- a/src/hassette/test_utils/__init__.py
+++ b/src/hassette/test_utils/__init__.py
@@ -65,6 +65,7 @@ from .helpers import make_task_bucket as make_task_bucket
 from .helpers import patch_loop_getaddrinfo as patch_loop_getaddrinfo
 from .helpers import wire_up_app_running_listener as wire_up_app_running_listener
 from .helpers import wire_up_app_state_listener as wire_up_app_state_listener
+from .helpers import write_app as write_app
 from .helpers import write_app_toml as write_app_toml
 from .helpers import write_test_app_with_decorator as write_test_app_with_decorator
 from .mock_hassette import make_mock_hassette as make_mock_hassette
--- a/src/hassette/test_utils/helpers.py
+++ b/src/hassette/test_utils/helpers.py
@@ -454,7 +454,7 @@ def write_app(app_dir: Path, filename: str, source: str) -> Path:
     """Write `source` as a module inside `app_dir`, creating the directory if needed.
 
     For app files whose exact source matters (auto-detect discovery, import behavior). Use
-    :func:`write_test_app_with_decorator` instead when a canned App subclass is enough.
+    `write_test_app_with_decorator` instead when a canned App subclass is enough.
     """
     app_dir.mkdir(parents=True, exist_ok=True)
     path = app_dir / filename
--- a/src/hassette/test_utils/helpers.py
+++ b/src/hassette/test_utils/helpers.py
@@ -450,6 +450,18 @@ def write_app_toml(
         tomli_w.dump(toml_dict, f)
 
 
+def write_app(app_dir: Path, filename: str, source: str) -> Path:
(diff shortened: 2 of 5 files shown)


NodeJSmith:
@coderabbitai review

PR #1729, opened 2026-08-29
test: extract repeated literals and use path aliases in execution-detail.test.tsx

### LINKED ISSUE #1662: Clean up duplicated literals and minor style nits in execution-detail.test.tsx
Automated quality scan of this test file (llm-checker, lazy-checker, nitpicker) found no correctness issues, but several duplicated literals and minor style inconsistencies worth cleaning up in one pass.

## Findings

**1. Repeated hardcoded literals with no shared constants**

Several literal values are hand-typed many times throughout the file instead of being hoisted into named constants at the top:

- The UUID `"abc12345-1234-5678-9abc-def012345678"` is inlined ~12 times as the `execution_id` passed to `createExecution` (lines 32, 38, 47, 57, 67, 79, 89, 99, 110, 119, 129, 141, 181).
- The literal `"abc-123"` (execution ID used in the `ExecutionDetailFetcher` describe block) is repeated 5 times (lines 200, 214, 228, 236, 247).
- The string `"handler"` (first positional arg to `createExecution`) is repeated 17 times with no named constant.
- The traceback string `"Traceback (most recent call last):\n  File ...\nValueError: bad input"` is duplicated verbatim in two tests (lines 71, 81).
- `duration_ms: 42` appears as an unnamed literal in two tests (lines 59, 238).

A single typo in any of these copies would silently desync it from the rest without failing anything else — e.g., the `truncateId` assertion on line 34 depends on the UUID literal matching exactly.

The file already demonstrates the fix pattern in three places (`const uuid = "..."` at lines 38, 150, 169) but doesn't apply it consistently — most occurrences are typed inline instead. Hoisting these into a handful of module-level constants (e.g. `TEST_EXECUTION_ID`, `TEST_ABC_ID`, `HANDLER_KIND`, `SAMPLE_TRACEBACK`) would remove the duplication in one pass and make the inline-vs-bound usage consistent throughout the file.

**Fix:** Add named constants near the top of the file (after imports) for the UUID, `"abc-123"`, `"handler"`, and the sample traceback string; replace all inline occurrences with references to them.

**2. Abbreviated variable name breaks the file's naming convention**

Line 173: `const btn = container.querySelector(...)` uses the abbreviation `btn`, while every other rendered-element handle in the file uses full descriptive names (`container`, `code`, `record`, or destructured `getByRole`/`getByTestId`/`findByText`).

**Fix:** Rename `btn` to a descriptive name consistent with the rest of the file (e.g. `copyButton`).

**3. Relative imports instead of the configured path alias**

Lines 8-10 import via `../../test/...` (`createExecution`, `createWouterMock`, `createTestQueryClient`) instead of the `@/*` path alias configured in `tsconfig.json`/`vite.config.ts`, which sibling non-test files in this same directory (`app-detail-header.tsx`, `code-tab.tsx`, `config-tab.tsx`) use routinely.

**Fix:** Switch to `@/test/factories`, `@/test/mock-wouter`, `@/test/query-test-utils`.

NodeJSmith:
## Summary

Cleans up duplicated literals and style inconsistencies in
`frontend/src/components/app-detail/execution-detail.test.tsx`, as reported by the automated
quality scan in #1662. Test-only change — no production code touched, no behavior change.

## What changed

**Hoisted repeated literals into module-level constants** (after the imports):

- `TEST_EXECUTION_ID` — the `"abc12345-1234-5678-9abc-def012345678"` UUID, previously inlined
  ~12 times as `execution_id`. The `truncateId` heading assertion depends on this literal
  matching exactly, so a typo in any one copy could have silently desynced it.
- `FETCHER_EXECUTION_ID` — the `"abc-123"` ID used across the `ExecutionDetailFetcher` block
  (5 occurrences).
- `HANDLER_KIND` — the `"handler"` first positional arg to `createExecution` (17 occurrences).
- `SAMPLE_TRACEBACK` — the sample traceback string, previously duplicated verbatim in two tests.

This also makes the file internally consistent: it already used the `const uuid = "..."` pattern
in three places but typed the value inline everywhere else. Those three local bindings are now
folded into the shared constants.

**Renamed `btn` to `copyButton`** in the clipboard test, matching the descriptive naming used for
every other element handle in the file.

**Switched relative imports to the `@/*` path alias** — `@/test/factories`, `@/test/mock-wouter`,
`@/test/query-test-utils`, and `@/test/server` instead of `../../test/...`. Sibling non-test files
in this directory (`app-detail-header.tsx`, `code-tab.tsx`, `config-tab.tsx`) already use the alias.

The one finding not addressed is the `duration_ms: 42` literal noted in the issue. It appears twice
in unrelated tests where the value is incidental to the assertion, so naming it would add
indirection without removing a real coupling.

## Testing

- `npx vitest run src/components/app-detail/execution-detail.test.tsx` — 21 passed
- `npx vitest run` (full frontend suite) — 107 files, 1548 passed

## Note on draft status

This is opened as a draft only because the change reached the worktree uncommitted rather than as
a reviewed commit — not because anything is failing. All verification above is green. Mark it ready
once you've eyeballed the diff.

Closes #1662

### DIFF
--- a/frontend/src/components/app-detail/execution-detail.test.tsx
+++ b/frontend/src/components/app-detail/execution-detail.test.tsx
@@ -5,11 +5,17 @@ import { http, HttpResponse } from "msw";
 import type { ReactNode } from "react";
 import { describe, expect, it, vi } from "vitest";
 
-import { createExecution } from "../../test/factories";
-import { createWouterMock } from "../../test/mock-wouter";
-import { createTestQueryClient } from "../../test/query-test-utils";
-import { server } from "../../test/server";
+import { createExecution } from "@/test/factories";
+import { createWouterMock } from "@/test/mock-wouter";
+import { createTestQueryClient } from "@/test/query-test-utils";
+import { server } from "@/test/server";
+
 import { ExecutionDetailContent, ExecutionDetailFetcher } from "./execution-detail";
+
+const HANDLER_KIND = "handler";
+const TEST_EXECUTION_ID = "abc12345-1234-5678-9abc-def012345678";
+const FETCHER_EXECUTION_ID = "abc-123";
+const SAMPLE_TRACEBACK = "Traceback (most recent call last):\n  File ...\nValueError: bad input";
 
 vi.mock("wouter", () => createWouterMock());
 
@@ -29,22 +35,21 @@ function Wrapper({ children }: { children: ReactNode }) {
 
 describe("ExecutionDetailContent", () => {
   it("renders truncated execution ID in heading", () => {
-    const record = createExecution("handler", { execution_id: "abc12345-1234-5678-9abc-def012345678" });
+    const record = createExecution(HANDLER_KIND, { execution_id: TEST_EXECUTION_ID });
     const { getByRole } = render(<ExecutionDetailContent record={record} />);
     expect(getByRole("heading").textContent).toContain("12345678");
   });
 
   it("renders full execution ID in code element", () => {
-    const uuid = "abc12345-1234-5678-9abc-def012345678";
-    const record = createExecution("handler", { execution_id: uuid });
+    const record = createExecution(HANDLER_KIND, { execution_id: TEST_EXECUTION_ID });
     const { container } = render(<ExecutionDetailContent record={record} />);
     const code = container.querySelector("code");
-    expect(code?.textContent).toBe(uuid);
+    expect(code?.textContent).toBe(TEST_EXECUTION_ID);
   });
 
   it("renders meta stats with duration, timestamp, and status", () => {
-    const record = createExecution("handler", {
-      execution_id: "abc12345-1234-5678-9abc-def012345678",
+    const record = createExecution(HANDLER_KIND, {
+      execution_id: TEST_EXECUTION_ID,
       duration_ms: 150,
       status: "success",
     });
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
Review pass (code / integration / WTF / llm / lazy / nitpicker) — three changes applied in b2f57f73, plus notes on what was deliberately not changed.

**Dropped `HANDLER_KIND`.** Issue #1662 asked for it, but the constant's name is its value: every call site traded a self-describing `createExecution("handler", …)` for a lookup at the top of the file. Inlining it back also makes the file internally coherent — the same pass left the status discriminants (`"success"`, `"error"`, `"timed_out"`) inline for exactly that reason, as did `duration_ms: 42` per the PR body. `TEST_EXECUTION_ID`, `FETCHER_EXECUTION_ID`, and `SAMPLE_TRACEBACK` all stay — those dedupe real repeated literals where a typo could silently desync a copy.

**Derived the truncated-heading assertion.** `toContain("12345678")` restated the tail of `TEST_EXECUTION_ID` as its own literal. Hoisting the ID into a constant 30 lines up made that coupling *harder* to see than when the two literals sat one line apart, so it now reads `toContain(TEST_EXECUTION_ID.slice(-8))`, matching `truncateId()`'s `"…" + id.slice(-8)`.

**Finished the alias conversion.** The two `vi.mock` specifiers were still relative while every other local import had moved to `@/`. Reviewers split on this — one argued no sibling test file aliases a `vi.mock` target, the other that mixed addressing inside one file undercuts the PR's own title. Went with internal consistency after verifying empirically that vitest resolves `@/` inside `vi.mock` (suite green: 21/21 in this file, 1548/1548 overall).

**Not changed:** this file is now one of ~7 test files using `@/test/*` against ~21 still on relative imports, so it's ahead of its same-depth siblings. That drift is real but pre-dates this PR and was what #1662 asked for; filing a follow-up rather than reverting. `execution-table.test.tsx` has the same duplicated-literal pattern and is untouched — also follow-up material.

PR #1738, opened 2026-08-30
test: collapse repeated render blocks in execution-section tests

### LINKED ISSUE #1674: Clean up repeated render blocks in execution-section.test.tsx
## Description

A code-quality scan of `frontend/src/components/app-detail/execution-section.test.tsx` surfaced one surviving finding (plus one minor naming nit). This predates the current work — the file is not being modified by any in-flight branch.

Two of three checkers (`llm-checker`, `lazy-checker`) returned CLEAN. The findings below are what survived validity assessment from the `nitpicker` pass.

### 1. Four near-identical `render()` blocks repeat the same four literal props, diverging from the sibling test convention

`execution-section.test.tsx:36-64` — the first four tests each render:

```tsx
<ExecutionSection heading="Executions" records={...} kind="handler" tableId="table-1" loading={...} />
```

Only `records` and `loading` vary; `heading="Executions"`, `kind="handler"`, and `tableId="table-1"` are re-typed at every call site. The result is that the prop actually under test is visually buried in a line of constant noise.

Sibling test files in the same directory already establish a local render-helper pattern for exactly this shape:

- `handler-health-card.test.tsx:25` — `function renderCard(item, { appKey = "test_app", instanceQs = "" } = {})`
- `overview-tab.test.tsx:31` — `function renderOverviewTab({ ... })`

**Proposed fix:** add a local `renderSection(overrides: Partial<ComponentProps<typeof ExecutionSection>>)` helper with the invariant props as defaults, collapsing the four blocks into one-line calls where only the varying prop is named. This subsumes the repeated `"Executions"` / `"table-1"` / `"handler"` literals without hoisting them into standalone constants.

Note: keep the `toBe("Executions")` assertion at line 39 comparing against a literal, not against the helper's default — sharing one constant between the render input and the assertion would make that test tautological.

The fifth test (`"passes the correct props to ExecutionTable"`, lines 66-95) deliberately overrides most props and should keep its explicit inline form.

### 2. Minor: `RECORD` fixture name restates its type

`execution-section.test.tsx:21` — `const RECORD: ExecutionRecord` is named after its TypeScript type and communicates nothing at its five call sites.

**Proposed fix:** rename to something like `SAMPLE_EXECUTION` or `OK_EXECUTION` (the fixture is a `status: "ok"`, non-leaked record).

## Acceptance Criteria

- [ ] A local render helper with default props exists in `execution-section.test.tsx`, following the `renderCard` / `renderOverviewTab` pattern used by sibling tests in `app-detail/`
- [ ] The four tests at lines 35-64 use the helper and name only the props under test
- [ ] The heading assertion still compares rendered output against a literal, not against the helper's default value
- [ ] The `RECORD` fixture is renamed to something that describes the fixture rather than its type
- [ ] All existing behavior coverage is preserved — the same five branches (heading render, loading spinner, empty-array table, populated table, prop pass-through) remain verified
- [ ] `npx vitest run src/components/app-detail/execution-section.test.tsx` passes

## Not filed (assessed as invalid)

- **Magic numbers `execution_start_ts: 1_700_000_000` and `duration_ms: 12` (lines 22-23)** — claimed to need named constants. In fact the mocked `ExecutionTable` (lines 11-15) reads only `records.length` and `tableId`, and no test asserts on either field; they exist solely to satisfy the `ExecutionRecord` type. Naming them would add noise without informing any assertion.

NodeJSmith:
## Summary

Collapses four near-identical `render()` blocks in `execution-section.test.tsx` behind a local
`renderSection()` helper, so each test names only the prop it is actually exercising instead of
re-typing the three invariant props at every call site.

Before, every one of the first four tests spelled out `heading="Executions"`, `kind="handler"`,
and `tableId="table-1"` alongside the two props under test — the varying prop was buried in
constant noise. The helper takes `Partial<ComponentProps<typeof ExecutionSection>>` overrides
with those invariants as defaults, following the `renderCard` / `renderOverviewTab` pattern
already established by sibling tests in `app-detail/`.

Also renames the `RECORD` fixture to `OK_EXECUTION`, which describes the fixture (a successful,
non-leaked execution) rather than restating its TypeScript type.

## Notes

- The heading test still asserts against the `"Executions"` string literal rather than the
  helper's default value — sharing one constant between the render input and the assertion would
  make that test tautological.
- The fifth test (`"passes the correct props to ExecutionTable"`) deliberately overrides most
  props, so it keeps its explicit inline form.
- No behavior change: all five branches (heading render, loading spinner, empty-array table,
  populated table, prop pass-through) remain verified.

## Testing

- `npx vitest run src/components/app-detail/execution-section.test.tsx` — 5 passed
- `npm run test` (full frontend suite) — 108 files, 1555 tests passed

Closes #1674

### DIFF
--- a/frontend/src/components/app-detail/execution-section.test.tsx
+++ b/frontend/src/components/app-detail/execution-section.test.tsx
@@ -1,4 +1,5 @@
 import { render } from "@testing-library/react";
+import type { ComponentProps } from "react";
 import { afterEach, describe, expect, it, vi } from "vitest";
 
 import type { ExecutionRecord } from "../shared/execution-table";
@@ -18,7 +19,7 @@ vi.mock("../shared/execution-table", async () => {
 
 const { ExecutionTable } = await import("../shared/execution-table");
 
-const RECORD: ExecutionRecord = {
+const OK_EXECUTION: ExecutionRecord = {
   execution_start_ts: 1_700_000_000,
   duration_ms: 12,
   status: "success",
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #1740, opened 2026-08-30
test: dedupe literals and setup in handler-health-card tests

### LINKED ISSUE #1701: Clean up repeated test-id/route literals and duplicated setup in handler-health-card.test.tsx
Surfaced by the automated quality scanner on `frontend/src/components/app-detail/handler-health-card.test.tsx`. All findings predate the scan; none are correctness bugs in the component under test.

## Finding 1 — Hand-typed `data-testid` and route literals repeated across 15+ sites

The test-id pattern `overview-health-card-<kind>-<id>` is typed out as a raw string in 9 places (lines 187, 199, 213, 226, 257, 263, 270, 279, 285), and the route pattern `/apps/<appKey>/handlers/<kind>/<id>` is reconstructed by hand 4 more times (lines 190, 202, 217, 230). The literal `"my_app"` appears as an `appKey` override in 4 tests (185, 197, 211, 224).

This is the finding most likely to cause a real defect rather than just noise: a typo in one `getByTestId` string makes that assertion fail loudly, but a typo in a `toHaveBeenCalledWith` route string makes the test assert against a URL the component never produces — and since the ids differ per test, there is no cross-check.

**Proposed fix:** add two small local helpers plus one constant near the existing `makeListenerItem`/`makeJobItem` factories:

```ts
const APP_KEY = "my_app";
const cardTestId = (kind: "listener" | "job", id: number) => `overview-health-card-${kind}-${id}`;
const handlerRoute = (kind: "listener" | "job", id: number) => `/apps/${APP_KEY}/handlers/${kind}/${id}`;
```

and use them at every call site.

## Finding 2 — Navigation-test setup duplicated verbatim four times

Lines 182-183, 194-195, 208-209, and 221-222 each repeat the identical pair:

```ts
const user = userEvent.setup();
mockNavigate.mockClear();
```

`clearMocks` is not enabled in `frontend/vitest.config.ts`, so the `mockClear()` is genuinely required — it is not dead code. But it belongs in a `beforeEach` shared by the navigation describes rather than being restated per test.

**Proposed fix:** hoist `mockNavigate.mockClear()` into a `beforeEach` covering the navigation tests, and either keep `userEvent.setup()` per test or return it from a small `setupNavCard(item)` helper that also does the render and `getByTestId` lookup.

## Finding 3 — Single-`it` `describe` wrappers whose title restates the test name

Eight of the file's 13 `describe` blocks contain exactly one `it`, and in each case the two titles say the same thing — e.g. `describe("HandlerHealthCard — name is not a link")` wrapping `it("renders handler name as a span, not an anchor")` (lines 234-242), and the same shape at lines 29, 47, 65, 80, 95, 112, 128, 165. The nesting layer groups nothing.

(The remaining five describes — avg duration, card click navigation, Enter key navigation, accessibility, data-testid — hold two or more tests each and are doing real grouping work. Leave those alone.)

**Proposed fix:** collapse the eight single-test describes into top-level `it()` calls with the descriptive name kept, or merge related ones into the existing multi-test describes where they belong (e.g. the failing/timed-out/no-error cases form one natural "error display" group).

## Finding 4 — Comment restates the assertions below it

Line 106's `// No error content should appear` says nothing the two `not.toContain` assertions on lines 107-108 do not already say.

**Proposed fix:** delete the comment.

## Acceptance Criteria

- [ ] `data-testid` and handler-route strings are built from shared helpers rather than hand-typed at each call site
- [ ] `"my_app"` is a single named constant
- [ ] `mockNavigate.mockClear()` lives in a `beforeEach` rather than being repeated in each navigation test
- [ ] Single-`it` `describe` wrappers are collapsed or merged into the multi-test groups; multi-test describes are left intact
- [ ] The redundant comment on line 106 is removed
- [ ] `npx vitest run src/components/app-detail/handler-health-card.test.tsx` passes with the same set of test names' coverage intact

## Assessed and excluded

One reviewer finding was dropped as not applicable: a suggestion to replace the `../../test/...` relative imports (lines 4-6) with the `@/test/...` alias. The `@` alias does exist (`vitest.config.ts`, `tsconfig`), but the relative climb is the dominant convention across the frontend test suite (20+ files) versus 7 using the alias. Changing it in this one file would fragment the convention further, not fix it; a suite-wide migration would be its own issue.

NodeJSmith:
## Summary

Removes repeated string literals and duplicated setup from `handler-health-card.test.tsx`. Test-only change — no production code touched, no behavior change.

The riskiest item was the hand-typed route strings. A typo in a `getByTestId` string fails loudly, but a typo in a `toHaveBeenCalledWith` route string silently asserts against a URL the component never produces, and since the ids differ per test there was no cross-check to catch it.

## What changed

**Shared helpers replace hand-typed literals.** The `overview-health-card-<kind>-<id>` test-id pattern was typed out in 9 places and the `/apps/<appKey>/handlers/<kind>/<id>` route reconstructed by hand in 4 more. Both now come from `cardTestId()` and `handlerRoute()`, with `"my_app"` hoisted to a single `APP_KEY` constant.

**`mockNavigate.mockClear()` moved into a `beforeEach`.** It was repeated verbatim in four navigation tests. It is genuinely required — `clearMocks` is not enabled in `vitest.config.ts` — so it was hoisted rather than deleted.

**Eight single-`it` `describe` wrappers collapsed.** Each restated its one test's name, so the nesting layer grouped nothing. Four became top-level `it()` calls; the five error-related ones merged into a single `HandlerHealthCard — error display` group. The six genuinely multi-test describes are untouched.

Two tests were renamed as a consequence: both listener and job variants were named `"renders handler name, kind chip, and run count"` and relied on their now-removed `describe` wrappers to tell them apart. They are now `"renders listener name..."` and `"renders job name..."`.

**Deleted a comment** that restated the two `not.toContain` assertions below it.

## Testing

- `npx vitest run src/components/app-detail/handler-health-card.test.tsx` — 21 passed
- Full frontend suite — 108 files, 1555 tests passed
- `prek -a` — all hooks passed; `prek pyright -a --stage pre-push` — passed

Test count is unchanged at 21, and a diff of the extracted `it()` names against `main` shows the two renames above as the only difference, so coverage is intact.

Closes #1701

### DIFF
--- a/frontend/src/components/app-detail/handler-health-card.test.tsx
+++ b/frontend/src/components/app-detail/handler-health-card.test.tsx
@@ -161,17 +161,19 @@ describe("HandlerHealthCard — avg duration", () => {
   });
 });
 
-it("omits last active when timestamp is null", () => {
-  const item = makeListenerItem({
-    listener_id: 1,
-    last_invoked_at: null,
-    failed: 0,
-    timed_out: 0,
-  });
-  const { container } = renderCard(item);
+describe("HandlerHealthCard — last active", () => {
+  it("omits last active when timestamp is null", () => {
+    const item = makeListenerItem({
+      listener_id: 1,
+      last_invoked_at: null,
+      failed: 0,
+      timed_out: 0,
+    });
+    const { container } = renderCard(item);
 
-  expect(container.textContent).not.toContain("ago");
-  expect(container.textContent).not.toContain("—");
+    expect(container.textContent).not.toContain("ago");
+    expect(container.textContent).not.toContain("—");
+  });
 });
 
 describe("HandlerHealthCard — card click navigation", () => {
@@ -224,15 +226,15 @@ describe("HandlerHealthCard — Enter key navigation", () => {
   });
 });
 
-it("renders handler name as a span, not an anchor", () => {
-  const item = makeListenerItem({ listener_id: 6 });
-  const { container } = renderCard(item);
-
-  expect(container.querySelector("a")).toBeNull();
-  expect(container.textContent).toContain("on_state_change");
-});
-
 describe("HandlerHealthCard — accessibility", () => {
+  it("renders handler name as a span, not an anchor", () => {
+    const item = makeListenerItem({ listener_id: 6 });
+    const { container } = renderCard(item);
+
+    expect(container.querySelector("a")).toBeNull();
+    expect(container.textContent).toContain("on_state_change");
+  });
+
   it("has role=button and aria-label for screen readers", () => {
     const item = makeListenerItem({ listener_id: 1, handler_method: "on_motion" });
     const { getByRole } = renderCard(item);
--- a/frontend/src/components/app-detail/handler-health-card.test.tsx
+++ b/frontend/src/components/app-detail/handler-health-card.test.tsx
@@ -1,5 +1,5 @@
 import userEvent from "@testing-library/user-event";
-import { describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import { createJob, createListener } from "../../test/factories";
 import { createWouterMock } from "../../test/mock-wouter";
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #1742, opened 2026-08-30
fix: keep non-active handler health cards out of the tab order

### LINKED ISSUE #1709: Clean up handler-health-card tooltip tabindex, duplicated stat-row class, and failed count naming
Quality scan of `frontend/src/components/app-detail/handler-health-card.tsx` (llm-checker, lazy-checker, nitpicker). Both the LLM-bias and deferred-debt passes came back clean; three hygiene findings survived verification against the source.

## 1. `StatTooltip` hardcodes `tabIndex={0}`, defeating the card's roving tabindex

**What's wrong:** `StatTooltip` (line 31-42) renders its trigger as `<span className={className} tabIndex={0}>` — a literal `0` for every caller. Meanwhile the card itself takes a roving `tabIndex: 0 | -1` prop (line 28) and applies it at line 76, and participates in `useRovingTabIndex` via `data-roving-item` (line 77). The whole point of that hook (`src/hooks/use-roving-tab-index.ts`) is that exactly one item in the grid is tab-reachable.

Each card renders between 1 and 4 `StatTooltip`s (lines 101, 108, 112, 120, 125), each an unconditional tab stop. So a grid of N cards produces roughly 4N tab stops instead of N — and a card explicitly marked `tabIndex={-1}` still contains focusable descendants. `handler-health-card.test.tsx:252-257` asserts only the card element's own tabindex, so this gap is untested.

**Proposed fix:** thread the card's `tabIndex` prop into `StatTooltip` (add a `tabIndex: 0 | -1` param, default it or pass explicitly at all call sites) so a non-active card's tooltip triggers get `-1`. Keep the trigger focusable on the active card so the tooltip content stays keyboard-reachable. Add a test asserting inner tooltip spans are `-1` when the card is `-1`.

## 2. Duplicated stat-row class string and near-identical row structure

**What's wrong:** lines 107 and 118 carry the byte-identical class string `"flex gap-3 font-mono text-[length:var(--text-mono-sm)] text-muted-foreground"`, and the two `<div>` blocks they head (106-116, 117-130) share the same shape: a wrapper row containing a leading `StatTooltip` and a trailing `ml-auto` `StatTooltip`. Nothing keeps the two class strings in sync if one is restyled.

**Proposed fix:** hoist the shared class into a module-level constant, or extract a small `StatRow` sub-component that owns the wrapper class and takes leading/trailing children. This also shortens the render body.

## 3. `failed` is named like a boolean but holds a count

**What's wrong:** `const failed = item.data.failed;` (line 56) sits two lines below `const failing = isFailing(item);` (line 47), which *is* a boolean. `failed` is a number, used as `failed > 0` (lines 117, 119) and passed to `formatRate(failed, runCount)` (line 121). Two near-identical names with different types next to each other invite a misread.

**Proposed fix:** rename to `failedCount` (3 references, all in this file).

## Excluded after verification

- **`color-mix(... 40% ...)` on line 70 flagged as a magic ratio.** Inline `color-mix` percentage literals in arbitrary Tailwind values are the established convention here, not a one-off: `button.tsx:22,24,27`, `alert-shell.tsx:13`, `handlers-rows.tsx:88`, `log-table-view.tsx:90`, `unified-handler-row.tsx:115`, `error-spotlight.tsx:26`. Naming this one ratio would make it the outlier.
- **Mixed `@/` alias and `../../` relative imports (lines 4-6 vs 8-11).** Claimed to be a deviation from project convention, but the project has no single convention to deviate from — 77 files under `frontend/src` use `@/`, 69 use `../../`. This file is not an outlier; changing it belongs in a repo-wide import-style decision, not a per-file cleanup.
- **`HandlerHealthCard` body exceeding a 50-line function guideline.** The body is a JSX render tree, and at 135 total lines the file is on the small end for this directory (`handlers-tab.tsx` 214, `config-tab.tsx` 231, `execution-detail.tsx` 232, `job-detail.tsx` 266). Finding 2's extraction addresses the real duplication; a split motivated only by line count would add indirection without reducing what a reader has to trace.

NodeJSmith:
## Summary

`HandlerHealthCard` participates in a roving tabindex (via `useRovingTabIndex` and the `tabIndex: 0 | -1` prop), so exactly one card in the grid should be tab-reachable. But its inner `StatTooltip` triggers hardcoded `tabIndex={0}`, so every tooltip in every card was an unconditional tab stop. A grid of N cards produced roughly 4N tab stops instead of N, and a card explicitly marked `tabIndex={-1}` still contained focusable descendants.

This threads the card's `tabIndex` through to `StatTooltip` at all five call sites. A non-active card now contributes no tab stops of its own; the active card keeps its tooltip triggers keyboard-reachable so the tooltip content is still available without a mouse.

Also picks up the two hygiene findings filed alongside it in #1709:

- Hoisted the byte-identical stat-row class string (duplicated across the two stat rows) into a module-level `STAT_ROW_CLASS` constant, so a restyle can't drift the two rows apart.
- Renamed `failed` to `failedCount`. It holds a number but sat two lines below `failing`, an actual boolean — two near-identical names with different types invited a misread.

## Testing

- Two new tests in `handler-health-card.test.tsx`: one asserts every `[tabindex]` element inside a `tabIndex={-1}` card is `-1`, the other asserts they are all `0` on the active card. The existing coverage only checked the card element's own tabindex, which is why this gap went unnoticed.
- `uv run nox -s dev` — 7034 passed, 1 skipped, 2 xfailed.
- `npx vitest run src/components/app-detail/handler-health-card.test.tsx` — 23 passed.
- `prek -a` and `prek -a --hook-stage pre-push` (includes pyright, tsc, eslint) — all hooks pass.

No visual change: the diff only alters `tabindex` attribute values and refactors a class string into a constant with an identical value.

Closes #1709

### DIFF
--- a/frontend/src/components/app-detail/handler-health-card.test.tsx
+++ b/frontend/src/components/app-detail/handler-health-card.test.tsx
@@ -257,6 +257,46 @@ describe("HandlerHealthCard — accessibility", () => {
     expect(getByTestId("overview-health-card-listener-1").getAttribute("tabindex")).toBe("-1");
   });
 
+  it("gives tooltip triggers the card's tabIndex so a non-active card adds no tab stops", () => {
+    const item = makeListenerItem({
+      listener_id: 1,
+      total_invocations: 10,
+      failed: 3,
+      avg_duration_ms: 12,
+      last_invoked_at: 1767225600,
+      last_error_message: "boom",
+      last_error_type: "ValueError",
+    });
+    const { getByTestId } = renderWithAppState(
+      <HandlerHealthCard item={item} appKey="test_app" instanceQs="" tabIndex={-1} />,
+    );
+    const card = getByTestId("overview-health-card-listener-1");
+    const inner = Array.from(card.querySelectorAll("[tabindex]"));
+
+    expect(inner.length).toBeGreaterThan(1);
+    for (const el of inner) {
+      expect(el.getAttribute("tabindex")).toBe("-1");
+    }
+  });
+
+  it("keeps tooltip triggers tab-reachable on the active card", () => {
+    const item = makeListenerItem({
+      listener_id: 1,
+      total_invocations: 10,
+      failed: 3,
+      avg_duration_ms: 12,
+      last_invoked_at: 1767225600,
+    });
+    const { getByTestId } = renderCard(item);
+    const card = getByTestId("overview-health-card-listener-1");
+    const inner = Array.from(card.querySelectorAll("[tabindex]"));
+
+    expect(inner.length).toBeGreaterThan(1);
+    for (const el of inner) {
+      expect(el.getAttribute("tabindex")).toBe("0");
+    }
+  });
+
   it("includes a visible focus outline utility", () => {
     const item = makeListenerItem({ listener_id: 1 });
     const { getByTestId } = renderCard(item);
--- a/frontend/src/components/app-detail/handler-health-card.tsx
+++ b/frontend/src/components/app-detail/handler-health-card.tsx
@@ -21,14 +21,14 @@ import {
 } from "./overview-tab-helpers";
 import type { UnifiedItem } from "./unified-handler-row";
 
+const STAT_ROW_CLASS = "flex gap-3 font-mono text-[length:var(--text-mono-sm)] text-muted-foreground";
+
 interface HandlerHealthCardProps {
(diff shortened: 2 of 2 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #1868, opened 2026-09-03
test: fix incorrect traceback-collapsed assertion and clean up listener test hygiene

### LINKED ISSUE #1776: Clean up test hygiene in handlers-tab.listener.test.tsx
Quality scan of `frontend/src/components/app-detail/handlers-tab.listener.test.tsx`. All findings are test-hygiene issues local to this file; no production behavior is affected.

## 1. "Traceback is initially collapsed" comment is wrong, and collapsed state is never asserted

Lines 236-238, in `handler error banner: shows expandable traceback when available`:

```tsx
// Traceback is initially collapsed
const tracebackContent = banner.querySelector("[data-testid='traceback-content']");
expect(tracebackContent).not.toBeNull();
```

`traceback-content` is not the traceback body — it is the always-rendered wrapper `<div>` that holds the toggle button (`error-banner.tsx:35-49`). It renders whenever `traceback` is truthy, regardless of expanded state. The element that is actually conditional is the `<pre>` inside it, gated on `traceExpanded` (`error-banner.tsx:44`).

So the assertion asserts the wrapper exists — it says nothing about collapsed state, and the comment above it claims the opposite of what the line checks. The test therefore never verifies that the traceback starts hidden; only the post-click expansion (line 241) is covered.

**Fix:** assert the absence of the traceback body before the click — e.g. `expect(banner.textContent).not.toContain("Traceback (most recent call last)")` — and drop or correct the comment. The sibling test at lines 253-267 already does this correctly for the registration toggle (`expect(queryByTestId("handler-registration-source")).toBeNull()` before the click).

## 2. Comments that restate the assertion directly below them

Three inline comments narrate the line that follows without adding information the code and test name do not already carry:

- Line 122 — `// Registration is collapsed by default` above `expect(queryByTestId("handler-registration-source")).toBeNull();`
- Line 233 — `// Traceback toggle button should be present` above `expect(toggle).not.toBeNull();`
- Line 239 — `// Expand it` above `await user.click(toggle!);`

**Fix:** delete all three. The comment at line 185 (`// Dropped events are data loss: warns, with drop-rate percentage`) should stay — it explains *why* the `warn` tone is the correct expectation, which the assertion does not convey.

## 3. Raw `querySelector` + non-null assertion instead of testing-library queries

Lines 234, 237, and 240 reach into the DOM with `banner.querySelector("[data-testid='...']")` and then force the result with `toggle!`:

```tsx
const toggle = banner.querySelector("[data-testid='traceback-toggle']");
expect(toggle).not.toBeNull();
...
await user.click(toggle!);
```

Every other interaction in this file uses `getByTestId` / `queryByTestId`, which throw a descriptive error on a missing element rather than yielding `null` and requiring a `!`. The only reason `querySelector` is needed here is to scope the lookup to `banner`; testing-library's `within()` covers that case.

**Fix:** use `within(banner).getByTestId("traceback-toggle")`, which removes both the null check and the non-null assertion. Same for `traceback-content` on line 237. (Lines 186, 202, and 216 also use `querySelector`, but those select by `data-tone` / iterate a NodeList — leave them.)

## 4. `listener_id: 24` reused by two unrelated tests

Line 174 (`handler stats row: shows Backpressure Dropped with warn tone when count > 0`) and line 224 (`handler error banner: shows expandable traceback when available`) both use `listener_id: 24`. Every other test in the file has a distinct ID. Harmless today since each test renders independently, but the IDs also appear in the routing argument (`"listener/24"`) and in test-ids (`listener-detail-24`), so a duplicate makes the file harder to grep and invites confusion if the tests are ever combined into one render.

**Fix:** give the traceback test its own unused ID.

## 5. Inconsistent test-name prefixes

Most tests carry an area prefix (`handler detail:`, `handler stats row:`, `listener detail:`, `handler error banner:`), but the first three do not:

- Line 38 — `shows listener detail pane when selectedHandler='listener/5'`
- Line 43 — `uses a semantic heading for the selected handler name`
- Line 49 — `renders modifier chips for listener in detail pane`

Separately, the prefixes themselves are not applied consistently: `listener detail: shows mode chip` (line 244) and `handler detail: shows source location when available` (line 64) both describe content in the same detail pane, but use different prefixes.

**Fix:** pick one vocabulary and apply it to all tests. Given the sibling file's grouping (see #1774, finding 6), promoting these string prefixes to nested `describe` blocks would settle both files the same way.

## Acceptance Criteria

- [ ] The traceback test asserts the traceback body is absent before the click, and the incorrect "initially collapsed" comment is removed
- [ ] The three assertion-restating comments (lines 122, 233, 239) are deleted
- [ ] `traceback-toggle` / `traceback-content` lookups use `within(banner).getByTestId(...)` instead of `querySelector` + `!`
- [ ] No two tests in the file share a `listener_id`
- [ ] Test names use one consistent grouping vocabulary (prefix or `describe`) across the whole file

NodeJSmith:
## Summary

Cleans up test hygiene in `frontend/src/components/app-detail/handlers-tab.listener.test.tsx`, per the findings in #1776:

- Fixes an incorrect "traceback is initially collapsed" comment/assertion — the old check inspected the always-rendered wrapper `div` (`traceback-content`), not the actual traceback body, so collapsed state was never verified. Now asserts `banner.textContent` does not contain the traceback text before the toggle is clicked.
- Removes three comments that only restated the assertion on the next line (lines 122, 233, 239 in the original).
- Replaces `querySelector(...)` + non-null assertion (`toggle!`) with `within(banner).getByTestId(...)` for the traceback toggle/content lookups, matching the testing-library convention used elsewhere in the file.
- Gives the traceback test its own `listener_id` (26) instead of reusing `24`, which was shared with an unrelated test.
- Applies a consistent `handler detail:` prefix to the three tests that previously had no area prefix, and to the one test that used a different prefix (`listener detail:`) for the same detail pane.

No production code changed — this is a test-file-only cleanup.

## Testing

- `npx vitest run src/components/app-detail/handlers-tab.listener.test.tsx` — 18 passed
- `uv run nox -s dev` — 7206 passed, 1 skipped, 2 xfailed

Closes #1776

### DIFF
--- a/frontend/src/components/app-detail/handlers-tab.listener.test.tsx
+++ b/frontend/src/components/app-detail/handlers-tab.listener.test.tsx
@@ -1,4 +1,4 @@
-import { waitFor } from "@testing-library/react";
+import { waitFor, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
@@ -35,18 +35,18 @@ describe("HandlersTab listener detail", () => {
     vi.clearAllMocks();
   });
 
-  it("shows listener detail pane when selectedHandler='listener/5'", () => {
+  it("handler detail: shows listener detail pane when selectedHandler='listener/5'", () => {
     const { getByTestId } = renderHandlersTab([createListener({ listener_id: 5 })], [], "listener/5");
     expect(getByTestId("listener-detail-5")).toBeDefined();
   });
 
-  it("uses a semantic heading for the selected handler name", () => {
+  it("handler detail: uses a semantic heading for the selected handler name", () => {
     const listener = createListener({ listener_id: 6, handler_method: "on_garage_opened" });
     const { getByRole } = renderHandlersTab([listener], [], "listener/6");
     expect(getByRole("heading", { level: 2, name: "on_garage_opened" })).toBeDefined();
   });
 
-  it("renders modifier chips for listener in detail pane", async () => {
+  it("handler detail: renders modifier chips for listener in detail pane", async () => {
     const listener = createListener({
       listener_id: 3,
       debounce: 0.5,
@@ -119,7 +119,6 @@ describe("HandlersTab listener detail", () => {
     const toggle = getByTestId("handler-registration-toggle");
     expect(toggle.getAttribute("aria-controls")).toBe("listener-detail-12-registration-source-panel");
     expect(toggle.textContent).toContain("show call");
-    // Registration is collapsed by default
     expect(queryByTestId("handler-registration-source")).toBeNull();
     await user.click(toggle);
     const registrationSource = getByTestId("handler-registration-source");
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #1913, opened 2026-09-05
refactor: dedupe class strings and flatten branching in config-tab

### LINKED ISSUE #1634: Clean up duplicated class strings and nested render branching in config-tab.tsx
Surfaced by an automated quality scan of `frontend/src/components/app-detail/config-tab.tsx` (llm-checker, lazy-checker, nitpicker). All findings predate the scan and are independent of any in-flight work. The llm-checker and lazy-checker passes came back clean; everything below is style/hygiene.

## 1. Raw-TOML container classes are duplicated across the two render branches

`config-tab.tsx:209-210` and `config-tab.tsx:220` repeat the same block near-verbatim:

```
mt-2 overflow-x-auto whitespace-pre rounded-sm border border-dashed border-border bg-muted p-3 font-mono text-xs
```

The highlighted `<div>` branch and the plain `<pre>` fallback must stay visually identical, but nothing enforces that today — editing one and forgetting the other silently diverges the two states.

**Fix:** extract a module-scope `RAW_TOML_CLASS` constant alongside the existing `DATA_TABLE_CLASS` / `SECTION_LABEL_CLASS`, and have both branches compose from it. Note that `code-tab.tsx:170-175` carries a near-identical Shiki token-color block (`[&_.shiki_span:not(.line):not(.line-num)]` vs config-tab s `[&_.shiki_span:not(.line)]`), so a shared helper in `components/shared/` is worth considering rather than a second per-file constant.

## 2. Nested ternary chain in the config-shape branching

`config-tab.tsx:168-199` is a three-way ternary chain (`isListConfig ? ... : isConfigRecord(appConfig) ? ... : ...`) with a further ternary nested inside the first branch at `config-tab.tsx:177-186`. JSX nesting reaches six levels in the list path. Tracing which branch renders what requires following indentation carefully.

**Fix:** extract the per-instance rendering (the `.map()` body at `config-tab.tsx:172-188`) into a named `ConfigInstance` component in this file, next to the existing `AppConfigContent`. That collapses one ternary level and removes two levels of nesting.

## 3. Oversized ternary branches for the raw-config panel

`config-tab.tsx:206-225` — both branches of `tomlHtml ? ... : ...` are 10+ lines of substantial JSX. Combined with finding 1, these read better as a small named `RawConfigBlock` component (or two named render variables) than inline in a ternary.

## 4. `ConfigTab` mixes fetch orchestration with deep render branching

`config-tab.tsx:98-231` — the component is ~134 lines, past the 50-line guideline, blending the data-fetch effect with its nested best-effort highlighting try/catch, loading/error branching, and the deeply-branched shape-dependent JSX. High-level "list config or record config?" decisions sit interleaved with raw Tailwind class strings.

**Fix:** largely resolved as a byproduct of findings 1-3. If it is still unwieldy afterward, lift the fetch-and-highlight effect into a small `useAppConfig(appKey)` hook returning `{ loading, error, configData, tomlHtml }`.

## 5. `DATA_TABLE_CLASS` is one unbroken ~500-character line

`config-tab.tsx:19` combines unrelated `th` and `td` rules — borders, padding, alignment, typography casing/tracking, hover state — on a single line with no grouping. It is impossible to visually audit for ordering consistency or accidental duplication. Lines `config-tab.tsx:208-213` in the same file already demonstrate the readable alternative (a `cn()` call with one concern per line).

**Fix:** break it into grouped string segments joined by `cn()`, matching the style used at 208-213.

## 6. Table column widths do not sum to 100%

`config-tab.tsx:43` sets `w-[30%]` on the Key column and `config-tab.tsx:46` sets `w-[55%]` on Value, leaving 15% unaccounted for in a two-column table. Either the numbers are intentional and need a comment explaining the slack, or they are stale and should be `30%`/`70%`.

## 7. Small repeated class strings

Each of these appears twice with no shared constant:

- `min-w-0 px-4 pb-4` — `config-tab.tsx:167` and `config-tab.tsx:202` (the two grid columns; these genuinely must stay in sync)
- `font-mono text-sm` — `config-tab.tsx:55` and `config-tab.tsx:58`
- `text-sm text-muted-foreground` — `config-tab.tsx:151` and `config-tab.tsx:185`

**Fix:** the grid-column pair is the one worth extracting to a constant. The other two are low-value; fold them in only if touching those lines anyway.

## 8. Duplicated literals

- `"no configuration values"` — `config-tab.tsx:36` and `config-tab.tsx:198`, two `EmptyState` call sites with no shared constant, so the two empty states can drift apart in wording.
- `"toml"` — `config-tab.tsx:118` (highlighter arg) and `config-tab.tsx:122` (`lang:`). Minor, but the two must agree.

## 9. Abbreviated names inconsistent with the rest of the file

Every other identifier in the file spells concepts out (`appConfig`, `configData`, `manifestValues`, `frameworkFields`, `displayValues`), but:

- `config-tab.tsx:118` — `hl` for the highlighter
- `config-tab.tsx:172` — `instanceCfg` abbreviates "config" that is spelled out everywhere else
- `config-tab.tsx:33` vs `config-tab.tsx:69` — `SimpleConfigTable` names its prop `config` while `AppConfigContent` names the same `ConfigRecord` shape `appConfig`

**Fix:** rename to `highlighter`, `instanceConfig`, and settle on one prop name across the two sibling components.

## Acceptance Criteria

- [ ] Raw-TOML container classes live in one shared constant used by both the highlighted and fallback branches
- [ ] Per-instance config rendering is extracted into a named component; no nested ternaries remain in the JSX
- [ ] Raw-config panel branches are extracted so neither ternary branch is a 10+ line JSX literal
- [ ] `DATA_TABLE_CLASS` is broken into grouped, readable segments
- [ ] Table column widths sum to 100%, or carry a comment explaining the slack
- [ ] The two grid-column class strings share a constant
- [ ] `"no configuration values"` and `"toml"` each appear once
- [ ] `hl` and `instanceCfg` renamed; `config`/`appConfig` prop naming unified across the two local components
- [ ] Existing `config-tab.test.tsx` still passes (it asserts on the Shiki class strings at lines 195-196, so keep those class names intact through any extraction)

## Filtered as invalid

Four scanner findings were dropped with evidence:

- `grid-cols-[1.4fr_1fr]` flagged as a magic value — arbitrary grid templates are used in 11 other files (`multi-instance.tsx:107`, `stats-strip.tsx:36`, `card.tsx:45`), and CLAUDE.md s CSS section endorses arbitrary Tailwind values.
- Mixed `@/` alias and `../../` relative imports — this is the repo convention, not drift; `code-tab.tsx`, `handlers-tab.tsx`, and `registration-footer.tsx` do exactly the same (24 files use `../../api`, 12 use `@/api`).
- `loading` should be `isLoading` — 3 files use `const [loading`, zero use `isLoading`.
- `err` vs `error` naming inconsistency — `error` is already bound as state at `config-tab.tsx:100` in the same scope, so the catch binding cannot reuse the name.

NodeJSmith:
## Summary

Cleans up duplicated class strings and nested render branching in `frontend/src/components/app-detail/config-tab.tsx`, surfaced by an automated quality scan (issue #1634).

- Extract `RAW_TOML_CLASS` / `RAW_TOML_HIGHLIGHTED_CLASS` so the highlighted and plain-text raw-config panels can't silently diverge
- Extract `CONFIG_COLUMN_CLASS` for the two grid columns, which must stay visually in sync
- Extract `ConfigInstance` and `RawConfigBlock` components to remove the nested/oversized ternary branches from the JSX
- Break `DATA_TABLE_CLASS` into grouped, readable segments (borders/padding, typography, hover state)
- Fix table column widths so they sum to 100% (30%/70%, was 30%/55%)
- Deduplicate the `"no configuration values"` and `"toml"` literals into named constants
- Rename `hl` → `highlighter`, `instanceCfg` → `instanceConfig`, and unify `config`/`appConfig` prop naming across `SimpleConfigTable` and `AppConfigContent`

No behavior change — this is a pure structural/readability refactor of a single file.

## Testing

- `uv run nox -s dev` — 7678 passed, 1 skipped, 2 xfailed
- `prek -a` — all hooks passed (ruff, eslint, prettier, stylelint, tsc, house-lint, etc.)
- Existing `config-tab.test.tsx` passes unchanged, including its assertions on the Shiki class strings

Closes #1634


NodeJSmith:
@coderabbitai review

PR #1914, opened 2026-09-05
refactor: centralize duplicated handler/job execution kind union

### LINKED ISSUE #1685: Centralize the duplicated "handler" | "job" kind union used in execution-section.tsx
Automated quality scan of `frontend/src/components/app-detail/execution-section.tsx` (three parallel reviewers: llm-checker, lazy-checker, nitpicker) surfaced one finding that survived validity assessment.

### 1. `"handler" | "job"` literal union is restated inline instead of shared

`execution-section.tsx:8` declares:

```ts
kind: "handler" | "job";
```

This is not imported from anywhere — it's typed out fresh in this file's `ExecutionSectionProps`. The same literal union is independently restated (not imported from one canonical source) in several other hand-written files:

- `frontend/src/components/shared/execution-table.tsx:58`
- `frontend/src/components/app-detail/registration-footer.tsx:11`
- `frontend/src/components/app-detail/detail-header.tsx:22`
- `frontend/src/hooks/use-scoped-execution.ts:33`

(Generated files under `frontend/src/api/generated-types.ts` and `ws-types.ts` also contain this union, but those are codegen output and out of scope for a manual fix.)

Because the union is copy-pasted rather than centralized, there's nothing enforcing that all these call sites stay in sync if a third `kind` value is ever introduced — each site would need to be found and updated independently, and a missed one would only surface as a type error (or worse, a silent behavioral gap) at whichever site was missed.

**Proposed fix:** Export a named type alias (e.g. `export type ExecutionKind = "handler" | "job";`) from `execution-table.tsx` (which already owns the related `ExecutionRecord` type), then import and reuse it in `execution-section.tsx` and the other hand-written call sites listed above.

## Other findings considered and excluded

The nitpicker also flagged the `hasData` derived boolean (`execution-section.tsx:28`) as insufficiently named/commented. That finding was assessed against the code and excluded: `hasData = records !== undefined` is checked three lines later at its single use site, reads as a direct yes/no question, and the reviewer itself characterized it as "very minor" and noted it "technically passes the checklist bar" — there's no concrete drift or confusion risk to point to, so it did not survive the validity bar for filing.

NodeJSmith:
## Summary

The `"handler" | "job"` literal union was independently restated in five hand-written
frontend files instead of being imported from one canonical source, so a third `kind`
value would need to be found and updated at every call site by hand.

This exports `ExecutionKind` from `execution-table.tsx` (which already owns the related
`ExecutionRecord` type) and reuses it in:

- `execution-section.tsx`
- `registration-footer.tsx`
- `detail-header.tsx`
- `use-scoped-execution.ts`

No behavior change — this is a type-only consolidation.

## Testing

- `uv run nox -s dev` — full backend suite passes (one unrelated flake,
  `test_watcher_survives_transition_to_waiting`, failed under xdist parallelism but
  passed cleanly in isolation — a known timing race in scheduler tests, not touched by
  this change)
- `prek -a` — all hooks pass, including `tsc` and `eslint` on the frontend

Closes #1685

### DIFF
--- a/frontend/src/components/app-detail/detail-header.tsx
+++ b/frontend/src/components/app-detail/detail-header.tsx
@@ -3,6 +3,7 @@ import type { ReactNode } from "react";
 import { Badge, type BadgeVariant } from "@/components/ui/badge";
 
 import type { StatusKind } from "../../utils/status";
+import type { ExecutionKind } from "../shared/execution-table";
 import { StatusShape } from "../shared/status-shape";
 
 // Maps a StatusKind onto the flattened "kind-*" Badge variants (formerly Chip's
@@ -19,7 +20,7 @@ interface DetailHeaderProps {
   name: string;
   kindLabel: string;
   statusKind: StatusKind;
-  kind: "handler" | "job";
+  kind: ExecutionKind;
   subtitle?: string | null;
   headerActions?: ReactNode;
 }
--- a/frontend/src/components/app-detail/execution-section.tsx
+++ b/frontend/src/components/app-detail/execution-section.tsx
@@ -1,11 +1,11 @@
 import type { HandlerKind } from "../../utils/app-routes";
-import { type ExecutionRecord, ExecutionTable } from "../shared/execution-table";
+import { type ExecutionKind, type ExecutionRecord, ExecutionTable } from "../shared/execution-table";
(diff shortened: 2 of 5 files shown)


NodeJSmith:
@coderabbitai review

PR #1928, opened 2026-09-06
fix: clean up query-param parsing and error-label handling in overview-tab-helpers

### LINKED ISSUE #1871: Clean up query-param parsing and error-label handling in overview-tab-helpers
Consolidated findings from a code-quality scan (llm-checker, lazy-checker, nitpicker) of `frontend/src/components/app-detail/overview-tab-helpers.ts`. Each finding was verified against the current code before filing.

### 1. `handlerHref` bypasses `parseInstanceParam` and skips validation

`overview-tab-helpers.ts:12` re-implements "read `instance` off a query string" inline:

```ts
instanceQs ? { instance: new URLSearchParams(instanceQs).get("instance") } : undefined,
```

`utils/app-routes.ts:47` already exports `parseInstanceParam()`, purpose-built for exactly this — it trims and validates the value against a decimal regex, returning `undefined` for malformed input. The inline version passes the raw string straight through, so a malformed `?instance=abc` propagates into every generated handler link instead of being dropped. It also hardcodes the `"instance"` param name in two roles (query key and object property) with nothing tying them together.

**Fix:** `{ instance: parseInstanceParam(new URLSearchParams(instanceQs).get("instance")) }` — reuses the single source of truth for instance-param parsing and gains its validation.

### 2. No-op `kind` ternary in `handlerHref`

`overview-tab-helpers.ts:7`:

```ts
const kind: HandlerKind = item.kind === "listener" ? "listener" : "job";
```

`UnifiedItem.kind` is already typed `"listener" | "job"` (`unified-handler-row.tsx:16-25`), which is exactly `HandlerKind` (`app-routes.ts:63`). The ternary maps each value to itself.

**Fix:** pass `item.kind` directly to `handlerPath()` and delete the line.

### 3. `itemErrorType` returns a display label, not a type

`overview-tab-helpers.ts:37`:

```ts
return item.data.last_error_type ?? (item.data.timed_out > 0 ? "timed out" : null);
```

When `last_error_type` is absent, the fallback is the human-facing phrase `"timed out"` — a status description, not an error type. Both callers (`handler-health-card.tsx:62`, `error-spotlight.tsx:21`) render the value as a display label, so the behavior is fine but the name over-promises: a reader calling `itemErrorType` expecting a type/category gets a sentence fragment in the fallback case.

**Fix:** rename to `itemErrorLabel` (2 call sites) so the name matches what the function actually produces.

### 4. `"timed out"` label hardcoded in three files

The standalone status label `"timed out"` is independently hardcoded in:

- `components/app-detail/overview-tab-helpers.ts:37`
- `components/shared/execution-table.tsx:50` (`warn: "timed out"`)
- `components/app-detail/execution-detail.tsx:45`

A wording change requires hunting down all three. (The `"{n} timed out"` sentence fragments in `unified-handler-row.tsx:133` and `pages/handlers-rows.tsx:163` are grammatical contexts and can stay inline.)

**Fix:** extract a shared `TIMED_OUT_LABEL` constant (e.g. in `utils/status.ts`) and use it at the three standalone-label sites.

## Acceptance Criteria

- [ ] `handlerHref` uses `parseInstanceParam()` instead of passing the raw query value through
- [ ] The redundant `kind` ternary in `handlerHref` is removed
- [ ] `itemErrorType` is renamed to `itemErrorLabel` (or equivalent) with both call sites updated
- [ ] The standalone `"timed out"` label is a shared constant used by all three files

---

*One reviewer finding was assessed as likely invalid and excluded: centralizing the per-kind `item.kind === "listener" ? ... : ...` accessors into a field-mapping object. `UnifiedItem` is a discriminated union, so the branch on the discriminant is what TypeScript requires to narrow `item.data`; a mapping object can't accept the un-narrowed union without casts, and a generic match helper would be longer at every one-line call site than the ternary it replaces.*

NodeJSmith:
## Summary

Cleans up several code-quality issues in `overview-tab-helpers.ts` and its consumers, surfaced by a quality scan (llm-checker, lazy-checker, nitpicker).

### Changes

- **Use `parseInstanceParam()` in `handlerHref`** instead of passing raw query values to `handlerPath`. This gains the same validation/normalization that every other call site already uses, closing a consistency gap where malformed instance params would be passed through unchecked.

- **Remove redundant `kind` ternary** — `handlerHref` was mapping `item.kind` ("listener" | "job") through a ternary to produce a `HandlerKind`, but `UnifiedItem.kind` already matches the `HandlerKind` type, so the mapping was a no-op.

- **Rename `itemErrorType` → `itemErrorLabel`** — the function returns a display label string (e.g. "timed out"), not an error type category. The new name reflects its actual purpose. Updated all three call sites.

- **Extract `TIMED_OUT_LABEL` constant** to `utils/status.ts` — the string `"timed out"` was hardcoded in three separate files (`overview-tab-helpers.ts`, `execution-detail.tsx`, `execution-table.tsx`). Replaced all occurrences with the shared constant.

## Testing

- All 1601 frontend tests pass with full coverage thresholds met
- All 29 prek hooks pass (including eslint, prettier, tsc, stylelint)
- Pyright clean
- 7688/7689 backend tests pass (1 pre-existing failure in `test_file_watcher` unrelated to this change — verified against main)

Closes #1871

### DIFF
--- a/frontend/src/components/app-detail/error-spotlight.tsx
+++ b/frontend/src/components/app-detail/error-spotlight.tsx
@@ -6,7 +6,7 @@ import { cn } from "@/lib/utils";
 import { STATUS_DOT_SIZE } from "../../utils/constants";
 import { StatusShape } from "../shared/status-shape";
 import { OVERVIEW_SECTION_CLASS } from "./overview-section";
-import { handlerHref, itemErrorMessage, itemErrorType } from "./overview-tab-helpers";
+import { handlerHref, itemErrorLabel, itemErrorMessage } from "./overview-tab-helpers";
 import type { UnifiedItem } from "./unified-handler-row";
 
 const SPOTLIGHT_LIMIT = 3;
@@ -18,7 +18,7 @@ interface SpotlightEntryProps {
 }
 
 function SpotlightEntry({ item, appKey, instanceQs }: SpotlightEntryProps) {
-  const errorType = itemErrorType(item);
+  const errorLabel = itemErrorLabel(item);
   const errorMessage = itemErrorMessage(item);
   const href = handlerHref(appKey, item, instanceQs);
 
@@ -33,7 +33,7 @@ function SpotlightEntry({ item, appKey, instanceQs }: SpotlightEntryProps) {
       <span className="shrink-0 whitespace-nowrap font-mono text-[length:var(--text-mono-sm)] font-medium text-foreground">
         {item.name}
       </span>
-      {errorType && <span className="shrink-0 whitespace-nowrap text-sm text-destructive">{errorType}</span>}
+      {errorLabel && <span className="shrink-0 whitespace-nowrap text-sm text-destructive">{errorLabel}</span>}
       {errorMessage && (
         <span className="min-w-0 flex-1 truncate text-sm text-foreground-secondary" title={errorMessage}>
           {errorMessage}
--- a/frontend/src/components/app-detail/execution-detail.tsx
+++ b/frontend/src/components/app-detail/execution-detail.tsx
@@ -10,7 +10,7 @@ import type { components } from "../../api/generated-types";
 import { useDocumentTitle } from "../../hooks/use-document-title";
 import { STATUS_DOT_SIZE } from "../../utils/constants";
 import { formatDuration, formatTimestamp, truncateId } from "../../utils/format";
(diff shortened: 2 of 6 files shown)


NodeJSmith:
@coderabbitai review

PR #1929, opened 2026-09-06
test: clean up duplication and comment bloat in overview-tab.test.tsx

### LINKED ISSUE #1875: Clean up duplication and comment bloat in overview-tab.test.tsx
Automated quality scan (llm-checker + lazy-checker + nitpicker) surfaced the following in `overview-tab.test.tsx`. All findings below passed a validity check against the actual file content.

## Findings

**1. Duplicate render helper**
`renderOverview()` (line 594-599) is a near-verbatim copy of `renderOverviewTab()` (line 31-48) — same component, same props shape, same `storeOverrides: { uptimeSeconds: 120 }` — but hardcodes empty `listeners`/`jobs` arrays and doesn't return the render result. `renderOverviewTab()` already accepts `listeners`/`jobs` as parameters, so the "Real-time refetch" tests can call it directly (`renderOverviewTab({ listeners: [], jobs: [] })`) and `renderOverview()` can be deleted. This also removes the duplicated `uptimeSeconds: 120` literal and the confusing near-identical naming between the two helpers.

**2. Missing `ActivityFeedEntry` factory**
Six tests (around lines 347-359, 375-387, 409-421, 433-489, 503-526, 534-544) hand-construct full `ActivityFeedEntry` object literals (7-9 fields each) inline, repeating boilerplate fields (`app_key`, `error_type: null`, etc.) unrelated to what each test actually checks. The file already uses shared factories for other fixtures (`createListener`, `createJob`, `createLogEntry` from `../../test/factories`). Add a `createActivityFeedEntry()` factory with sensible defaults alongside those, and use it at each call site.

**3. Repeated refetch-test scaffold**
The three "Real-time refetch" tests (line 601-678) repeat an identical five-line scaffold — `setupActivityCounter()`, render, wait for initial fetch, capture `countAfterMount` — before diverging only on the `executionCompleted` payload and final assertion. Extract a small helper that takes the payload and expected outcome (refetch vs. no-refetch).

**4. Magic number and non-descriptive param name**
Line 650 uses a bare `setTimeout(r, 700)` while the two adjacent tests (lines 624, 677) correctly reference the imported `WS_DEBOUNCE_MAX_WAIT_MS` constant for the same debounce-window wait. Use the same constant here so the wait can't silently drift out of sync with the debounce window it's testing. The single-letter `r` resolver parameter should also be spelled out (`resolve`) per the codebase's naming conventions.

**5. Comment bloat**
- Line 154: `// Initially only 3 visible` sits directly above the assertion checking exactly that — pure restatement, delete.
- Line 276: `// The failing listener (id=2) should appear first` — same pattern, restates the assertion immediately below it, delete.
- Lines 373-374: a 2-line comment narrates a past incident (`Regression: the nowrap duration/time columns overflowed the mobile viewport, making the entire main column horizontally scrollable.`) instead of stating the durable invariant the test enforces. Replace with a one-line comment stating the invariant (e.g. "activity table scrolls horizontally instead of the whole column") or drop it — the assertions already encode the behavior.

**6. Repeated `"test_app"` string literal**
The string literal `"test_app"` (and secondary variant `"my_app"`) is repeated inline roughly 20 times across the file with no shared constant. Extract `const TEST_APP_KEY = "test_app"` (and `TEST_APP_KEY_ALT = "my_app"` if the second identity is intentionally distinct) so a future rename of the fixture app is a one-line edit.

## Proposed fix

Mechanical cleanup, no behavior change: delete the duplicate render helper in favor of parameterizing the existing one, add one new factory function, extract one small test-scaffold helper, swap one magic number for the existing constant, trim three comments, and add two string constants.

NodeJSmith:
## Summary

Reduce duplication and clean up comment bloat in `overview-tab.test.tsx`.

### Changes

- **Add `createActivityFeedEntry` factory** to `frontend/src/test/factories.ts` — six tests were hand-constructing 7-9 field `ActivityFeedEntry` objects inline, repeating boilerplate (`app_key`, `error_type: null`, etc.) unrelated to what each test checks. Now each test passes only the fields it cares about.

- **Extract `setupRefetchTest` helper** — the three real-time refetch tests repeated an identical scaffold (setup counter, render, wait for initial fetch, capture count). Extracted into one helper returning `{ getFetchCount, countAfterMount }`.

- **Delete duplicate `renderOverview` helper** — was a near-verbatim copy of `renderOverviewTab` with hardcoded empty arrays. The existing `renderOverviewTab({ listeners: [], jobs: [] })` does the same thing.

- **Extract `TEST_APP_KEY` / `OTHER_APP_KEY` constants** — the string `"test_app"` appeared ~20 times and `"my_app"` appeared ~5 times inline. Now defined once at the top. URL assertions use template literals referencing the constant.

- **Fix magic number** — one test used bare `setTimeout(r, 700)` while adjacent tests used the imported `WS_DEBOUNCE_MAX_WAIT_MS` constant for the same debounce-window wait. Aligned to the constant and renamed the single-letter `r` resolver to `resolve`.

- **Trim comments** — removed three comments that restated the assertion immediately below them, and shortened one regression-narrating comment to a durable invariant statement.

- **Use `createExecutionCompletedPayload` factory** — the refetch tests were hand-constructing `executionCompleted` payloads inline; replaced with the existing factory from `factories.ts`.

### Testing

All 7689 tests pass (`uv run nox -s dev`). All lint and type checks pass (`prek -a`, `prek pyright -a --stage pre-push`). No behavior changes — purely mechanical cleanup.

Closes #1875

### DIFF
--- a/frontend/src/components/app-detail/overview-tab.test.tsx
+++ b/frontend/src/components/app-detail/overview-tab.test.tsx
@@ -3,16 +3,19 @@ import userEvent from "@testing-library/user-event";
 import { http, HttpResponse } from "msw";
 import { describe, expect, it, vi } from "vitest";
 
-import type { components } from "../../api/generated-types";
 import { WS_DEBOUNCE_MAX_WAIT_MS } from "../../hooks/use-query-invalidator";
 import { useAppStore } from "../../state/store";
-import { createJob, createListener, createLogEntry } from "../../test/factories";
+import {
+  createActivityFeedEntry,
+  createExecutionCompletedPayload,
+  createJob,
+  createListener,
+  createLogEntry,
+} from "../../test/factories";
 import { createWouterMock } from "../../test/mock-wouter";
 import { renderWithAppState } from "../../test/render-helpers";
 import { server } from "../../test/server";
 import { OverviewTab } from "./overview-tab";
-
-type ActivityFeedEntry = components["schemas"]["ActivityFeedEntry"];
 
 // Overview tab tests are split into two groups:
 //  1. Props-only tests (error spotlight, health grid) — no context needed
@@ -28,10 +31,13 @@ vi.mock("wouter", () =>
   }),
 );
 
+const TEST_APP_KEY = "test_app";
+const OTHER_APP_KEY = "my_app";
+
 function renderOverviewTab({
   listeners = [createListener()],
   jobs = [createJob()],
-  appKey = "test_app",
+  appKey = TEST_APP_KEY,
   instanceQs = "",
   resolvedInstanceIndex = 0,
 } = {}) {
@@ -151,7 +157,6 @@ describe("OverviewTab — Error Spotlight", () => {
       createListener({ listener_id: 5, failed: 1 }),
     ];
     const { getAllByTestId, getByTestId } = renderOverviewTab({ listeners, jobs: [] });
-    // Initially only 3 visible
     expect(getAllByTestId(/^overview-error-spotlight-entry-/).length).toBe(3);
     const btn = getByTestId("overview-error-spotlight-show-more");
     expect(btn.textContent).toContain("2");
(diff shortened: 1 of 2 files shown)


NodeJSmith:
@coderabbitai review

PR #1939, opened 2026-09-07
test: state durable invariants in lifecycle regression docstrings

### LINKED ISSUE #1880: Clean up narrated-incident docstrings in lifecycle regression tests
## Description

Surfaced by `/mine-clean-code` (llm-checker) while reviewing branch `1333` (108-testing-infra-split). All findings are pre-existing — outside the hunks that branch actually changed (confirmed via `git diff origin/1860...HEAD -U0`).

A handful of regression-test docstrings narrate the incident/discovery story (specific bug numbers, "poisoning every later test," a named reviewer's catch) rather than stating the durable invariant in one sentence with a ticket pointer, per `coding-style.md`'s Self-Contained Comments rule. The reviewing agent flagged these as low-confidence — the narrative sits alongside genuinely load-bearing technical context in each case — so use judgment on how much to strip versus keep.

## Key Items

- `tests/unit/resources/test_lifecycle_transitions.py:317-330` (`test_handle_failed_on_terminal_resource_is_noop`): docstring narrates "On Python 3.11 that error escaped harness teardown before the global singleton was reset, poisoning every later test on the xdist worker" instead of stating the invariant ("handle_failed() on a terminal resource must be a no-op, even under strict_lifecycle") with a ticket pointer.
- `tests/unit/test_framework_injection_points.py:134-141` (`TestHarnessTeardownClearsSingletonOnError`): repeats the same "#1059 poisoning every later test on the xdist worker" narrative in a second file.
- `tests/unit/resources/test_lifecycle_transitions.py:446-457` (`test_cancel_from_worker_thread_redispatches_onto_loop_thread`): attributes the regression to "the Codex review finding on lifecycle.py's cancel()" — a named reviewer catch rather than a ticket link.

## Acceptance Criteria

- [ ] Each docstring above states the durable invariant in one sentence, keeps a ticket/issue link if one exists, and drops the incident play-by-play (specific Python version, "poisoning," named reviewer) per `coding-style.md`'s Self-Contained Comments guidance — unless closer reading shows the narrative is genuinely necessary to explain why the test can't be simplified, in which case leave it and close as won't-fix with rationale.

NodeJSmith:
## Summary

Rewrites three regression-test docstrings that narrated how a bug was found instead of stating what the test pins. Per `coding-style.md`'s Self-Contained Comments rule, a comment has to read clearly to someone with no memory of the incident that produced it.

## What changed

`tests/unit/resources/test_lifecycle_transitions.py`

- `test_handle_failed_on_terminal_resource_is_noop` — leads with the invariant (terminal states have no valid transition to `FAILED`, so a late failure after teardown is benign and `handle_failed()` must leave the status unchanged). The `RuntimeError("cannot schedule new futures after shutdown")` example and the `strict_lifecycle`/`VALID_TRANSITIONS` mechanics are load-bearing and stay; the Python 3.11 version specifics and the xdist-worker-poisoning play-by-play are gone. The `#1059` pointer is kept.
- `test_cancel_from_worker_thread_redispatches_onto_loop_thread` — drops the "Regression test for the Codex review finding" attribution. The full explanation of the race (cancel() running synchronously on the worker thread, seeing `_pending_start_task` still `None`) was already in the docstring and is untouched.

`tests/unit/test_framework_injection_points.py`

- `TestHarnessTeardownClearsSingletonOnError` — same treatment; states that an escape before the contextvar reset leaks `HASSETTE_INSTANCE` and that the reset must live in a `finally`, without re-telling the 3.11 xdist story that the other file also told. Keeps the `#1059` pointer.
- The nested `_TeardownBoom` docstring now says the test pins the invariant for any escape rather than "just the specific 3.11 exception," and the inline comment states what the `finally` must do rather than what happened before the fix.

Docstrings and comments only — no test logic, assertions, or source changes.

## Testing

- `uv run nox -s dev` — 7689 passed, 1 skipped, 2 xfailed
- `prek -a` — all hooks pass

Closes #1880

### DIFF
--- a/tests/unit/resources/test_lifecycle_transitions.py
+++ b/tests/unit/resources/test_lifecycle_transitions.py
@@ -318,15 +318,12 @@ async def test_running_to_stopped_direct_is_valid():
 async def test_handle_failed_on_terminal_resource_is_noop(terminal: ResourceStatus):
     """handle_failed() on an already-terminal resource is a no-op, even in strict mode.
 
-    Regression for #1059: during teardown a late submit-after-shutdown error
-    (``RuntimeError("cannot schedule new futures after shutdown")``) could surface on a
-    resource that had already reached STOPPED. handle_failed() then drove a STOPPED → FAILED
-    transition, which VALID_TRANSITIONS forbids — raising InvalidLifecycleTransitionError under
-    strict_lifecycle. On Python 3.11 that error escaped harness teardown before the global
-    singleton was reset, poisoning every later test on the xdist worker.
-
-    Terminal states (STOPPED, EXHAUSTED_DEAD) cannot transition to FAILED, so failing one is
-    benign — handle_failed() must leave the status unchanged instead of raising.
+    Regression for #1059. Terminal states (STOPPED, EXHAUSTED_DEAD) have no valid transition to
+    FAILED, so a late failure arriving after teardown completed — such as a submit-after-shutdown
+    ``RuntimeError("cannot schedule new futures after shutdown")`` — is benign. handle_failed()
+    must leave the status unchanged rather than drive a STOPPED → FAILED transition that
+    VALID_TRANSITIONS forbids, which would raise InvalidLifecycleTransitionError under
+    strict_lifecycle.
     """
     hassette = make_mock_hassette(strict_lifecycle=True, sealed=False)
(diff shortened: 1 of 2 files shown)


NodeJSmith:
@coderabbitai review

PR #1942, opened 2026-09-07
test: name repeated fixture literals in registration-footer tests

### LINKED ISSUE #1892: Clean up repeated fixture literals in registration-footer.test.tsx
## Description

Quality-scanner pass over `frontend/src/components/app-detail/registration-footer.test.tsx`. Two of three checkers (llm-checker, lazy-checker) returned CLEAN — the file has no comment bloat, no deferred debt, and no duplicated shared factories. The nitpicker surfaced repeated inline fixture literals and one naming inconsistency. All findings predate the current work and are cosmetic; the tests themselves cover the component's render branches correctly.

### 1. `"apps/foo.py:12"` repeated 3x with an implicit coupling to the expected line number

`frontend/src/components/app-detail/registration-footer.test.tsx:15,27,38` all pass `sourceLocation="apps/foo.py:12"`, and line 33 separately asserts `expect(onViewCode).toHaveBeenCalledWith(12)`. The `12` in the assertion is derived from the `:12` suffix in the fixture string (the component parses it via `parseSourceLocation` — see `registration-footer.tsx:31,57`), but nothing in the test makes that relationship visible. Editing the fixture string without editing the assertion silently desyncs them.

**Fix:** hoist a named fixture pair above `describe`, e.g. `const SOURCE_LINE = 12;` and `` const SOURCE_LOCATION = `apps/foo.py:${SOURCE_LINE}`; ``, then use both in the tests so the assertion reads `toHaveBeenCalledWith(SOURCE_LINE)`.

### 2. `"handler-detail-1"` repeated 4x

`registration-footer.test.tsx:9,15,26,38` each re-declare the same `testId` prop value inline. Note this value is opaque to the assertions — the component only uses `testId` to build `aria-controls`/`aria-labelledby` ids (`registration-footer.tsx:32-33`), and every `getByTestId` in the test queries a `kind`-derived id instead. So there is no correctness risk here, just four copies of the same arbitrary string.

**Fix:** hoist to a single `const HANDLER_TEST_ID = "handler-detail-1";` at the top of the file. Low priority — bundle with item 1 rather than as its own change.

### 3. Abbreviated variable name inconsistent with the rest of the file

`registration-footer.test.tsx:31` uses `const btn = getByTestId("view-in-code-btn")`, while comparable locals elsewhere in the same file are spelled out (`toggle` at line 48, `user` at lines 21/44).

**Fix:** rename `btn` to `viewCodeButton` (or `button`) for consistency.

## Acceptance Criteria

- [ ] `apps/foo.py:12` and the expected line number `12` are defined once as named constants, with the assertion referencing the named line number rather than a bare literal
- [ ] `"handler-detail-1"` is declared once and reused across the four handler tests
- [ ] `btn` is renamed to match the file's spelled-out local naming convention
- [ ] `npx vitest run src/components/app-detail/registration-footer.test.tsx` passes

## Excluded from this issue

The nitpicker also flagged `"scheduler.run_in(foo, 5)"` and `"job-detail-42"` (line 46) as magic strings. Assessed as not actionable: each appears exactly once, neither is asserted on elsewhere, and inline single-use fixture data in the test that consumes it is the intended shape — there is no duplication to consolidate and no drift risk to prevent.

NodeJSmith:
## Summary

Cleans up repeated inline fixture literals in `frontend/src/components/app-detail/registration-footer.test.tsx`, surfaced by the quality scanner in #1892. Test-only change — no production code or behavior is touched.

## What changed

- **Named the source-location fixture pair.** `"apps/foo.py:12"` appeared in three tests, and a fourth assertion separately expected `toHaveBeenCalledWith(12)`. That `12` is derived from the `:12` suffix (the component parses it via `parseSourceLocation`), but nothing in the test made the coupling visible — editing the fixture string would have silently desynced the assertion. Both are now defined once as `SOURCE_LINE` and `` SOURCE_LOCATION = `apps/foo.py:${SOURCE_LINE}` ``, and the assertion reads `toHaveBeenCalledWith(SOURCE_LINE)`.
- **Hoisted the repeated handler testId.** `"handler-detail-1"` was re-declared inline in four tests; it is now a single `HANDLER_TEST_ID` constant. No correctness risk either way — the component only uses `testId` to build `aria-controls`/`aria-labelledby` ids — this is purely deduplication.
- **Spelled out an abbreviated local.** `btn` → `viewCodeButton`, matching the file's other locals (`toggle`, `user`).

Deliberately left alone: the single-use `"scheduler.run_in(foo, 5)"` and `"job-detail-42"` literals. Each appears exactly once and is not asserted on elsewhere, so inline fixture data in the test that consumes it is the right shape — there is no duplication to consolidate and no drift to prevent.

## Testing

- `npx vitest run src/components/app-detail/registration-footer.test.tsx` — 5 passed
- `uv run nox -s dev` — 7690 passed, 1 failed. The single failure is
  `tests/integration/test_file_watcher.py::test_event_emitted_on_file_change`, a real-clock
  2s deadline that flakes under 4 parallel xdist workers. It is unrelated to this branch:
  the branch's only changed file is the frontend test above, and `test_file_watcher.py` and
  all of `src/hassette/` are byte-identical to `origin/main`. It passes on isolated re-run
  (`uv run pytest tests/integration/test_file_watcher.py` — 1 passed).
- `prek -a` — all 29 hooks pass (ruff, eslint, stylelint, tsc, prettier, house-lint)

Closes #1892

### DIFF
--- a/frontend/src/components/app-detail/registration-footer.test.tsx
+++ b/frontend/src/components/app-detail/registration-footer.test.tsx
@@ -4,15 +4,19 @@ import { describe, expect, it, vi } from "vitest";
 
 import { RegistrationFooter } from "./registration-footer";
 
+const HANDLER_TEST_ID = "handler-detail-1";
+const SOURCE_LINE = 12;
+const SOURCE_LOCATION = `apps/foo.py:${SOURCE_LINE}`;
+
 describe("RegistrationFooter", () => {
   it("renders nothing when no sourceLocation and no registrationSource", () => {
-    const { container } = render(<RegistrationFooter kind="handler" testId="handler-detail-1" />);
+    const { container } = render(<RegistrationFooter kind="handler" testId={HANDLER_TEST_ID} />);
     expect(container.innerHTML).toBe("");
   });
 
   it("shows source location when provided", () => {
     const { getByTestId } = render(
-      <RegistrationFooter kind="handler" testId="handler-detail-1" sourceLocation="apps/foo.py:12" />,
+      <RegistrationFooter kind="handler" testId={HANDLER_TEST_ID} sourceLocation={SOURCE_LOCATION} />,
     );
     expect(getByTestId("handler-source-location")).not.toBeNull();
   });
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #1791, opened 2026-09-02
chore: retype role parameters to ResourceRole in service_watcher

NodeJSmith:
## Summary

`role` is always a real `ResourceRole` enum value at runtime — `Service.role` is declared
`ClassVar[ResourceRole]`, and both `ServiceStatusPayload.role` and
`HassetteServiceEvent.from_service_status()`'s `role` parameter are already correctly typed.
But every `role`-accepting method on `ServiceWatcher` declared `role: object`, so passing one
of those values into a correctly-typed function required a local
`# pyright: ignore[reportArgumentType]` suppression at the call site — suppressing a type
error that only existed because the parameter was under-typed in the first place.

This retypes all 10 `role` parameters in `src/hassette/core/service_watcher.py` to
`ResourceRole` (`ResourceRole | None` for `log_if_admission_blocked`, which accepts `None`)
and drops the two suppressions that loose typing was producing.

## Changes

- `service_key`, `get_service`, `set_service_status`, `emit_service_status_event`,
  `handle_restart_refused`, `handle_exhaustion`, `cooldown_and_retry`, `execute_restart`,
  `await_service_readiness` — `role: object` → `role: ResourceRole`
- `log_if_admission_blocked` — `role: object | None` → `role: ResourceRole | None`
- Removed both now-unnecessary `# pyright: ignore[reportArgumentType]` suppressions at the
  `ServiceStatusPayload` and `from_service_status()` call sites
- Added `ResourceRole` to the existing `hassette.types` import

Annotations only — no runtime behavior change.

## Testing

- `uv run nox -s dev` — 7188 passed, 1 skipped, 2 xfailed
- `prek -a` — all 28 hooks pass
- `prek pyright -a --stage pre-push` — passes with no new errors

Closes #1768

PR #1954, opened 2026-09-08
test: widen shutdown-hook race test timeout past CI scheduling jitter

### LINKED ISSUE #1794: Fix flaky shutdown-hook race test whose real-clock margin is only 85ms
## Description

`tests/unit/resources/lifecycle/test_shutdown.py::test_resource_own_hanging_hook_wins_race_against_coordinator_timeout`
failed on CI for PR #1791 (an annotations-only diff that touches nothing in the lifecycle path):

```
AssertionError: assert <TeardownCause.SHUTDOWN_HOOK_FAILED: 'shutdown_hook_failed'> in
  (<TeardownCause.FORCED_TERMINAL: 'forced_terminal'>,
   <TeardownCause.SHUTDOWN_BODY_TIMED_OUT: 'shutdown_body_timed_out'>)
```

Confirmed a flake, not a regression: the test file is byte-identical to `origin/main`, the test
passed 15/15 locally in a loop, and a plain CI rerun of the same commit went green.

This is a variant of the pattern tracked by #1503, but it does not match that issue's stated
(a)+(b) shape — there is no separate deliberate real-time hold competing with the override.
Instead the override is small enough that the lifecycle budget's *internal* split leaves a
sub-100ms real-clock margin.

## Root cause

The test sets `resource_shutdown_timeout_seconds = SHORT_SHUTDOWN_TIMEOUT_SECONDS`
(`src/hassette/test_utils/helpers.py:54` — `0.1`). Running that through
`compute_shutdown_budget(0.1, 0.0)`:

| Stage | Deadline |
|---|---|
| hooks pool (inner bound the test expects to win) | **15 ms** |
| body deadline | 90 ms |
| total deadline (coordinator's outer `asyncio.wait` bound) | 100 ms |

The test's own docstring asserts the hook's inner timeout "always fires first by construction."
The *ordering* is indeed correct by construction (15 < 90 < 100), but the absolute margin between
the hook's bound and the coordinator's is only **85 ms of real wall-clock time**. Under CI's
`-n 4` xdist workers on a loaded runner, 85 ms of asyncio callback-scheduling delay is easily
reachable, and the coordinator's outer bound then fires first — producing exactly the observed
`FORCED_TERMINAL` + `SHUTDOWN_BODY_TIMED_OUT` pair instead of `SHUTDOWN_HOOK_FAILED`.

Note that `0.1` is far below the ~3.33s threshold documented at
`src/hassette/resources/lifecycle.py:55-62`, so every stage is in the proportionally-scaled
branch of `compute_shutdown_budget()` rather than getting its named floor.

## Suggested fix

Per CLAUDE.md's "Config-driven real-clock timeouts" guidance, widen the override rather than
shrinking anything else — nothing in this test depends on the shutdown budget being small, only
on the hook hanging past its pool. Margins at candidate values:

| `resource_shutdown_timeout_seconds` | hooks bound | total bound | margin |
|---|---|---|---|
| 0.1 (current) | 15 ms | 100 ms | **85 ms** |
| 1.0 | 150 ms | 1000 ms | 850 ms |
| 4.0 | 1100 ms | 4000 ms | 2900 ms |

`4.0` is the first value at or above the ~3.33s threshold, so the budget uses its real named
floors instead of the scaled branch — which also makes the test exercise the production code path
rather than the degenerate one.

Reproduce the failure deterministically before fixing by driving the value further down
(e.g. `0.01`) and confirming the failure signature matches the one above.

## Scope check

`SHORT_SHUTDOWN_TIMEOUT_SECONDS` is shared — it is also used by
`tests/unit/resources/lifecycle/test_total_timeout.py` and
`tests/unit/resources/lifecycle/test_force_terminal.py`. Some of those tests may genuinely depend
on the scaled-branch behaviour, so prefer a local override in the affected test over changing the
shared constant, unless an audit shows every consumer wants the wider value.

## Acceptance Criteria

- [ ] The failure is reproduced deterministically by lowering the timeout, with a signature
      matching the CI output above
- [ ] `test_resource_own_hanging_hook_wins_race_against_coordinator_timeout` uses a timeout whose
      hooks-to-total margin is comfortably above CI scheduling jitter
- [ ] Other `SHORT_SHUTDOWN_TIMEOUT_SECONDS` consumers are checked, and either left alone with a
      note or migrated deliberately
- [ ] The test's docstring is updated so its "by construction" claim states the real invariant
      (ordering is guaranteed; the absolute margin is what must be kept generous)

NodeJSmith:
## Summary

`test_resource_own_hanging_hook_wins_race_against_coordinator_timeout` was flaky on CI — it failed on PR #1791, an annotations-only diff that touches nothing in the lifecycle path, and a plain rerun of the same commit went green.

The test asserts that a resource's *own* hanging shutdown hook hits its inner `asyncio.timeout` (producing `SHUTDOWN_HOOK_FAILED`) before the coordinator's outer `asyncio.wait([body_task], timeout=...)` bound fires. It set `resource_shutdown_timeout_seconds = SHORT_SHUTDOWN_TIMEOUT_SECONDS` (`0.1`), which `compute_shutdown_budget()` splits into a 15 ms hooks-pool bound and a 100 ms total deadline.

`COORDINATOR_MARGIN_FRACTION` guarantees the *ordering* of those bounds (15 < 90 < 100), and the docstring claimed that made the hook "always fire first by construction." But ordering isn't the whole invariant: the body task still has to be scheduled and reach `run_hooks()` inside that window, and the absolute real-clock margin was only 85 ms. On a loaded `-n 4` xdist runner, 85 ms of asyncio callback-scheduling delay is easily reachable — the coordinator's outer bound wins instead, and the report comes back `FORCED_TERMINAL` + `SHUTDOWN_BODY_TIMED_OUT` with no hook evidence at all. That is exactly the observed CI failure.

## What changed

- Switched the test to the existing `GENEROUS_SHUTDOWN_TIMEOUT_SECONDS` (`5.0`). Nothing in this test depends on the shutdown budget being small — only on the hook hanging past its pool — so per CLAUDE.md's "Config-driven real-clock timeouts" guidance the fix widens the override rather than shrinking the deliberate delay under test. At `5.0` the budget is also above the ~3.33s threshold in `lifecycle.py`, so it uses its real named floors instead of the degenerate proportionally-scaled branch, meaning the test now exercises the production code path.
- Rewrote the test docstring to state the real invariant (the absolute wall-clock gap between the hooks pool and the total deadline) instead of the "by construction" ordering claim that misled the original value choice.
- Documented on `SHORT_SHUTDOWN_TIMEOUT_SECONDS` that it is only for tests that *want* the coordinator's outer bound to fire, and to reach for `GENEROUS_SHUTDOWN_TIMEOUT_SECONDS` otherwise. Every other consumer of the short value asserts the timeout/force-terminal branch does fire, so they correctly keep it.

No production code is touched — this is test infrastructure only.

## Testing

- `uv run nox -s dev` — 7716 passed, 1 skipped, 2 xfailed.
- `prek -a` — all hooks pass. `prek pyright -a --stage pre-push` — passes.
- The fixed test run 20x in a loop: 20/20 pass, ~0.38s each. Despite the 5.0s budget the test still resolves in well under half a second, because the hook's inner bound is what actually ends it.

Closes #1794

### DIFF
--- a/tests/support/helpers.py
+++ b/tests/support/helpers.py
@@ -47,7 +47,13 @@ SETTLE_SECONDS = 0.05
 """Default settle window: seconds to let a stray extra handler call land before a negative assertion."""
 
 SHORT_SHUTDOWN_TIMEOUT_SECONDS = 0.1
-"""Short ``resource_shutdown_timeout_seconds`` for tests that force a timeout/force-terminal branch."""
+"""Short ``resource_shutdown_timeout_seconds`` for tests that force a timeout/force-terminal branch.
+
+Only for tests that *want* the coordinator's outer bound to fire. It leaves under 100ms of real
+wall-clock margin between the hooks pool and the total deadline, so a test asserting that an
+inner bound (a hook's own ``asyncio.timeout``) wins that race will flake on a loaded CI runner.
+Use ``GENEROUS_SHUTDOWN_TIMEOUT_SECONDS`` for those.
+"""
 
 DB_HASSETTE_TELEMETRY_WRITE_QUEUE_MAX = 500
 """``make_mock_hassette`` database queue-max override — smaller than the production default (1000) for test speed."""
--- a/tests/unit/resources/lifecycle/test_shutdown.py
+++ b/tests/unit/resources/lifecycle/test_shutdown.py
@@ -312,12 +312,19 @@ async def test_resource_own_hanging_hook_wins_race_against_coordinator_timeout()
     and letting ``_shutdown_body()`` proceed through its later stages — rather than losing the
     race to the coordinator's own outer ``asyncio.wait([body_task], timeout=timeout)`` bound.
 
-    The up-front budget's ``COORDINATOR_MARGIN_FRACTION`` guarantees that the body's stages
-    (hooks pool + mandatory tail) finish before the coordinator's outer wait, so the hook's
-    inner timeout always fires first by construction.
+    The up-front budget's ``COORDINATOR_MARGIN_FRACTION`` fixes the *ordering* by construction
+    (hooks pool < body deadline < total deadline), but ordering alone is not what makes this
+    test pass: the body task still has to be scheduled and reach ``run_hooks()`` before the
+    coordinator's outer bound expires. The real invariant is therefore the absolute wall-clock
+    gap between the hooks pool and the total deadline, which must stay generous enough to
+    absorb CI scheduling jitter — hence ``GENEROUS_SHUTDOWN_TIMEOUT_SECONDS`` rather than the
+    short timeout the force-terminal tests use. That short value leaves under 100ms of margin,
+    which a loaded ``-n 4`` runner reaches; the coordinator's outer wait then fires first and
+    the report carries ``FORCED_TERMINAL`` + ``SHUTDOWN_BODY_TIMED_OUT`` with no hook evidence
+    at all.
     """
     hassette = make_mock_hassette(sealed=False)
-    hassette.config.lifecycle.resource_shutdown_timeout_seconds = SHORT_SHUTDOWN_TIMEOUT_SECONDS
+    hassette.config.lifecycle.resource_shutdown_timeout_seconds = GENEROUS_SHUTDOWN_TIMEOUT_SECONDS
 
     resource = HangingChild(hassette)
     await resource.initialize()


NodeJSmith:
@coderabbitai review

PR #1956, opened 2026-09-08
fix: remove loop-blocking not-ready retry from StateProxy read path

### LINKED ISSUE #1803: Remove the loop-blocking not-ready retry from the StateProxy read path
## Description

`StateProxy`'s read path uses `@_retry_on_not_ready` (`state_proxy.py:36-43`) on `get_state` (`:317`) and `yield_domain_states` (`:373`) — a **sync** tenacity retry (5 attempts, `wait_exponential_jitter(0.01..0.1)`). Sync tenacity waits are `time.sleep()`, and these methods run on the event loop thread (they're the sync reads behind `StateManager`). Three problems:

1. **The retry provably cannot succeed.** Cache freshness is loop-confined state — it only transitions when async code (`_run_synchronization`, `on_disconnect`) runs on the event loop. The retry's `time.sleep` blocks that same thread between attempts, so the condition it is polling is frozen for the duration. Every retry attempt re-reads state that cannot have changed; the decorator adds ~0.3-0.4s of loop blockage and then raises the same `ResourceNotReadyError` the first attempt would have.
2. **It taxes the shutdown budget.** The one place app code can actually trigger it is an `on_shutdown` hook reading state after `StateProxy.on_shutdown` reset freshness to UNAVAILABLE — burning ~0.3-0.4s of the shutdown hooks pool per read, on the loop thread, during the phase where budget math matters most. Normal runtime never sees UNAVAILABLE (post-bootstrap disconnects degrade to STALE, which reads fine), and pre-bootstrap windows are closed structurally by `AppBootstrapCoordinator` gating.
3. **It trips the framework's own instrumentation.** The Tier 2 block-IO guard patches `time.sleep` (`block_io_guard.py:59`), so the framework's own read path would be flagged as a blocking call.

App-author DX is unaffected by removal: readiness guarantees come from bootstrap gating (apps don't start until initial state capability commits), not from this retry. See the companion enhancement for an async freshness signal (`wait_fresh()`) — the genuinely useful DX feature in this area.

## Steps to Reproduce

1. Call `state_proxy.get_state("light.kitchen")` while `cache_freshness == UNAVAILABLE` (e.g. during `on_shutdown`, after `StateProxy.on_shutdown` ran).
2. Observe ~0.3-0.4s of event-loop blockage (Tier 2 guard flags `time.sleep` if enabled), followed by `ResourceNotReadyError` — identical outcome to attempt 1, delayed.

## Expected Behavior

An UNAVAILABLE cache read raises `ResourceNotReadyError` immediately; the event loop is never blocked by the state read path.

## Acceptance Criteria

- [ ] `_retry_on_not_ready` is removed; `get_state`/`yield_domain_states` raise `ResourceNotReadyError` on the first UNAVAILABLE check (i.e. behave like `get_state_once`, whose docstring contrast then needs updating)
- [ ] No `time.sleep`/sync tenacity waits remain on the StateProxy read path
- [ ] `yield_domain_states`' eager-check-then-generator structure is preserved or simplified now that the decorator constraint (see its docstring) is gone
- [ ] Existing not-ready tests updated: same exception, no retry delay

NodeJSmith:
## Summary

`StateProxy`'s sync read path (`get_state`, `yield_domain_states`) was wrapped in a **sync** tenacity retry — 5 attempts, `wait_exponential_jitter(0.01..0.1)` — that could never succeed.

Cache freshness is loop-confined state: it only transitions when async code (`_run_synchronization`, `on_disconnect`) runs on the event loop. But these reads run on that same event loop thread, and sync tenacity waits are `time.sleep()`. So the retry blocked the very thread that would have to change the condition it was polling. Every attempt re-read frozen state, added ~0.3-0.4s of event-loop blockage, and then raised the same `ResourceNotReadyError` the first check would have raised immediately.

Two knock-on costs:

- **Shutdown budget.** The one place app code can actually hit this is an `on_shutdown` hook reading state after `StateProxy.on_shutdown` reset freshness to `UNAVAILABLE` — burning ~0.3-0.4s of the shutdown hooks pool per read, on the loop thread, exactly when budget math matters most.
- **Framework instrumentation.** The Tier 2 block-IO guard patches `time.sleep`, so the framework's own read path tripped its own blocking-call detector.

## What changed

- Removed `_retry_on_not_ready` and its `MAX_RETRY_ATTEMPTS` / `RETRY_ON_NOT_READY_*` constants, plus the now-unused `tenacity` import. An `UNAVAILABLE` cache raises on the first check.
- Kept `yield_domain_states`' eager-check-then-nested-generator structure, but restated its rationale independently of the removed decorator. The eager check still earns its place: collapsing it into a plain generator function would defer the check to the first `next()`, so a caller that builds the generator without iterating it would silently read an unavailable cache instead of raising.
- Updated the `get_state` / `get_state_once` docstrings, which previously contrasted the two on retry behavior that no longer exists.
- Replaced `test_state_proxy_yield_retry.py` with `test_state_proxy_not_ready_reads.py`, which pins both surviving invariants: the eager raise, and single-check-no-retry across all four read entry points (`get_state`, `get_state_once`, `yield_domain_states`, `get_domain_states`). The no-retry test counts readiness checks rather than timing the call, so it's deterministic.

App-author DX is unaffected. Readiness guarantees come from `AppBootstrapCoordinator` gating (apps don't start until initial state capability commits), not from this retry. Normal runtime never sees `UNAVAILABLE` either — post-bootstrap disconnects degrade to `STALE`, which reads fine.

The async synchronization retry machinery (`_schedule_retry` / `_compute_retry_delay`) is untouched. That path is genuinely async and non-blocking; only the sync read-path decorator is removed.

## Testing

- `uv run nox -s dev` — 7720 passed, 1 skipped, 2 xfailed
- `prek -a` — all hooks pass
- `prek -a --hook-stage pre-push` — all hooks pass, including pyright and the schema-freshness check

## Follow-up (not in this PR)

With the retry gone, `get_state` and `get_state_once` are behaviorally identical — `get_state` is a one-line delegation. They were only ever distinguished by retry behavior. Worth collapsing to one name, but that's an API-surface change with callers in `tests/integration/test_state_proxy.py`, so it's left out of this fix.

Closes #1803


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #1963, opened 2026-09-08
test: clean up boilerplate in unified-handler-row tests

### LINKED ISSUE #1906: Clean up test boilerplate in unified-handler-row.test.tsx
Automated quality scan of `unified-handler-row.test.tsx` surfaced a few mechanical cleanups. One related finding (duplicated `makeListenerItem`/`makeJobItem` shape) is already tracked by #1282 and is excluded here to avoid overlap.

## Key Items

- **Redundant comment** (`unified-handler-row.test.tsx:189`): `// Error message shown` sits directly above `expect(errSubline).not.toBeNull()`, restating what the next two assertion lines already say. It's the only inline comment in the file. Delete it.
- **Repeated render boilerplate** (~30 call sites, e.g. lines 50, 56, 66-67, 79, 85, 91, 97-98, 104, 111, 117, 166, 172, 188, 200, 217, 228, 236-261, 268-325): `render(<UnifiedHandlerRow item={item} isSelected={false} onSelect={() => {}} />)` is repeated near-verbatim throughout the file, with `isSelected`/`onSelect` only varying in the handful of tests that specifically exercise selection/click behavior (lines 111, 121-150). Extract a `renderRow(item, overrides?)` helper defaulting `isSelected={false}` and a shared no-op `onSelect`, so the tests that actually vary those props stand out instead of blending in.
- **Raw `data-testid` string literals repeated with no shared constants** (e.g. `"handler-row-desc"` at lines 80, 218; `"handler-row-subline-err"` at lines 191, 201, 219; `"handler-row-mode-chip"` at lines 237, 243, 249, 255, 261; `"schedule-status-badge"` at lines 307, 313, 319, 325; `"handler-row-next-run"` at line 229): each testid is inlined at every call site. Renaming any of these in the component requires a manual find/replace across this file with no compiler safety net. Extract to local named constants (or a small shared testid map if other test files in this directory share the same ids).
- **Unexplained magic number** (`unified-handler-row.test.tsx:225`): `Math.floor(Date.now() / 1000) + 60` uses a bare `60` for "60 seconds in the future" with nothing indicating why 60 was chosen. Name it (e.g. `const FUTURE_OFFSET_SECONDS = 60`) or extract a small `futureUnixTimestamp(offsetSeconds)` helper.
- **Inconsistent parameter typing** (`unified-handler-row.test.tsx:15` and `:30`): `makeListenerItem`/`makeJobItem` both take `(overrides = {}, itemOverrides: ItemOverrides = {})` — `itemOverrides` is explicitly typed but the sibling `overrides` param (passed through to `createListener`/`createJob`) is left implicitly typed. Both serve the same "override bag" role; give `overrides` an explicit type (or a shared alias) for consistency.

## Acceptance Criteria

- [ ] Redundant comment at line 189 removed
- [ ] Repeated `render(...)` calls extracted into a shared `renderRow()` helper (or equivalent) and call sites updated
- [ ] Repeated `data-testid` string literals extracted into named constants
- [ ] Magic `60` offset at line 225 named or extracted into a helper
- [ ] `overrides` parameter on `makeListenerItem`/`makeJobItem` given an explicit type consistent with `itemOverrides`
- [ ] `cd frontend && npm run test:coverage` still passes

NodeJSmith:
## Summary

Test-only cleanup of `unified-handler-row.test.tsx`, surfaced by an automated quality scan.
No production code changes. Two assertions changed what they check (see "Assertion changes"
below); everything else is mechanical.

## What changed

- **Extracted a `renderRow()` helper.** `render(<UnifiedHandlerRow item={item} isSelected={false} onSelect={() => {}} />)`
  was repeated near-verbatim at ~30 call sites. `renderRow(item, overrides?)` defaults
  `isSelected` to `false` and supplies a shared module-level `noop` for `onSelect`, so the
  handful of tests that genuinely vary those props (selection state, click/keyboard handlers)
  now stand out instead of blending into the boilerplate. Its `RowPropOverrides` type is
  derived from the component's own props via `ComponentProps`, so a prop rename cannot drift
  past the helper.
- **Named the repeated `data-testid` literals** as `SUBLINE_ERR_TESTID`, `MODE_CHIP_TESTID`,
  `NEXT_RUN_TESTID`, and `SCHEDULE_STATUS_BADGE_TESTID`, matching the `*_TESTID` suffix form
  already used across this test suite. Renaming a testid in the component previously meant a
  manual find/replace through this file with no compiler safety net.
- **Shared the row testid helper.** `rowTestId(kind, id)` and `ROW_TESTID_PREFIX` now live in
  `unified-row.test-helpers.ts`, following the existing `handler-health.test-helpers.ts`
  precedent. `handler-list.test.tsx` had independently defined an identically named helper for
  the same `unified-row-<kind>-<id>` format; it now imports the shared one, so the format has a
  single source of truth on the test side. The helper is typed with the component's exported
  `UnifiedItemKind` rather than a re-declared string union. `handlers-tab.navigation.test.tsx`,
  which hardcoded the same literals, now uses the helper too — the format is defined in exactly
  one place on the test side.
- **Named the magic `60`** in the job `next_run` fixture as `FUTURE_OFFSET_SECONDS`, with a
  docstring explaining it is the offset past "now" for a job that has not fired yet.
- **Typed the override bags.** `makeListenerItem`/`makeJobItem` took `(overrides = {}, itemOverrides: ItemOverrides = {})`
  — the second param was typed but the first was implicitly `{}`. Both now carry explicit
  types (`Partial<ListenerData>` / `Partial<JobData>`), matching their sibling.
- **Dropped the one redundant inline comment** (`// Error message shown`), which restated
  the two assertion lines directly beneath it.

## Assertion changes

Two tests asserted that a `data-testid="handler-row-desc"` element was absent. That testid does
not exist anywhere in `unified-handler-row.tsx`, so both checks were vacuously true and tested
nothing. They were not ported mechanically:

- `does not render subtitle when humanDescription is null` is now
  `omits the description from the aria-label when humanDescription is null`, and asserts the
  `aria-label` the component actually produces.
- `does not show a description line for healthy handlers` drops the dead check and keeps its two
  real assertions.

Net effect: 36 test cases before and after; total `expect(` statements go from 43 to 42, with no
real coverage lost.

The duplicated `makeListenerItem`/`makeJobItem` item-shape construction is deliberately left
alone — it is already tracked by #1282, and touching it here would overlap that work.

## Testing

- `unified-handler-row.test.tsx` (36) and `handler-list.test.tsx` (9) — all pass
- `npx vitest run` — 110 files / 1601 tests pass
- `npx tsc --noEmit` and `npx eslint` — clean
- `prek -a` — all hooks pass

The `pr-screenshots` guard does not apply here: `check_pr_screenshots.py` explicitly excludes
`*.test.tsx` from its rendered-file set, and `unified-row.test-helpers.ts` is not a rendered file.

Closes #1906

### DIFF
--- a/frontend/src/components/app-detail/handler-list.test.tsx
+++ b/frontend/src/components/app-detail/handler-list.test.tsx
@@ -3,18 +3,10 @@ import { describe, expect, it, vi } from "vitest";
 
 import { createJob, createListener } from "../../test/factories";
 import { HandlerList } from "./handler-list";
+import type { UnifiedItemKind } from "./unified-handler-row";
+import { ROW_TESTID_PREFIX, rowTestId } from "./unified-row.test-helpers";
 
 const HANDLER_LIST_TEST_ID = "handler-list";
-
-// vi.hoisted so the row test-id convention is shared between the hoisted vi.mock
-// factory below and the assertions that read it.
-const { ROW_TEST_ID_PREFIX, rowTestId } = vi.hoisted(() => {
-  const prefix = "unified-row-";
-  return {
-    ROW_TEST_ID_PREFIX: prefix,
-    rowTestId: (kind: string, id: number) => `${prefix}${kind}-${id}`,
-  };
-});
 
 // Mock UnifiedHandlerRow to isolate HandlerList behavior — the row component
 // calls query hooks which require the Zustand app store (useAppStore) and MSW.
@@ -23,7 +15,7 @@ vi.mock("./unified-handler-row", () => ({
     item,
     isSelected,
   }: {
-    item: { kind: string; id: number; name: string; humanDescription: string | null };
+    item: { kind: UnifiedItemKind; id: number; name: string; humanDescription: string | null };
     isSelected: boolean;
     onSelect: () => void;
   }) => (
@@ -108,7 +100,7 @@ describe("HandlerList", () => {
     const { container } = render(
       <HandlerList listeners={listeners} jobs={jobs} selectedId={null} onSelect={() => {}} />,
     );
-    const rows = container.querySelectorAll(`[data-testid^='${ROW_TEST_ID_PREFIX}']`);
+    const rows = container.querySelectorAll(`[data-testid^='${ROW_TESTID_PREFIX}']`);
     expect(Array.from(rows, (row) => row.getAttribute("data-testid"))).toEqual([
       rowTestId("listener", 2),
       rowTestId("job", 6),
--- a/frontend/src/components/app-detail/handlers-tab.navigation.test.tsx
+++ b/frontend/src/components/app-detail/handlers-tab.navigation.test.tsx
@@ -7,6 +7,7 @@ import { createWouterMock } from "../../test/mock-wouter";
 import { renderWithAppState } from "../../test/render-helpers";
 import { HandlersTab } from "./handlers-tab";
 import { renderHandlersTab } from "./handlers-tab.test-helpers";
+import { rowTestId } from "./unified-row.test-helpers";
 
 // Mock child components that make API calls
 vi.mock("../shared/execution-table", () => ({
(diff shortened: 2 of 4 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #1966, opened 2026-09-08
refactor: name grid column widths and unify token typing in spacing-tokens showcase

### LINKED ISSUE #1924: Clean up style inconsistencies in spacing-tokens showcase component
## Description

Style and hygiene findings in the spacing-tokens design showcase component, surfaced by a quality scan. All are pre-existing on main.

## Key Items

- **Inconsistent `.map()` parameter naming** (`spacing-tokens.tsx:74`): The three token-list `.map()` callbacks use inconsistent parameter names — `token` (line 56), `r` (line 74), `shadow` (line 92). `r` should be `radius` to match the descriptive style of its siblings.

- **Inconsistent type annotations on token arrays** (`spacing-tokens.tsx:15,34,42`): `SPACING` is explicitly typed via the `SpacingToken` interface, but `RADII` and `SHADOWS` are bare inferred-type array literals with no named interface despite serving the same role (typed token lists rendered in loops). Either all three should have named interfaces or none should.

- **Magic pixel values in grid template** (`spacing-tokens.tsx:57`): `grid-cols-[100px_40px_minmax(0,1fr)]` hardcodes two layout pixel values. In a file whose purpose is showcasing tokenized design values, unnamed pixel values are an inconsistency. Extract to named constants or a shared grid-template class.

## Acceptance Criteria

- [ ] Rename `.map()` parameter `r` to `radius` on line 74
- [ ] Either add named interfaces for `RADII` and `SHADOWS` to match `SpacingToken`, or remove the `SpacingToken` interface and let all three infer (adding interfaces preferred for consistency)
- [ ] Extract `100px` and `40px` grid column widths to named constants

NodeJSmith:
## Summary

Style and consistency cleanup in the spacing-tokens design showcase component. Pure refactor — no behavior or rendering change.

The component renders three token lists (spacing scale, border radii, shadows) but treated them inconsistently: only one had a named interface, one `.map()` callback used a single-letter parameter, and the spacing grid hardcoded two unnamed pixel widths in a file whose entire purpose is showcasing tokenized design values.

## What changed

All three changes are in `frontend/src/components/design/spacing-tokens.tsx`:

- **Consistent `.map()` parameter naming.** Renamed the `RADII.map()` parameter from `r` to `radius`, so all three token-list callbacks (`token`, `radius`, `shadow`) use descriptive names.
- **Consistent type annotations.** Added `RadiusToken` and `ShadowToken` interfaces to match the existing `SpacingToken`, so every token list is explicitly typed rather than only the first.
- **Named grid column widths.** Extracted the `100px` / `40px` spacing-row column widths into `SPACING_VAR_COLUMN_WIDTH` and `SPACING_PX_COLUMN_WIDTH`.

### Why the grid moved to an inline style

Tailwind resolves arbitrary values by scanning source text for the literal class string. A `grid-cols-[...]` class assembled at runtime from the extracted constants would never be seen by the scanner and would emit no CSS, silently breaking the layout. The row therefore uses an inline `gridTemplateColumns` built from the same constants.

The computed value is byte-for-byte equivalent to what the old class produced (`100px 40px minmax(0, 1fr)`), the class had no responsive or state variants, and nothing else targeted it — so the rendered layout is unchanged.

## Testing

- `npm run build` (frontend) — passes
- `npm test` (frontend) — 1601 tests across 110 files pass
- `prek -a` — all hooks pass, including `tsc`, `eslint`, `stylelint`, and `prettier`

### Note on the Python suite

`uv run nox -s dev` reported one failure on each of two local runs, but a **different** test each time — `test_service_watcher.py::test_always_failing_service_stops_after_max_attempts` on the first run, `test_file_watcher.py::test_event_emitted_on_file_change` on the second. Both are timing-sensitive integration tests, and the first passes in isolation.

These are not related to this change: the branch touches exactly one `.tsx` file and zero Python files. A full `nox -s dev` run against `origin/main` in a separate worktree was green (7729 passed), so this is local flakiness under `-n 4` parallel load rather than a regression. Flagging it rather than dropping it — `test_always_failing_service_stops_after_max_attempts` asserts `shutdown_event.is_set()` immediately after awaiting the failure event, with no `wait_for` guard, which looks like a genuine latent race in the test itself.

## Visual evidence

Labeled `no-visual-change`. This is a naming/typing refactor plus an equivalent grid-template rewrite; the rendered output is identical, and this component is not covered by the `docs/screenshots.yml` manifest.

Closes #1924

### DIFF
--- a/frontend/src/components/design/spacing-tokens.tsx
+++ b/frontend/src/components/design/spacing-tokens.tsx
@@ -12,6 +12,27 @@ interface SpacingToken {
   px: number;
 }
 
+interface RadiusToken {
+  name: string;
+  cssVar: string;
+  px: number;
+}
+
+interface ShadowToken {
+  name: string;
+  cssVar: string;
+  label: string;
+}
+
+/**
+ * Column widths for the spacing scale rows, applied as an inline grid template.
+ * Tailwind resolves arbitrary values by scanning source text, so a `grid-cols-[...]`
+ * class assembled from these constants would never emit CSS.
+ */
+const SPACING_VAR_COLUMN_WIDTH = "100px";
+const SPACING_PX_COLUMN_WIDTH = "40px";
+const SPACING_ROW_TEMPLATE = `${SPACING_VAR_COLUMN_WIDTH} ${SPACING_PX_COLUMN_WIDTH} minmax(0, 1fr)`;
+
 const SPACING: SpacingToken[] = [
   { name: "px", cssVar: "--sp-px", px: 1 },
   { name: "0", cssVar: "--sp-0", px: 2 },
@@ -31,7 +52,7 @@ const SPACING: SpacingToken[] = [
 
 const MAX_SPACING_PX = SPACING[SPACING.length - 1].px;
 
-const RADII = [
+const RADII: RadiusToken[] = [
   { name: "sm", cssVar: "--r-sm", px: 6 },
   { name: "md", cssVar: "--r-md", px: 8 },
   { name: "lg", cssVar: "--r-lg", px: 12 },
@@ -39,7 +60,7 @@ const RADII = [
   { name: "pill", cssVar: "--r-pill", px: 999 },
 ];
 
-const SHADOWS = [
+const SHADOWS: ShadowToken[] = [
   { name: "shadow-1", cssVar: "--shadow-1", label: "Subtle" },
   { name: "shadow-2", cssVar: "--shadow-2", label: "Medium" },
   { name: "shadow-3", cssVar: "--shadow-3", label: "Elevated" },
@@ -54,7 +75,11 @@ export function SpacingTokens() {
         <h3 className={designGroupLabelClassName}>Spacing Scale</h3>
         <div className="flex flex-col gap-2">
           {SPACING.map((token) => (
-            <div key={token.cssVar} className="grid grid-cols-[100px_40px_minmax(0,1fr)] items-center gap-3">
+            <div
+              key={token.cssVar}
+              className="grid items-center gap-3"
+              style={{ gridTemplateColumns: SPACING_ROW_TEMPLATE }}
+            >
               <code className={designTokenCodeClassName}>{token.cssVar}</code>
               <span className="text-right font-mono text-xs text-foreground-secondary">{token.px}px</span>
               <div className="h-2 overflow-hidden rounded-sm bg-muted">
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
## Automated review battery — 6 reviewers

**Verdict: no blocking findings.** 0 CRITICAL, 0 HIGH. No code changes applied.

| Reviewer | Result |
|---|---|
| lazy-checker | CLEAN — 0 findings |
| integration-reviewer | WARN — 2 MEDIUM |
| wtf-reviewer | 2 LOW |
| llm-checker | 3 soft findings |
| nitpicker | 6 style findings |

CodeRabbit review completed with zero threads.

### Findings reviewed and intentionally not actioned

Three reviewers converged on the same two observations. Both were verified against the code and judged out of scope for this PR:

1. **The inline `gridTemplateColumns` style replaces a working static Tailwind class.** Accurate: the row template was already a static literal and did not need runtime assembly, so leaving Tailwind was self-imposed by the decision to name the widths. But naming those two widths is the explicit acceptance criterion in #1924, and the alternatives (dropping the constants, or promoting the widths to global CSS custom properties) either fail that criterion or add more machinery than the finding removes. The file already uses inline `style` in three other places (`width`, `borderRadius`, `boxShadow`), so this is not a new pattern within the component, and the computed value is byte-for-byte equivalent to the old class.

2. **`RadiusToken` is structurally identical to `SpacingToken`.** True — both are `{ name, cssVar, px }`. Kept distinct deliberately: spacing and radius are different domain concepts in a component whose entire purpose is distinguishing token categories, and merging them into one shape-neutral interface would couple two concepts that only coincidentally share a shape today.

Lower-severity nits (the unrelated `100px` literal at line 98, `gap-7` at line 116) are pre-existing on `main` and outside this issue's scope.

### CI

`test (3.13)` failed on `tests/unit/resources/lifecycle/test_total_timeout.py::test_root_identity_uses_total_timeout_not_resource_timeout`. Not introduced by this branch — the diff is one `.tsx` file and zero Python files, and the test file is byte-identical to `main`. The test asserts against real wall-clock time with a 0.1s margin (a 0.2s shutdown sleep against a 0.3s total timeout), which is thin under `-n 4` on a loaded runner. It passes locally (6/6 in the file). `main` itself had an unrelated timing failure in the same window (`test_always_failing_service_stops_after_max_attempts`, run 34216623295), so the suite has more than one latent real-clock race of this shape. Failed jobs have been rerun.

`file-sizes` and `duplicate-code` are the known `continue-on-error` baseline failures and are non-blocking.

PR #1972, opened 2026-09-08
test: name duplicated toggle locators in service-row.test.tsx

### LINKED ISSUE #1946: Clean up duplicated toggle locators in service-row.test.tsx
## Description

Automated quality scan of `frontend/src/components/diagnostics/service-row.test.tsx`. Two of three checkers (llm-checker, lazy-checker) reported the file clean — no comment bloat, no deferred debt, no dead code. The nitpicker surfaced locator duplication that survived validity assessment.

This is pre-existing, mechanical, and low-risk: a same-shape cleanup within one test file, no behavior change.

### 1. Toggle-button locators are duplicated inline instead of named once

The accessible-name regexes for the exception toggle are re-typed throughout the file:

- `/show exception/i` — `service-row.test.tsx:23`, `:33`, `:42`
- `/hide exception/i` — `service-row.test.tsx:38`, `:40`

Five inline `getByRole("button", { name: ... })` call sites for what is a single button with two labels.

### 2. Locator reference is named on one line and re-fetched inline on the next

`service-row.test.tsx:33` binds the button to `const toggle`, then `:40` re-queries the same DOM node inline (`await user.click(getByRole("button", { name: /hide exception/i }))`) rather than reusing `toggle`. React reuses the DOM node across the toggle — only `textContent` and `aria-expanded` change — so `toggle` is still a valid reference at that point.

Scoped note: the re-queries at `:38` and `:42` are **not** redundant. `getByRole` throws when no match is found, so re-querying by the flipped name is itself the assertion that the accessible name changed. Those should stay. Only `:40` is a genuine reuse miss.

## Proposed Fix

Extract the two accessible names as file-local constants and reuse the bound `toggle` where the element is only being clicked:

```tsx
const SHOW_EXCEPTION = { name: /show exception/i } as const;
const HIDE_EXCEPTION = { name: /hide exception/i } as const;
```

Then use `getByRole("button", SHOW_EXCEPTION)` / `getByRole("button", HIDE_EXCEPTION)` at the sites that assert the label flip, and `await user.click(toggle)` at `:40`.

## Acceptance Criteria

- [ ] `/show exception/i` and `/hide exception/i` each appear exactly once in the file, as named constants
- [ ] `service-row.test.tsx:40` clicks the already-bound `toggle` rather than re-querying
- [ ] The label-flip assertions at `:38` and `:42` still query by the flipped accessible name (behavior preserved, not collapsed into `toggle`)
- [ ] `cd frontend && npx vitest run src/components/diagnostics/service-row.test.tsx` passes

## Assessed and not filed

Two further nitpicker findings were dropped under the validity protocol, recorded here so a later scan does not re-raise them:

- **"`makeService` defaults should move to a shared factory"** — the shared-factory convention cited (`tests/support/factories.py`, `.claude/rules/test-conventions.md`) is scoped to the Python `tests/` tree; there is no frontend equivalent registry. `makeService` has exactly one consumer (`grep -rn "makeService" frontend/src` returns only this file). Speculative future duplication, not present debt.
- **"The second test covers both open and close in one `it` block"** — the test name states the round-trip explicitly (`toggles the exception text open and closed`), and the close-phase assertions require the open phase as setup. Splitting would duplicate setup without adding coverage.

NodeJSmith:
## Summary

Removes duplicated accessible-name locators in `service-row.test.tsx`. The exception-toggle button was located by re-typed inline regexes at five call sites (`/show exception/i` three times, `/hide exception/i` twice), and one site re-queried a DOM node that was already bound to a local.

Surfaced by an automated quality scan of the file. Pre-existing, mechanical, and scoped to a single test file — no production code touched, no behavior change.

## What changed

- Extracted the two accessible names as file-local constants, `SHOW_EXCEPTION` and `HIDE_EXCEPTION`, so each regex is written exactly once.
- The close click now reuses the already-bound `toggle` element instead of re-querying for it. React reuses the same DOM node across the toggle — only `textContent` and `aria-expanded` change — so the binding stays valid.

Deliberately preserved: the two assertions that re-query by the *flipped* accessible name. `getByRole` throws when nothing matches, so re-querying is itself the assertion that the accessible name changed on toggle. Collapsing those into `toggle` would have quietly dropped that coverage. A comment now records why.

## Testing

- `npx vitest run src/components/diagnostics/service-row.test.tsx` — 2 passed
- Full frontend suite — 110 files, 1603 passed
- `prek -a` — all hooks pass; `prek pyright -a --stage pre-push` — pass

The `test:` commit type keeps this out of the generated changelog, which is correct for a test-only cleanup.

Closes #1946

### DIFF
--- a/frontend/src/components/diagnostics/service-row.test.tsx
+++ b/frontend/src/components/diagnostics/service-row.test.tsx
@@ -5,6 +5,9 @@ import { describe, expect, it } from "vitest";
 import type { MergedService } from "./merge-services";
 import { ServiceRow } from "./service-row";
 
+const SHOW_EXCEPTION = { name: /show exception/i } as const;
+const HIDE_EXCEPTION = { name: /hide exception/i } as const;
+
 function makeService(overrides: Partial<MergedService> = {}): MergedService {
   return {
     resource_name: "bus",
@@ -20,7 +23,7 @@ function makeService(overrides: Partial<MergedService> = {}): MergedService {
 describe("ServiceRow", () => {
   it("does not render the exception toggle when there is no exception", () => {
     const { queryByRole } = render(<ServiceRow service={makeService({ exception: null })} />);
-    expect(queryByRole("button", { name: /show exception/i })).toBeNull();
+    expect(queryByRole("button", SHOW_EXCEPTION)).toBeNull();
   });
 
   it("toggles the exception text open and closed", async () => {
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #1975, opened 2026-09-08
chore: rename TelemetryCounters to TelemetryHealth

### LINKED ISSUE #1951: Rename TelemetryCounters — the type carries a degradation flag, not only counters
Quality scan of `frontend/src/components/diagnostics/telemetry-panel.tsx`. Three checkers ran (llm-checker, lazy-checker, nitpicker); the file is otherwise clean — one finding survived validity assessment.

## Finding: `TelemetryCounters` is not a counters type

`frontend/src/components/diagnostics/telemetry-panel.tsx:6-12` exports an interface named `TelemetryCounters` that holds four numeric counters plus `telemetryDegraded: boolean` — a status flag, not a count. The doc comment on line 5 concedes the mismatch in its own wording ("Telemetry drop counters **and degradation flag**"), which is the tell: the name needed a qualifier to describe what the type actually holds.

The consequence is at the consumer, not here. `frontend/src/pages/diagnostics.tsx:45` declares `telemetry: TelemetryCounters`, so a reader who reaches for that field expecting a bag of numbers gets a mixed shape. The type also does double duty as `TelemetryPanel`'s props type (line 20), so the name has to read correctly in both roles.

**Proposed fix:** rename the interface to `TelemetryHealth` and update the three references — the declaration and the props annotation in `telemetry-panel.tsx`, and the import plus field type at `frontend/src/pages/diagnostics.tsx:9,45`. `TelemetryHealth` covers counters and the degradation flag together and still reads correctly at the `telemetry:` field site, where `TelemetryPanelProps` would not. Once renamed, the qualifier in the line 5 doc comment can shorten accordingly.

No human judgment needed: the rename is mechanical, confined to three call sites, and changes no behavior.

## Not filed

Four other findings were assessed and dropped as invalid:

- **Inline `data-testid` string literals** — `diag-*` testids are written inline throughout the frontend; centralizing them is a codebase-wide convention change, not a defect in this file.
- **Tailwind class ordering on line 29** — the repo has no enforced class-order convention and no `prettier-plugin-tailwindcss`, so there is no convention being violated.
- **`telemetryDegraded` should be `isTelemetryDegraded`** — the prop deliberately mirrors the app store's field name (`frontend/src/state/store.ts:60,78`), which is read the same way in `system-health.tsx`, `use-telemetry-health.ts`, and `alert-banner`. Renaming it here alone would break that alignment.
- **`title="telemetry health"` lowercase vs `ariaLabel="Telemetry health"`** — deliberate and documented. `frontend/src/components/diagnostics/panel.tsx:21` specifies the heading is "rendered lowercase as authored" and line 23 notes the aria label is "spelled out for screen readers, so not always `title`".

NodeJSmith:
## Summary

Renames the frontend `TelemetryCounters` interface to `TelemetryHealth`, and collapses it together with the store's structurally identical `TelemetryHealthFields` into a single type.

The interface holds four drop counters (`droppedOverflow`, `droppedExhausted`, `droppedShutdown`, `errorHandlerFailures`) plus `telemetryDegraded` — a boolean status flag, not a count. The doc comment had to say "drop counters **and degradation flag**" to describe the shape, which is the tell that the name was covering only part of it.

The mismatch showed up at the consumer rather than the declaration: `diagnostics.tsx` declares `telemetry: TelemetryCounters`, so a reader reaching for that field expecting a bag of numbers got a mixed shape instead. The type also does double duty as `TelemetryPanel`'s props, so `TelemetryPanelProps` was not an option — the name has to read correctly in both roles. `TelemetryHealth` covers counters and the flag together and reads correctly at the `telemetry:` field site.

It also lines the type up with the naming already used across the rest of the telemetry surface: `useTelemetryHealth`, `setTelemetryHealth`, and the panel's own `title="telemetry health"`.

## Deduplicating against the store

Review surfaced that the rename landed one suffix away from an existing type: `store.ts` already declared `TelemetryHealthFields` with the same five fields, hand-synced with the panel's copy. `TelemetryHealth` vs `TelemetryHealthFields` reads as a typo or a subset rather than two independent declarations of one shape, so the two are now a single exported `TelemetryHealth` living in `store.ts` — the module that owns the shape, since `setTelemetryHealth` writes it and `useTelemetryHealth` populates it from the API. The panel and the diagnostics page both reference it. Importing a type from the store is the established pattern here (`ConnectionStatus`, `ServiceStatusEntry`, `AppStatusEntry`, `TimePreset`), including `merge-services.ts` in the same directory.

The doc comment is also corrected while being rewritten. It claimed these fields are "carried on the WS stream"; they are not. `useTelemetryHealth` polls `GET /api/telemetry/status` on a 30s–120s backoff and is the only caller of `setTelemetryHealth`, so there is no WebSocket in this data path. The claim predated this PR, but the PR rewrites that exact line.

## Changes

- `frontend/src/state/store.ts` — renamed `TelemetryHealthFields` to `TelemetryHealth` (now the single declaration of this shape) and updated `setTelemetryHealth`'s parameter type.
- `frontend/src/components/diagnostics/telemetry-panel.tsx` — deleted the duplicate interface; the props annotation now references the store's type.
- `frontend/src/pages/diagnostics.tsx` — the type import moves to `../state/store`; the panel import is value-only.

Type-only changes throughout. No runtime behavior, no rendered output, and no API surface changes.

## Testing

- `npx tsc --noEmit` — clean
- `npm run test` (frontend) — 110 files, 1603 tests passed
- `npx eslint src --max-warnings=0` — clean
- `prek -a` and `prek pyright -a --hook-stage pre-push` — all hooks pass
- `uv run nox -s dev` — 7737 passed, 1 skipped, 2 xfailed (first commit)

Verified no `TelemetryCounters` or `TelemetryHealthFields` references remain anywhere in the repo.

Labeled `no-visual-change`: this renames TypeScript identifiers only, so no pixels move and there is no screenshot to regenerate.

Closes #1951

### DIFF
--- a/frontend/src/components/diagnostics/telemetry-panel.tsx
+++ b/frontend/src/components/diagnostics/telemetry-panel.tsx
@@ -1,15 +1,7 @@
+import type { TelemetryHealth } from "../../state/store";
 import { AlertShell } from "../shared/alert-shell";
 import { DropCounterRow } from "./drop-counter-row";
 import { Panel, PANEL_BANNER_CLASS } from "./panel";
-
-/** Telemetry health as carried on the WS stream: drop counters and the degradation flag. */
-export interface TelemetryHealth {
-  droppedOverflow: number;
-  droppedExhausted: number;
-  droppedShutdown: number;
-  errorHandlerFailures: number;
-  telemetryDegraded: boolean;
-}
 
 export function TelemetryPanel({
   droppedOverflow,
--- a/frontend/src/components/diagnostics/telemetry-panel.tsx
+++ b/frontend/src/components/diagnostics/telemetry-panel.tsx
@@ -2,8 +2,8 @@ import { AlertShell } from "../shared/alert-shell";
 import { DropCounterRow } from "./drop-counter-row";
 import { Panel, PANEL_BANNER_CLASS } from "./panel";
 
-/** Telemetry drop counters and degradation flag, as carried on the WS stream. */
-export interface TelemetryCounters {
+/** Telemetry health as carried on the WS stream: drop counters and the degradation flag. */
+export interface TelemetryHealth {
   droppedOverflow: number;
   droppedExhausted: number;
   droppedShutdown: number;
@@ -17,7 +17,7 @@ export function TelemetryPanel({
   droppedShutdown,
   errorHandlerFailures,
   telemetryDegraded,
-}: TelemetryCounters) {
+}: TelemetryHealth) {
   return (
     <Panel title="telemetry health" ariaLabel="Telemetry health" data-testid="diag-telemetry-panel">
       {telemetryDegraded && (
--- a/frontend/src/pages/diagnostics.tsx
+++ b/frontend/src/pages/diagnostics.tsx
@@ -6,12 +6,12 @@ import { BootIssuesPanel } from "../components/diagnostics/boot-issues-panel";
 import { LoggingPanel } from "../components/diagnostics/logging-panel";
 import { type MergedService, mergeServices } from "../components/diagnostics/merge-services";
 import { ServicesPanel } from "../components/diagnostics/services-panel";
-import { type TelemetryHealth, TelemetryPanel } from "../components/diagnostics/telemetry-panel";
+import { TelemetryPanel } from "../components/diagnostics/telemetry-panel";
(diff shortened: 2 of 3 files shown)


NodeJSmith:
@coderabbitai review

PR #1976, opened 2026-09-08
refactor: deduplicate API error format and credentials literal in client.ts

### LINKED ISSUE #1967: Clean up duplicated error-message format and credentials literal in client.ts
## Description

A quality-scanner pass (llm-checker, lazy-checker, nitpicker) over `frontend/src/api/client.ts` surfaced two pre-existing hygiene findings. Both predate the current work. The llm-checker found nothing — the file's comments, including the `postSession` doc block, document real cross-file constraints rather than narrating history, so no comment-bloat cleanup is needed here.

### 1. The `API error: <status> <statusText>` fallback format is written twice

`ApiError`'s constructor already applies this exact fallback when no `message` is passed (`client.ts:12`), and `apiFetch` relies on that behavior by handing the optional `detail` straight to the constructor (`client.ts:41`). `postSession` re-derives the same string by hand instead (`client.ts:81`):

```ts
// client.ts:12
super(message ?? `API error: ${status} ${statusText}`);

// client.ts:81
const message = detail ?? `API error: ${response.status} ${response.statusText}`;
```

Changing the format (adding the request path, trimming the `statusText`, localizing it) requires editing both sites, and nothing fails if only one is updated — the two paths drift silently, so an error surfaced by `postSession` and one surfaced by `apiFetch` can start reading differently.

**Proposed fix:** extract a module-level formatter and call it from both places.

```ts
function formatApiError(status: number, statusText: string): string {
  return `API error: ${status} ${statusText}`;
}
```

`ApiError`'s constructor becomes `super(message ?? formatApiError(status, statusText))`; `postSession` becomes `const message = detail ?? formatApiError(response.status, response.statusText)`. This preserves current behavior exactly and leaves `postSession`'s deliberate bypass of `apiFetch` (documented at `client.ts:57-64`) intact — only the message-formatting duplication is removed, not the throw-vs-return distinction that bypass exists for.

### 2. `credentials: "same-origin"` is repeated across both fetch call sites

`client.ts:32` (`apiFetch`) and `client.ts:70` (`postSession`) each hardcode the literal, while the file's other shared request values (`BASE_URL`, `JSON_CONTENT_TYPE`) are already centralized at the top. A future change to the credentials mode has to find both call sites.

**Proposed fix:** add `const FETCH_CREDENTIALS: RequestCredentials = "same-origin";` alongside `BASE_URL`/`JSON_CONTENT_TYPE` and reference it from both fetch calls.

## Assessed and excluded

The nitpicker also flagged the inline `"/auth/session"` path at `client.ts:68` as a magic string needing a named constant. Excluded: the literal is used exactly once, inside the only function that talks to that route, and the route is already named in that function's doc comment one line above. Extracting it would add a layer of indirection without removing any duplication.

## Acceptance Criteria

- [ ] The `API error: <status> <statusText>` template appears exactly once in `client.ts`, used by both `ApiError`'s default message and `postSession`'s fallback.
- [ ] `credentials: "same-origin"` is defined once as a module-level constant and referenced by both `apiFetch` and `postSession`.
- [ ] `postSession` still returns `PostSessionResult` (never throws `ApiError`) and still bypasses `apiFetch`, preserving the login form's inline error rendering.
- [ ] Existing `client.test.ts` coverage passes unchanged — no behavior change to any error message text.

NodeJSmith:
## Summary

Removes two small duplications in `frontend/src/api/client.ts` that a quality scan flagged.

The `API error: <status> <statusText>` fallback string was written out twice — once as `ApiError`'s default message and once in `postSession`, which deliberately bypasses `apiFetch` and so had to build its own fallback. The two copies could drift independently. They now share a module-level `formatApiError(status, statusText)` helper.

The `credentials: "same-origin"` literal appeared in both `fetch` call sites. It is now a `FETCH_CREDENTIALS` constant, typed as `RequestCredentials`, sitting alongside the existing `BASE_URL` and `JSON_CONTENT_TYPE` constants at the top of the module.

## Why

Both are single-source-of-truth cleanups. The error-format duplication is the one with real drift risk: a future change to the fallback wording would need to be made in two places, and missing one would produce two different messages for the same condition depending on which code path produced it.

## Not changed

This is a pure refactor with no behavior change:

- The rendered message text is byte-identical.
- `postSession` still returns its `PostSessionResult` shape rather than throwing.
- `postSession` still bypasses `apiFetch` on purpose — that was left alone.

## Testing

- `cd frontend && npm test -- --run` — 110 files, 1603 tests passed
- `prek -a` — all hooks pass, including `eslint`, `prettier`, `stylelint`, and `tsc`

No visual evidence required: `client.ts` is not a rendered file, so it falls outside the `frontend/src/**/*.{tsx,css}` trigger in `tools/frontend/check_pr_screenshots.py`.

Closes #1967

### DIFF
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -2,6 +2,12 @@
 
 const BASE_URL = "/api";
 const JSON_CONTENT_TYPE = "application/json";
+const FETCH_CREDENTIALS: RequestCredentials = "same-origin";
+
+/** Fallback message for a non-ok response whose body carried no usable detail. */
+function formatApiError(status: number, statusText: string): string {
+  return `API error: ${status} ${statusText}`;
+}
 
 export class ApiError extends Error {
   constructor(
@@ -9,7 +15,7 @@ export class ApiError extends Error {
     public readonly statusText: string,
     message?: string,
   ) {
-    super(message ?? `API error: ${status} ${statusText}`);
+    super(message ?? formatApiError(status, statusText));
     this.name = "ApiError";
   }
 }
@@ -29,7 +35,7 @@ export async function apiFetch<T>(path: string, init?: RequestInit): Promise<T> 
   const url = `${BASE_URL}${path}`;
   const response = await fetch(url, {
     ...init,
-    credentials: "same-origin",
+    credentials: FETCH_CREDENTIALS,
     headers: {
       Accept: JSON_CONTENT_TYPE,
       ...init?.headers,
@@ -67,7 +73,7 @@ export async function postSession(token: string): Promise<PostSessionResult> {
   try {
     response = await fetch(`${BASE_URL}/auth/session`, {
       method: "POST",
-      credentials: "same-origin",
+      credentials: FETCH_CREDENTIALS,
       headers: { "Content-Type": JSON_CONTENT_TYPE },
       body: JSON.stringify({ token }),
     });
@@ -78,6 +84,6 @@ export async function postSession(token: string): Promise<PostSessionResult> {
   if (response.ok) return { ok: true };
 
   const detail = await extractErrorMessage(response);
-  const message = detail ?? `API error: ${response.status} ${response.statusText}`;
+  const message = detail ?? formatApiError(response.status, response.statusText);
   return { ok: false, message };
 }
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2015, opened 2026-09-08
docs: trim narrated history from ExecutionStatus and AppManifestsChangedData docstrings

### LINKED ISSUE #1978: Trim narrated-history and backend-narrative bloat from ExecutionStatus and AppManifestsChangedData docstrings
Surfaced by an automated quality scan of `frontend/src/api/ws-types.ts`. That file is generated (`@generated from ws-schema.json`), so **both fixes belong in the Python docstrings upstream**, followed by `uv run python scripts/export_schemas.py --types`. Do not hand-edit `ws-types.ts`.

Both findings are comment/docstring bloat that the codegen faithfully round-trips into the TypeScript artifact, where the surrounding context that made the prose make sense in Python is absent.

## 1. `ExecutionStatus` docstring narrates migration history and Pydantic internals

**Where:** `src/hassette/types/types.py:78-86` → `frontend/src/api/ws-types.ts:27-35` (also propagated to `generated-types.ts`)

**What's wrong:** The docstring for a five-member string enum spends four of its six lines on implementation history and framework trivia:

- *"migration 001 introduced the original four values (`success`, `error`, `cancelled`, `timed_out`); migration 009 added `skipped`"* — narrated history. Which migration added which member is recoverable from git and the migration files themselves; it is not something a reader needs in order to use the type, and it becomes stale the next time a value is added.
- *"Pydantic v2 coerces plain strings to enum members on construction and serialises back to plain strings in JSON responses."* — generic Pydantic v2 behavior that applies to every `StrEnum` in the codebase, not a property of this enum. It is meaningless in the generated TypeScript, where the type is already a plain string union.

This conflicts with the Self-Contained Comments rule in `coding-style.md` (state the durable invariant, not the point-in-time narrative).

**Proposed fix:** Reduce to the durable facts:

```python
class ExecutionStatus(StrEnum):
    """Status values for handler invocations and job executions.

    Must stay in sync with the ``executions.status`` CHECK constraint.
    """
```

Then regenerate (`uv run python scripts/export_schemas.py --types`) and commit the refreshed `ws-types.ts` / `generated-types.ts`.

## 2. `AppManifestsChangedData` docstring buries its one actionable instruction under backend narrative

**Where:** `src/hassette/schemas/domain_models.py:132-140` → `frontend/src/api/ws-types.ts:153-162`

**What's wrong:** Nine lines of prose for `Record<string, never>` (an empty payload). The load-bearing part for a consumer — *the payload is empty on purpose; treat receipt as "refetch, don't inspect"* — is genuinely useful and should stay, because the bare type communicates none of it. But it arrives last, after two sentences of backend implementation detail that a frontend consumer cannot act on: the internal constant name `HASSETTE_EVENT_APP_LOAD_COMPLETED`, the two distinct trigger scenarios (bootstrap/reload pass vs. metadata-only config edit), and the `display_name` example. The "carries no fields" fact is then restated three separate ways.

**Proposed fix:** Keep the semantics, drop the internal-trigger narrative:

```python
"""Payload for a manifest refresh broadcast over WebSocket.

Carries no fields and does not identify which apps changed — it is a refetch
signal, not a diff. Clients should treat receipt as "manifest state may be
stale, refetch" rather than inspect the payload.
"""
```

Then regenerate as above.

## Acceptance criteria

- [ ] `ExecutionStatus` docstring no longer references migration numbers or Pydantic coercion behavior
- [ ] `AppManifestsChangedData` docstring keeps the refetch-signal semantics and drops the internal-event/trigger-scenario narrative
- [ ] `uv run python scripts/export_schemas.py --types` re-run; regenerated `ws-types.ts` and `generated-types.ts` committed
- [ ] `prek -a` clean; schema-freshness check passes

## Excluded during validity review

One further finding was raised and rejected: that the `ExecutionCompletedData` docstring (`ws-types.ts:131-136`) merely restates what the type already encodes. It does not — the type declares `kind: "handler" | "job"` with `listener_id?: number | null` and `job_id?: number | null` as independent optionals. The correspondence the comment states (`listener_id` set when `kind='handler'`, `job_id` when `kind='job'`) is not expressible in that shape and is not derivable from the field declarations, so the comment carries real information. No change proposed there.

NodeJSmith:
## Summary

Trims two docstrings that carried narrated history and backend implementation detail into the generated TypeScript API types, where the surrounding Python context that made the prose coherent is absent.

Both docstrings are the upstream source for `ws-types.ts` and `generated-types.ts`, so the fix is made in Python and the generated artifacts are regenerated — not hand-edited.

### `ExecutionStatus` (`src/hassette/types/types.py`)

Four of six docstring lines were implementation history and framework trivia:

- Which migration introduced which member — recoverable from git and the migration files, and goes stale the next time a value is added.
- Generic Pydantic v2 string-coercion behavior that applies to every `StrEnum` in the codebase, and is meaningless in the generated TypeScript where the type is already a plain string union.

Reduced to the one durable constraint the type must satisfy: it must stay in sync with the `executions.status` CHECK constraint.

### `AppManifestsChangedData` (`src/hassette/schemas/domain_models.py`)

Nine lines of prose for an empty payload (`Record<string, never>`). The load-bearing instruction for a consumer — the payload is empty on purpose, so treat receipt as "refetch, don't inspect" — arrived last, after backend narrative a frontend consumer cannot act on: the internal event constant name, the two distinct trigger scenarios, and a `display_name` example. The "carries no fields" fact was restated three ways.

Kept the refetch-signal semantics (the bare type communicates none of it), dropped the internal trigger narrative.

## Changes

- `src/hassette/types/types.py` — `ExecutionStatus` docstring
- `src/hassette/schemas/domain_models.py` — `AppManifestsChangedData` docstring
- Regenerated via `uv run python scripts/export_schemas.py --types`: `frontend/openapi.json`, `frontend/ws-schema.json`, `frontend/src/api/generated-types.ts`, `frontend/src/api/ws-types.ts`, `frontend/src/api/ws-validator.generated.ts`

No behavior change — docstrings and their regenerated schema/type artifacts only.

## Testing

- `prek -a` — all hooks pass
- `prek -a --hook-stage pre-push` — all hooks pass, including `pyright` and **"Check schemas are up to date"**, which confirms the committed generated artifacts match a fresh codegen run

No frontend rendered files (`.tsx`, `.css`) changed, so there is no visual result to capture — only generated API type declarations.

Closes #1978


NodeJSmith:
@coderabbitai review

NodeJSmith:
Automated autofix review pass — reviewed at `54ce869`.

**CodeRabbit:** review completed with zero threads and zero review comments.

**Review battery** (code, integration, WTF, llm-checker, lazy-checker, nitpicker):

- **Fixed (1):** the trimmed `AppManifestsChangedData` docstring introduced "manifest **state** may be stale" where the repo's canonical collocation — and the pre-trim wording — is "manifest **status**" (`ManifestStatus` in `types/enums.py`, plus three frontend tests using "stale manifest status"). Reverted that one word and regenerated the WS artifacts.
- **Skipped — trimmed trigger narrative:** flagged as information loss (the removed "also fires on metadata-only config edits" detail). The consumer-facing instruction is trigger-independent — an empty payload means "refetch" either way — so the removed detail changes no consumer action. The related wording in `runtime_query_service.on_app_manifests_changed` ("after a full app load/reload pass") is **pre-existing on `main`**, untouched by this PR, and its cross-reference to `AppManifestsChangedData` still resolves. Matches the validity review already recorded in #1978.
- **Skipped — dropped Pydantic v2 coercion note:** generic `StrEnum` behavior, meaningless in the generated TypeScript where the type is already a plain string union. Per #1978's rationale.
- **Skipped — docstring wrap width (~80 vs ~90 chars):** `ruff format` clean, well under the 120 limit.

**Filed as follow-up:** #2039 — the retained `ExecutionStatus` invariant ("must stay in sync with the `executions.status` CHECK constraint") has no automated backstop; no test iterates `list(ExecutionStatus)` against the live constraint. Pre-existing gap, verified against `main`.

**Verification:** `tools/check_schemas_fresh.py` exit 0 (committed artifacts match a fresh codegen run); `prek -a` and `prek -a --hook-stage pre-push` including `pyright` all pass; 74 backend tests (`test_schema_freshness`, `test_runtime_query_service`, `test_ws_models`) and 26 frontend `use-websocket` tests pass. `duplicate-code` and `file-sizes` fail as expected baseline noise (`continue-on-error`).

Auto-merge (squash) enabled.

Issue #1984, opened 2026-09-08
Extract a manifest-serving, URL-capturing mock client helper in tests/unit/cli/test_client.py


NodeJSmith:
## Description

The instance-resolution tests in `tests/unit/cli/test_client.py` each hand-roll the same
`httpx.MockTransport` handler: serve the app-manifest list JSON for `MANIFESTS_ENDPOINT`, and an
empty `b"[]"` JSON array for every other request. Two of them then wrap that handler in a second
`tracking_handler` closure that appends `str(request.url)` to a local `captured_urls` list.

The file already has both halves as separate helpers: `make_manifest_list()` (`:105`) builds the
manifest payload, and `url_capturing_client()` (`:111`) builds a client whose request URLs are
recorded. What's missing is the one that does both.

**DECIDED (placement):** add a local helper, e.g. `manifest_resolving_client(instances, app_key=...) -> tuple[HassetteCLIClient, list[str]]`, directly beside `make_manifest_list()`/`url_capturing_client()` in `test_client.py` — not in `tests/unit/cli/conftest.py`, and not by extending `MockTransportBuilder`. Per `.claude/rules/test-conventions.md`, shared `tests/support/`-level placement is for patterns needed across 3+ files; this shape is currently only exercised within `test_client.py`'s own instance-resolution tests. If `test_commands_app.py` later needs the same shape, promote it then rather than speculatively building the shared version now.

## Where

- `tests/unit/cli/test_client.py:565-571` — `test_name_instance_resolves_to_index`
- `tests/unit/cli/test_client.py:594-600` — `test_unknown_instance_name_exits_nonzero`
- `tests/unit/cli/test_client.py:637-643` — `test_stopped_instance_still_resolves_by_name`
- `tests/unit/cli/test_client.py:672-678` — `test_ambiguous_instance_name_exits_nonzero`
- The `tracking_handler` closures at `:577-581` and `:647-651` duplicating `url_capturing_client()`

(Line numbers as of this update — shifted slightly from the original filing but the same four sites.)

## Acceptance Criteria

- [ ] `manifest_resolving_client(instances, app_key=...)` (or equivalent name) exists in `tests/unit/cli/test_client.py`, beside `make_manifest_list()`/`url_capturing_client()`
- [ ] All four instance-resolution tests use it instead of a local `def handler(...)` / `def tracking_handler(...)`
[...]

PR #2022, opened 2026-09-08
test: extract a raw-body mock client helper in cli client tests

### LINKED ISSUE #1982: Extract a raw-body mock transport helper in tests/unit/cli/test_client.py
## Description

`tests/unit/cli/test_client.py` hand-rolls the same mock-transport client setup over and over: an
inline `def handler(_req: httpx.Request) -> httpx.Response` that returns one fixed **raw**
(non-JSON-serializable) body, wrapped in `httpx.MockTransport(handler)`, passed straight into
`HassetteCLIClient(config, json_mode=..., transport=...)` alongside a `make_host_port_config()`.

The file already has a `make_transport(status_code, body, *, raise_exc)` helper (test_client.py:36),
but it `json.dumps()` whatever body it is given, so it cannot produce the bodies these tests need —
HTML, bare `not json` text, or deliberately malformed UTF-8 bytes. Every malformed-response test
therefore writes its own four-line handler instead. The bodies vary only in status code, byte
content, `content-type` header, and the client's `json_mode`/`debug_mode` flags.

**Goal:** extend the existing transport helper (or add a sibling — e.g. a `raw_content=` /
`content_type=` path on `make_transport`, and/or a small `make_raw_body_client(...)` that returns a
ready `HassetteCLIClient`) so a test can declare "this endpoint returns these bytes with this
content-type" in one line, and use it at every call site with this shape.

## Where

PMD flagged these three:

- `tests/unit/cli/test_client.py:264-270`
- `tests/unit/cli/test_client.py:275-281`
- `tests/unit/cli/test_client.py:287-293`

Reading the file turned up the same shape at (line numbers as of filing):

- `:236-239` — 503 with an HTML body (`TestToleratedServiceUnavailable`)
- `:339-344` and `:356-361` — 200 with malformed UTF-8 bytes, human mode and debug mode
- `:384-387` and `:395-398` — 200 with `not json` under a JSON content-type (`TestPostMalformedResponse`)

Any other instance of the same shape found while doing the work is in scope too — the list above is
supporting evidence, not a boundary.

### Closely related, same file

A second near-identical block repeats four times in the instance-resolution tests: a handler that
returns a serialized manifest list for `MANIFESTS_ENDPOINT` and `b"[]"` for anything else
(`:572-580`, `:603-610`, `:646-653`, `:681-688`), twice further wrapped in a `tracking_handler` that
appends `str(request.url)` to a `captured_urls` list (`:585-590`, `:655-660`) — which duplicates the
existing `url_capturing_client()` helper at `:79`. Worth collapsing in the same pass if the helper
shape ends up covering it; `conftest.py`'s route-builder (`:263-292`) may already be the right home.

## Acceptance Criteria

- [ ] A shared helper exists for building a `MockTransport`-backed CLI client that returns a fixed
      raw body with a caller-specified status code and `content-type`.
- [ ] Every test in `tests/unit/cli/test_client.py` matching that shape uses the helper instead of a
      local `def handler(...)`.
- [ ] The helper lives beside the existing `make_transport()` (or in `tests/unit/cli/conftest.py` if
      other CLI test modules need it), per the "check for an existing factory first" rule in
      `.claude/rules/test-conventions.md`.
- [ ] `uv run pytest tests/unit/cli/test_client.py` passes, and the tests still assert the same
      behavior (clean exit, no `Traceback` in stderr, correct JSON error envelope).

NodeJSmith:
## Summary

`tests/unit/cli/test_client.py` repeated the same four-line setup at eight call sites: an inline
`def handler(_req) -> httpx.Response` returning one fixed raw body, wrapped in
`httpx.MockTransport(...)`, passed into `HassetteCLIClient` alongside a `make_host_port_config()`.

The file's existing `make_transport()` helper couldn't serve these tests — it `json.dumps()` any
body it's handed, so it can't produce an HTML error page, a bare `not json` string, or the
deliberately malformed UTF-8 bytes the decode-error tests need. Every malformed-response test
therefore hand-rolled its own transport.

This adds `make_raw_body_client()` as a sibling to `make_transport()` and uses it everywhere that
shape appeared. A test now declares "this endpoint returns these bytes with this content-type" in
one line:

```python
client = make_raw_body_client(b'{"a": "\xff"}', content_type="application/json; charset=utf-8")
```

Omitting `content_type` leaves the header off entirely, matching a proxy that serves a raw error
page. `json_mode` and `debug_mode` are keyword-only passthroughs to the client.

## Call sites converted

All eight instances of the shape, across three test classes:

- `TestTolerate503` — 503 with an HTML body
- `TestMalformedSuccessResponse` — non-JSON 200 body (exit code, human-mode stderr, JSON envelope),
  plus the two malformed-UTF-8 cases in human and debug mode
- `TestPostMalformedResponse` — `not json` under a JSON content-type, both tests

Net effect on the file: **36 insertions, 52 deletions**. No test assertions changed — the same
behavior is still verified (clean exit code 1, `not valid JSON` / `not valid UTF-8` in stderr, no
`Traceback` leaking, correct JSON error envelope).

## Not in scope

The issue's "Closely related, same file" note flagged a second repeated block in the
instance-resolution tests (a manifest-list route handler, four occurrences). That's a genuinely
different shape — route-dependent responses rather than one fixed raw body — so this helper doesn't
cover it and it's left alone rather than forced into an ill-fitting abstraction. The four listed
acceptance criteria are all about the raw-body shape and are met.

## Testing

- `uv run nox -s dev` — **7737 passed, 1 skipped, 2 xfailed** (exit 0)
- `prek -a` — all 29 hooks pass, including the `check_test_factories.py` shadowing hook
  (`make_raw_body_client` doesn't collide with any name in `tests/support/`)

Test-only change; no runtime source touched, so no docs or frontend updates apply.

Closes #1982


NodeJSmith:
@coderabbitai review

NodeJSmith:
Automated autofix review pass.

**CodeRabbit:** no threads — the review was rate limited, so there was nothing to triage.

**Review battery** (code-reviewer, integration-reviewer, wtf-reviewer, llm-checker, lazy-checker, nitpicker): code-reviewer, llm-checker and lazy-checker came back clean. lazy-checker independently confirmed the "not in scope" claim in the PR body — the leftover manifest-list handler cluster is a genuinely different shape and is already tracked by #1984.

Three reviewers converged on one MEDIUM: `make_raw_body_client(content, status_code=200, ...)` inverted the positional order of its self-described sibling `make_transport(status_code=200, body=None, ...)`, inviting a swapped-argument call site. Fixed in e55d041a by making everything after `content` keyword-only, which removes the trap rather than trading one order for another. Remaining nitpicker findings are pre-existing file-wide patterns (the repeated `"content-type"` literal, the mixed `handler`/`_fixed_response` closure naming) and were left alone as out of scope for a test refactor.

**CI note:** the failing `security` check is not introduced by this PR. The diff touches no dependency files — `uv.lock` and `pyproject.toml` are byte-identical to `main`. `pip-audit` began reporting newly published advisories against the pinned `httpx2`/`httpcore2` 2.5.0 (CVE-2026-84378 through -84382, CVE-2026-84381) sometime between 20:58Z and 21:18Z today: #2015's `security` job passed at 20:58Z on this exact lockfile, and this PR's failed at 21:18Z. It needs a dependency bump on `main`, not a change here. `file-sizes` and `duplicate-code` are the usual `continue-on-error` baseline noise.

Auto-merge (squash) is enabled; it will merge once the dependency bump clears `security`.

PR #2044, opened 2026-09-08
test: trim comment bloat and harden the async settle in app.test.tsx

### LINKED ISSUE #1985: Clean up comment bloat and a fragile async settle in frontend/src/app.test.tsx
A quality scan of `frontend/src/app.test.tsx` (llm-checker, lazy-checker, nitpicker) surfaced four surviving findings. The file's structure, helper extraction, and factory reuse are all sound — every finding below is localized. Nothing here is a correctness bug in the tests.

## 1. Comment narrates a past fix's rationale (`app.test.tsx:566-570`)

The comment above `"includes a multi-instance app whose live overlay shows one running and one failed instance"` reads:

> Exercises `appLiveStatus()`'s multi-instance branch (app-data.ts), not just its fast-path passthrough of a manifest already tagged "degraded" — **the harder, real-world case this fix's rationale is built on**: a cached manifest can still read "running" while one of its instances has actually failed, until an execution event refetches the grid.

"this fix's rationale is built on" narrates a point-in-time change rather than stating a durable invariant, which the house self-contained-comments rule specifically calls out. A reader with no memory of the PR cannot tell which fix is meant.

**Fix:** keep the invariant, drop the narrative. e.g. `// Covers appLiveStatus()'s multi-instance branch: a cached manifest can read "running" while one instance has already failed, until an execution event refetches the grid.`

## 2. Verbose trailing comment restating the test name (`app.test.tsx:259-262`)

A 4-line comment sits after the final assertion in the visibilitychange smoke test:

> The handler should not throw — functional smoke test. / Tick increment is verified implicitly: the handler calls `state.tick.value++` which would throw if state were invalid. The `useRelativeTime` hook tests verify that tick increments cause re-renders with updated strings.

The first line restates the test's own name; the rest describes behavior this test does not assert, and points at a different file's coverage. It also trails the assertions rather than preceding anything.

**Fix:** collapse to one line above the `act()` block, e.g. `// Smoke test only — tick-increment re-render behavior is covered in useRelativeTime's own tests.`

## 3. Mock comment carries a history clause (`app.test.tsx:109-116`)

The 6-line block above the `alert-banner` mock ends with "…so the existing content-based `App — FailedAppsAlert` tests above keep working unmodified." That clause describes an edit that was made, not a property a future reader needs. The two substantive points (spy to confirm mount; wrap rather than replace so `AlertBanner` render counts are observable) are each one line.

**Fix:** trim to the two durable points, drop the "keep working unmodified" clause.

## 4. Raw `setTimeout(…, 0)` flush as a negative-assertion settle point (`app.test.tsx:497-499`)

```tsx
render(<App />);
await new Promise((resolve) => setTimeout(resolve, 0));
expect(callCount).toBe(0);
```

Two problems. It is the only bespoke async idiom in a file that otherwise uses `screen.findBy*` and, in the visibilitychange suite, `vi.useFakeTimers`. And as a settle point for a negative assertion it proves very little: a fetch scheduled after more than one macrotask (or behind a longer microtask chain) would still leave `callCount === 0` and the test would pass regardless.

**Fix:** settle on an actual rendered signal instead — await something that proves `App` has mounted and its initial data path has run (e.g. `await screen.findByTestId(...)` on an element the shell renders), then assert `callCount === 0`. That ties the assertion to observable app state rather than to one tick of the event loop.

## Filtered out during validity assessment

Four further findings were assessed and dropped as not holding up:

- **"Inline testids violate the file's centralization pattern"** — `SELECTORS` (lines 20-26) holds CSS attribute-selector strings consumed by `container.querySelector`, a different call path from `findByTestId`; `ALERT_BANNER_TESTID` is extracted because it appears in seven tests across two describe blocks, while `cmd-result-app-garage_app` and friends appear once or twice inside a single test each. Extraction here tracks reuse count, so there is no pattern being violated.
- **`listener_id: 42` duplicated across fixture, testid, and route** — all three literals sit within six lines of one test and are conventional explicit test data.
- **The wouter dynamic-import comment (lines 36-41) and the login-route hook comment (lines 525-528)** — both state durable hazards (import-order unsafety; 401 before a credential exists) that are not derivable from the adjacent code.
- **Extracting a `renderApp()` helper** — `userEvent.setup()` + `render(<App />)` is two lines of standard RTL boilerplate; wrapping it hides the render without removing a decision.

NodeJSmith:
## Summary

Cleans up `frontend/src/app.test.tsx` following a quality scan (#1985). Three comments are trimmed to their durable invariants, and one fragile async settle point is replaced with a real rendered signal. No test behavior beyond the settle-point change; the file's structure, helper extraction, and factory reuse were already sound.

## What changed

**Comment hygiene (findings 1-3)** — three comments narrated point-in-time changes or restated the test's own name rather than stating something a future reader needs:

- The `alert-banner` mock block dropped its "so the existing tests keep working unmodified" clause, keeping the two durable points: `TelemetryDegradedBanner` is spied to confirm `app.tsx` mounts it, and `AlertBanner` is *wrapped* rather than replaced so its render count stays observable.
- The visibilitychange smoke test's 4-line trailing comment collapsed to one line above the `act()` block. It had restated the test name and described assertions this test does not make.
- The multi-instance `FailedAppsAlert` comment dropped "the harder, real-world case this fix's rationale is built on" — a reader with no memory of that PR can't tell which fix is meant — while keeping the invariant it documents.

**Fragile async settle (finding 4)** — the command-palette "does not fetch when closed" test settled on `await new Promise((resolve) => setTimeout(resolve, 0))` before asserting `callCount === 0`. That proves very little for a negative assertion: a fetch scheduled behind more than one macrotask would leave `callCount === 0` and pass regardless. It was also the only bespoke async idiom in a file that otherwise uses `screen.findBy*`.

It now seeds a manifest and awaits `screen.findByTestId("app-entry-garage_app")` — the sidebar renders one entry per manifest, so the await ties the assertion to a mounted app whose initial data path actually ran, rather than to one tick of the event loop.

## Testing

- `npx vitest run src/app.test.tsx` — 42/42 passed, including the reworked palette test.
- `uv run nox -s dev` — 7737 passed, 1 skipped, 2 xfailed.
- `prek -a` — all hooks passed (ruff, prettier, eslint, stylelint, tsc, house-lint, and the frontend guards).

No visual evidence section: the only changed file is `*.test.tsx`, which `tools/frontend/check_pr_screenshots.py` excludes from the rendered-file trigger.

Closes #1985

### DIFF
--- a/frontend/src/app.test.tsx
+++ b/frontend/src/app.test.tsx
@@ -106,14 +106,10 @@ vi.mock("./hooks/use-query-params", () => ({
   useQueryParams: () => ({ get: () => null, set: vi.fn() }),
 }));
 
-// Spy on TelemetryDegradedBanner to verify it is mounted in the layout shell.
-// Component-level signal behaviour is fully tested in alert-banner.test.tsx;
-// here we only care that app.tsx renders the component at all.
-//
-// AlertBanner is also wrapped (not replaced) to count how many times FailedAppsAlert's child
-// actually re-renders — the render-count seam for the "App — FailedAppsAlert re-render scoping"
-// tests below. It still delegates to the real component so the existing content-based
-// "App — FailedAppsAlert" tests above keep working unmodified.
+// TelemetryDegradedBanner is replaced with a spy to confirm app.tsx mounts it; its own signal
+// behaviour is covered in alert-banner.test.tsx.
+// AlertBanner is wrapped rather than replaced so it still renders for real while exposing a render
+// count to the "App — FailedAppsAlert re-render scoping" tests.
 const alertBannerRenderCount = vi.hoisted(() => ({ renders: 0 }));
 
 vi.mock("./components/layout/alert-banner", async (importOriginal) => {
@@ -252,14 +248,10 @@ describe("App — visibilitychange tick recovery", () => {
 
     Object.defineProperty(document, "hidden", { value: false, writable: true, configurable: true });
 
+    // Smoke test only — tick-increment re-render behavior is covered in useRelativeTime's own tests.
     act(() => {
       handlers.forEach((handler) => handler(new Event("visibilitychange")));
     });
-
-    // The handler should not throw — functional smoke test.
-    // Tick increment is verified implicitly: the handler calls state.tick.value++
-    // which would throw if state were invalid. The useRelativeTime hook tests
-    // verify that tick increments cause re-renders with updated strings.
 
     addSpy.mockRestore();
   });
@@ -494,8 +486,11 @@ describe("App — command palette", () => {
         return HttpResponse.json<ListenerWithSummary[]>([]);
       }),
     );
+    seedManifests([createManifest({ app_key: "garage_app", display_name: "Garage App", status: "running" })]);
     render(<App />);
-    await new Promise((resolve) => setTimeout(resolve, 0));
+    // Settle on the manifests fetch resolving (the sidebar renders one entry per manifest), so the
+    // assertion reflects a mounted app whose initial data path ran rather than one event-loop tick.
+    await screen.findByTestId("app-entry-garage_app");
     expect(callCount).toBe(0);
   });
 });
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2053, opened 2026-09-08
test: extract shared breakpoint-drift seeding helper

### LINKED ISSUE #1988: Extract repeated breakpoint-drift test fixture seeding into a shared helper
## Description

`tests/unit/tools/test_check_breakpoint_drift.py` repeats the same "seed a consistent frontend
tree, then run `main()`" setup inline in every `test_main_*` test. The shared shape is: write the
JS breakpoint constant (`hooks/use-media-query.ts`), write the Tailwind `@theme` registration
(`global.css`), write a `@media (max-width: Npx)` rule into a stylesheet — all agreeing on the
same pixel value so the checker reports no drift — then `monkeypatch.setattr(sys, "argv", ...)`
before calling `main()`. Each test then perturbs exactly one layer to produce the behavior it is
actually asserting (an extra uncovered media query, an unregistered Tailwind screen, or one of
the three source layers omitted entirely to trigger an error path).

Surfaced by the `duplicate-code` CI job (PMD CPD). This is copy-paste setup, not designed
parallel structure — the file already pulls its tmp_path tree setup from a shared helper
(`make_frontend_src` in `tests/unit/tools/conftest.py`), so sharing this next layer follows the
convention already in place rather than inventing one.

## Goal

Extract the consistent-breakpoint seeding into a single shared helper (a fixture or a function
alongside `frontend_env`, taking per-layer overrides so a test can omit or replace one layer)
and use it at every call site with this shape. The perturbation each test applies — the extra
600px query, the `App.tsx` utility class, the deliberately missing layer — should stay visible at
the call site; only the identical baseline seeding moves into the helper.

## Where

PMD flagged these three as one block:

- `tests/unit/tools/test_check_breakpoint_drift.py:168-173` — `test_main_ok_prints_covered_breakpoints`
- `tests/unit/tools/test_check_breakpoint_drift.py:179-184` — `test_main_fails_when_css_breakpoint_has_no_js_constant`
- `tests/unit/tools/test_check_breakpoint_drift.py:190-195` — `test_main_fails_on_unregistered_tailwind_screen`

Other occurrences of the same shape found while reading, also in scope:

- `test_main_ok_when_all_breakpoints_covered` — the full three-layer seed
- `test_main_errors_when_no_theme_registrations` — JS constant layer only
- `test_main_errors_when_no_media_queries` — JS constant + `@theme` layers
- `test_main_errors_when_no_js_constants_file` / `test_main_smoke_test_flag_passes` — share only
  the `sys.argv` patch, which is worth folding into the same helper

Any other instance of this shape found while doing the work is in scope too, not just the
locations listed here — including sibling `tests/unit/tools/test_check_*.py` files if their
`frontend_env` fixtures grew the same seeding pattern.

## Acceptance Criteria

- [ ] A single shared helper (fixture or function) performs the consistent-breakpoint seeding of
      the JS constants file, `global.css` `@theme` registration, and `@media` stylesheet.
- [ ] The helper supports omitting or overriding an individual layer, so the error-path tests use
      it rather than keeping their own inline partial seeds.
- [ ] Every `test_main_*` test in `test_check_breakpoint_drift.py` uses the helper; no inline
      copy of the baseline seed remains.
- [ ] Each test's distinguishing perturbation is still visible at its own call site.
- [ ] `uv run pytest tests/unit/tools/test_check_breakpoint_drift.py` passes with the same set of
      tests as before (no test dropped or merged away).

NodeJSmith:
## Summary

`tests/unit/tools/test_check_breakpoint_drift.py` repeated the same three-layer "seed a consistent
frontend tree, then call `main()`" setup inline in every `test_main_*` test: a JS breakpoint
constant (`hooks/use-media-query.ts`), a Tailwind `@theme` registration (`global.css`), and a
`@media (max-width: Npx)` rule — all agreeing on 768px so the checker reports no drift — followed
by a `sys.argv` patch. PMD CPD (the `duplicate-code` CI job) flagged three of those tests as a
single duplicate block.

This extracts that baseline into a `seed_breakpoint_sources()` helper next to the existing
`frontend_env` fixture, following the convention the file already uses for its tmp-tree setup
(`make_frontend_src` in `tests/unit/tools/conftest.py`).

## What changed

- Added three module-level seed constants (`JS_CONSTANT_SEED`, `THEME_REGISTRATION_SEED`,
  `MEDIA_QUERY_SEED`) and a `seed_breakpoint_sources()` helper that writes all three layers and
  patches `sys.argv`.
- The helper takes per-layer overrides: passing `None` for `js`, `theme`, or `media` omits that
  layer, and passing a replacement string perturbs it. Omitting `src` entirely writes no files and
  only patches `sys.argv`, which covers the bare-tree and `--smoke-test` cases.
- Every `test_main_*` test now calls the helper. The error-path tests use layer omission
  (`theme=None, media=None`) instead of keeping their own partial inline seeds.
- Each test's distinguishing perturbation stays at its own call site and is commented: the extra
  uncovered 600px media query, the `App.tsx` utility class referencing an unregistered Tailwind
  screen, and the deliberately missing layer.

Net effect on the test file is +47/-23 lines — the helper plus its docstring costs more than the
inline seeds it replaces, but the duplicate block is gone and the intent of each test is now the
only thing at its call site.

## Scope note

The issue listed sibling `tests/unit/tools/test_check_*.py` files as conditionally in scope "if
their `frontend_env` fixtures grew the same seeding pattern." Only `test_check_dead_tokens.py` has
a similar-looking fixture, and its seeding is a different shape (CSS custom-property declarations
for the dead-token checker, not the three-layer breakpoint agreement), so it was left alone.

## Testing

- `uv run nox -s dev` — full unit + integration suite (excludes docker/e2e/system markers)
- `prek -a` — ruff, pyright, and the rest of the hook set

Test-only change: no source, schema, docs, or frontend files touched, so there is nothing to
regenerate and no visual evidence applies.

Closes #1988


NodeJSmith:
@coderabbitai review

PR #2061, opened 2026-09-08
test: extract shared instance-scoped action route setup in CLI app tests

### LINKED ISSUE #1992: Extract shared instance-scoped action route setup in CLI app command tests
## Description

The `duplicate-code` CI job (PMD CPD, `tools/check_duplicate_code.py`) flagged a block repeated 3+ times in `tests/unit/cli/test_commands_app.py`.

The shared shape is **constructing the two-route mock client for an instance-scoped app action**: a `GET /api/apps/manifests` route (used for instance-name resolution) paired with a `POST /api/apps/{key}/instances/{index}/{action}` route returning an `ActionResponse`, both handed to `cli_client_factory.build_with_routes([...])`. Every test in `TestCmdAppActionRouting` that exercises the `--instance` path rebuilds that same two-entry list inline, varying only the manifest instances, the requested index, and the server-confirmed `instance_index`.

The file already has `_action_response()`, `_instance()`, and `_manifest_route()` helpers — the missing piece is a helper one level up that assembles the whole route pair, so each test names only what it actually varies.

## Fix

Extract the shared instance-scoped route-pair construction into a single local helper (e.g. `_instance_action_routes(action, *, instances=..., requested_index=1, confirmed_index=1)`, returning the route list ready for `build_with_routes`) and use it at every call site with this shape. It needs to accommodate the existing variants: a manifest route that 503s, a `POST` that 404s, a confirmed index that deliberately disagrees with the requested one, and multi-instance manifests.

While in there, `TestCmdAppReload.test_prompts_for_confirmation_with_instance_name` hand-inlines the bodies of `_instance()` and `_manifest_route()` instead of calling them — it should go through the same helpers.

If the assembled shape turns out to be useful beyond this file, `tests/unit/cli/conftest.py` is the natural home; otherwise keeping it file-local next to the existing `_manifest_route` helper is fine.

## Where

PMD flagged these three as the sampled cluster:

- `tests/unit/cli/test_commands_app.py:443-453` — `test_json_output_includes_instance_index_app_level` / `test_success_message_instance_level` region
- `tests/unit/cli/test_commands_app.py:473-483`
- `tests/unit/cli/test_commands_app.py:515-525`

Reading the file, the same shape also appears at:

- `tests/unit/cli/test_commands_app.py:407-419` — `test_with_instance_resolves_and_hits_instance_route`
- `tests/unit/cli/test_commands_app.py:488-500` — `test_success_message_falls_back_to_raw_selector_when_instance_unresolvable`
- `tests/unit/cli/test_commands_app.py:531-543` — `test_warns_on_server_instance_index_mismatch`
- `tests/unit/cli/test_commands_app.py:551-563` — `test_no_warning_when_server_instance_index_matches`
- `tests/unit/cli/test_commands_app.py:654-662` — `test_error_on_instance_out_of_range` (404 variant)
- `tests/unit/cli/test_commands_app.py:668-687` — `TestCmdAppReload.test_prompts_for_confirmation_with_instance_name` (fully inlined variant)

These line numbers are supporting evidence, not a boundary. **Any other instance of the same shape found while doing the work is in scope too**, whether or not it is listed here.

## Acceptance Criteria

- [ ] A single helper builds the instance-scoped manifest + action route pair, covering the 503-manifest, 404-action, index-mismatch, and multi-instance variants.
- [ ] Every call site in `tests/unit/cli/test_commands_app.py` with this shape uses it, including the inlined one in `TestCmdAppReload`.
- [ ] Each test's remaining setup names only the values it actually varies.
- [ ] `uv run pytest tests/unit/cli/test_commands_app.py` passes; `prek -a` is clean.

NodeJSmith:
## Summary

Test-only refactor of `tests/unit/cli/test_commands_app.py`. Every test exercising an
instance-scoped app action (`start` / `stop` / `reload` with `--instance`) rebuilt the same
two-entry route list inline: a `GET /api/apps/manifests` route for instance-name resolution
paired with a `POST /api/apps/{key}/instances/{index}/{action}` route. The nine copies varied
only in the manifest instances, the requested index, and the index the server echoes back —
everything else was duplicated boilerplate.

## What changed

Adds `_instance_action_routes()` alongside the existing `_action_response`, `_instance`, and
`_manifest_route` helpers. It builds the route pair from the parameters that actually differ
between tests and covers every variant present in the file:

- default single-instance success case
- explicit multi-instance manifests, and an index that resolves to no manifest entry
- `manifest_unavailable=True` — the 503 telemetry-outage path
- `confirmed_index` — the server-confirmed-index-mismatch warning path
- `action_status` / `action_detail` — the failed-action path (404 instance-not-found)

All call sites now route through it. `TestCmdAppReload.test_prompts_for_confirmation_with_instance_name`
previously hand-inlined the bodies of `_instance` and `_manifest_route` rather than calling
them; it now uses the same helpers as everything else.

Net effect is 95 lines removed, 59 added, with no change to what any test asserts.

## Testing

- `prek -a` — all hooks pass

No production code is touched, so there are no docs or frontend changes to make.

Closes #1992


NodeJSmith:
@coderabbitai review

PR #2068, opened 2026-09-09
test: extract shared JSON log-record helpers for logging unit tests

### LINKED ISSUE #1995: Extract shared JSON log-record parsing helper in logging unit tests
The `duplicate-code` CI job (PMD CPD) flagged a repeated block in the logging unit tests: the
five-line ritual for **pulling the last JSON log record out of a `StringIO` stream and asserting on
one of its fields**. Every copy does the same four steps — read `stream.getvalue()`, split into
non-blank lines, assert at least one line was emitted, then `json.loads()` the last one — before the
one line that actually differs (which key is checked, and against what).

## Goal

Extract the shared "parse the emitted JSON log records from a stream" step into a small test helper
(e.g. `last_json_record(stream) -> dict` in `tests/support/helpers.py`, or in the logging tests'
conftest alongside the `logging_pipeline` fixture), and use it at every call site with this shape so
each test is left with just its own assertion. Per `.claude/rules/test-conventions.md`, prefer adding
it to the shared `tests/support` surface over a file-local `make_*` if more than one test module ends
up using it.

There is a near-variant worth folding into the same helper family rather than leaving as a
one-off: `test_logging_correlation.py` selects the record line by substring match instead of taking
the last one. A sibling helper (e.g. `json_record_containing(stream, text) -> dict`) covers that
shape and would let both files stop hand-rolling the split/parse.

## Where

PMD flagged these three:

- `tests/unit/test_logging_setup.py:79-83`
- `tests/unit/test_logging_setup.py:130-134`
- `tests/unit/test_logging_setup.py:269-273`

Reading the file turned up two more copies of the identical block that PMD did not list:

- `tests/unit/test_logging_setup.py:93-95`
- `tests/unit/test_logging_setup.py:107-109`

And the substring-match variant:

- `tests/unit/test_logging_correlation.py:339-343`

The flagged line ranges are a sample, not a boundary. **Any other instance of the same shape found
while doing the work is in scope too**, not just the locations listed above.

## Acceptance Criteria

- [ ] A shared helper exists for extracting a parsed JSON log record from a test log stream.
- [ ] All copies of the split/assert/`json.loads` block in `tests/unit/test_logging_setup.py` use it.
- [ ] The substring-match variant in `tests/unit/test_logging_correlation.py` uses the helper (or a
      sibling helper in the same family) rather than its own inline parsing.
- [ ] Each test's remaining body is its own distinct assertion, not shared plumbing.
- [ ] `uv run pytest tests/unit/test_logging_setup.py tests/unit/test_logging_correlation.py` passes.

NodeJSmith:
## Summary

Two logging unit-test files each hand-rolled the same "read the stream, split into
lines, drop blanks, assert something was emitted, `json.loads` the one we want"
block before they could assert on a single field. Six copies across
`test_logging_setup.py` and one substring-matching variant in
`test_logging_correlation.py`.

This replaces all of them with two shared helpers in `tests/support/helpers.py`:

- `last_json_record(stream)` — parses the most recent JSON log record written to a
  test log stream.
- `json_record_containing(stream, text)` — parses the first record whose raw line
  contains `text`, for tests that need a specific record rather than the last one.

Both raise a descriptive `AssertionError` when no matching record was emitted, so a
test that logs nothing fails with "no log records were emitted to the stream"
instead of an `IndexError` or a confusing `json.loads` error.

Most call sites collapse to a single line — `assert last_json_record(stream)["level"]
== "warning"` — which puts the actual assertion, rather than the parsing boilerplate,
at the front of each test.

## Why

Per `.claude/rules/test-conventions.md`, a shape needed in 3+ files belongs in the
shared factory/helper modules rather than accumulating local copies. This one had
reached six. The new helpers are registered in that rule's canonical-helpers list so
they're discoverable before someone writes a seventh copy.

## Testing

- `prek -a` — all hooks pass

Test-only change; no `src/` behavior is touched.

Closes #1995


NodeJSmith:
@coderabbitai review

PR #2070, opened 2026-09-09
test: deduplicate endpoint literal and fixture shape in config-tab tests

### LINKED ISSUE #1998: Clean up repeated endpoint literal and duplicated fixtures in config-tab.test.tsx
Surfaced by an automated quality scan of `frontend/src/components/app-detail/config-tab.test.tsx`. Both findings are pre-existing, mechanical, and local to this one file. Two of the three checkers (llm-checker, lazy-checker) returned no findings — comments and mock boundaries in this file are clean.

### 1. The config endpoint path is repeated inline 6 times

`"/api/apps/:app_key/config"` is written out as a literal at `config-tab.test.tsx:83`, `:122`, `:137`, `:147`, `:165`, and `:183` — once in `beforeEach` and once in each per-test `server.use()` override. If the route ever changes, all six call sites have to move together, and a missed one silently falls through to the `beforeEach` default handler rather than failing loudly.

There is existing precedent for the fix in this codebase: `components/shared/log-table/use-log-data.test.ts:14` declares `const LOGS_ENDPOINT = "/api/logs/recent";` and reuses it across all six of its `http.get` registrations.

**Proposed fix:** declare `const CONFIG_ENDPOINT = "/api/apps/:app_key/config";` alongside the other top-of-file constants (`APP_KEY`, `MASK_SENTINEL`, `HOST`, `PORT`) and use it at all six call sites.

### 2. `defaultConfig` and `noSchemaConfig` duplicate their shared shape verbatim

`defaultConfig` (`config-tab.test.tsx:31-55`) and `noSchemaConfig` (`:58-68`) are independently authored objects that repeat the same five fields identically — `app_key`, `filename: "test_app.py"`, `class_name: "TestApp"`, `enabled: true`, `autostart: true`. Only `app_config`, `config_toml`, and the presence of `config_schema` actually differ, which is the whole point of the second fixture. The `"test_app.py"` and `"TestApp"` literals in particular are duplicated even though the sibling value `APP_KEY` directly above them was already extracted into a constant.

The two fixtures can silently drift on the shared fields, and the duplication obscures which fields the "no schema" test case is actually varying.

**Proposed fix:** extract the shared five fields into a `baseAppMeta` object and compose both fixtures from it (`{ ...baseAppMeta, app_config: ..., config_toml: ..., config_schema: ... }`). A shared factory in `test/factories.ts` is **not** warranted here — `config_toml` appears in only this one test file, well under the bar for promoting a fixture to the shared registry.

## Findings assessed and dropped

Recorded so a later scan does not re-file them:

- **Centralize the `data-testid` string literals into constants.** The stated rationale was that it removes drift risk, but a constant local to the test file cannot prevent drift with the component — `config-tab.tsx` declares those testid strings independently, so a rename there breaks the test whether or not the test side is centralized. The claimed benefit does not hold.
- **`HOST = "192.168.1.1"` is a hard-coded environment value.** It is already a named constant at `:24`, is inert fixture data asserted against rendered text, and is never connected to. Nothing about it is environment-specific.
- Two further findings ("scattered constants" for the endpoint path, "naming inconsistency" for the fixtures) restated items 1 and 2 above and were merged into them.

## Acceptance Criteria

- [ ] `"/api/apps/:app_key/config"` appears exactly once in the file, as a named constant, and all six `http.get` registrations reference it
- [ ] `"test_app.py"` and `"TestApp"` each appear exactly once in the file
- [ ] `defaultConfig` and `noSchemaConfig` are composed from a shared base object rather than repeating the five common fields
- [ ] `cd frontend && npx vitest run src/components/app-detail/config-tab.test.tsx` still passes with all 13 tests green

## Autonomous fix suitability

Suitable. Both fixes have a single unambiguous shape, touch one file, involve no behavior or design decision, and are verified by the existing tests continuing to pass.

NodeJSmith:
## Summary

Removes two mechanical duplications in `frontend/src/components/app-detail/config-tab.test.tsx`. Test-only change — no production code or behavior is touched.

## What changed

**Endpoint literal extracted to a constant.** `"/api/apps/:app_key/config"` was written inline at six `http.get` registrations (once in `beforeEach`, once per test override). A missed call site during a route rename would have silently fallen through to the `beforeEach` default handler instead of failing loudly. It is now a single `CONFIG_ENDPOINT` constant alongside the other top-of-file constants, matching the existing precedent in `components/shared/log-table/use-log-data.test.ts`.

**Shared fixture fields composed from one base object.** `defaultConfig` and `noSchemaConfig` independently repeated the same five fields (`app_key`, `filename`, `class_name`, `enabled`, `autostart`), leaving them free to drift and obscuring which fields the "no schema" case actually varies. Both now spread a `baseAppMeta` object, so only the genuinely differing fields (`app_config`, `config_toml`, `config_schema`) appear per fixture.

A shared factory in `test/factories.ts` was deliberately not used — `config_toml` appears in this file only, below the bar for promoting a fixture to the shared registry.

## Testing

- `npx vitest run src/components/app-detail/config-tab.test.tsx` — 13 passed

Literal counts verified against the issue's acceptance criteria: `/api/apps/:app_key/config`, `test_app.py`, and `TestApp` each now appear exactly once in the file.

Closes #1998

### DIFF
--- a/frontend/src/components/app-detail/config-tab.test.tsx
+++ b/frontend/src/components/app-detail/config-tab.test.tsx
@@ -2,9 +2,12 @@ import { render, screen, waitFor } from "@testing-library/react";
 import { delay, http, HttpResponse } from "msw";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
+import type { components } from "../../api/generated-types";
 import { server } from "../../test/server";
 import { getShikiHighlighter } from "../../utils/shiki";
 import { ConfigTab } from "./config-tab";
+
+type AppConfigResponse = components["schemas"]["AppConfigResponse"];
 
 /** Mocked at the getShikiHighlighter boundary (not the "shiki" package) so each test controls
  *  resolution/rejection directly — the real module caches per-language, which would make a
@@ -28,8 +31,9 @@ const PORT = 8080;
 /** Long enough to keep the request in flight while the component unmounts mid-request. */
 const MOCK_RESPONSE_DELAY_MS = 100;
 
-/** Fields shared by every app config response fixture below. */
-const baseAppMeta = {
+/** Fields shared by every app config response fixture below. Typed against the generated
+ *  response model so a backend field rename fails the type check instead of silently drifting. */
+const baseAppConfigFields: Pick<AppConfigResponse, "app_key" | "filename" | "class_name" | "enabled" | "autostart"> = {
   app_key: APP_KEY,
   filename: "test_app.py",
   class_name: "TestApp",
@@ -39,7 +43,7 @@ const baseAppMeta = {
 
 /** App config response with a schema that marks 'token' as a secret via anyOf. */
 const defaultConfig = {
-  ...baseAppMeta,
+  ...baseAppConfigFields,
   app_config: {
     token: MASK_SENTINEL,
     host: HOST,
@@ -62,7 +66,7 @@ const defaultConfig = {
 
 /** App config response without a schema — falls back to SimpleConfigTable. */
 const noSchemaConfig = {
-  ...baseAppMeta,
+  ...baseAppConfigFields,
   app_config: {
     api_key: "some-value",
   },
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2081, opened 2026-09-09
test: render via renderSection helper and derive mocked ExecutionTable prop type

### LINKED ISSUE #2010: Clean up ExecutionSection test helper usage and mock prop typing
Two hygiene findings surfaced by a quality scan of `frontend/src/components/app-detail/execution-section.test.tsx`. Both are pre-existing, mechanical, and local to this file.

## Key Items

- **`execution-section.test.tsx:72-85` — the fourth test bypasses the `renderSection` helper it should use.** `renderSection` (line 31) exists specifically to avoid repeating the `<ExecutionSection .../>` JSX and accepts `Partial<ComponentProps<typeof ExecutionSection>>` as overrides. The other three tests use it; `"passes the correct props to ExecutionTable"` hand-writes a full `render(<ExecutionSection ... />)` call instead. Every prop it sets (`heading`, `records`, `kind`, `tableId`, `loading`, `appKey`, `handlerKind`, `handlerId`, `instanceQs`) is declared on `ExecutionSectionProps` (`execution-section.tsx:5-14`), so all of them fit through `overrides`.

  Fix: `renderSection({ records: [SUCCESS_EXECUTION], kind: "job", tableId: "table-42", appKey: "my_app", handlerKind: "listener", handlerId: 7, instanceQs: "?instance=0" })`. The assertion block at lines 89-100 is unchanged.

- **`execution-section.test.tsx:12` — the mock hand-declares a shadow of `ExecutionTable`'s prop shape.** The `vi.mock` factory types its stub as `(props: { records: ExecutionRecord[]; tableId: string })`, a manually-maintained subset that can silently drift if `ExecutionTableProps` (`shared/execution-table.tsx:152`) changes. The same file already derives types the right way for `ExecutionSection` at line 31 via `ComponentProps`.

  Fix: derive it instead — e.g. `Pick<ComponentProps<typeof import("../shared/execution-table").ExecutionTable>, "records" | "tableId">`. A type-position `import(...)` is erased at compile time, so it does not conflict with `vi.mock`'s hoisting the way a value reference would.

## Acceptance Criteria

- [ ] The `"passes the correct props to ExecutionTable"` test renders via `renderSection` with overrides rather than its own `render(<ExecutionSection ... />)` call, with assertions unchanged.
- [ ] The mocked `ExecutionTable` stub's prop type is derived from the real component's props rather than hand-declared.
- [ ] `cd frontend && npm run test -- execution-section` passes and type checking is clean.

## Not filed

The scan also flagged the inline literals in that test (`handlerId={7}`, `tableId="table-42"`, `instanceQs="?instance=0"`) as magic values needing named constants. Rejected: each is asserted verbatim 15 lines below at `execution-section.test.tsx:91-97`, so the setup/assertion pairing is already visible at a glance — hoisting them to constants would put indirection between the two halves of the test rather than removing it.

NodeJSmith:
## Summary

Two mechanical hygiene fixes in `frontend/src/components/app-detail/execution-section.test.tsx`. Test-only change — no production code or behavior is touched.

## What changed

**1. `"passes the correct props to ExecutionTable"` now renders via the `renderSection` helper.**

The file already defines `renderSection(overrides)` to avoid repeating the `<ExecutionSection .../>` JSX, and three of the four tests use it. This one hand-wrote its own full `render(<ExecutionSection ... />)` call. Every prop it set is declared on `ExecutionSectionProps`, so all of them pass cleanly through the helper's `Partial<ComponentProps<typeof ExecutionSection>>` overrides. The assertion block below it is unchanged, so the test still proves exactly what it did before.

**2. The mocked `ExecutionTable` stub derives its prop type instead of hand-declaring one.**

The `vi.mock` factory typed its stub as a literal `{ records: ExecutionRecord[]; tableId: string }` — a manually-maintained subset of the real `ExecutionTableProps` that would silently drift if the real component's props changed. It now derives the shape:

```ts
Pick<ComponentProps<typeof import("../shared/execution-table").ExecutionTable>, "records" | "tableId">
```

A type-position `import(...)` is erased at compile time, so it does not collide with `vi.mock`'s hoisting the way a value reference would.

## Why

Both findings are pre-existing and local to this one file. The second is the one with actual teeth: a hand-written prop shadow gives no signal when the real component's props change, so the mock can quietly stop resembling what it stands in for. Deriving it makes `tsc` the enforcement mechanism rather than a reader noticing.

The issue also flagged the inline literals (`handlerId={7}`, `tableId="table-42"`, `instanceQs="?instance=0"`) as magic values. Deliberately left alone — each is asserted verbatim about fifteen lines below, so hoisting them to named constants would insert indirection between the two halves of the test instead of removing it.

## Testing

- `npx vitest run src/components/app-detail/execution-section.test.tsx` — 5 passed
- `npx vitest run` (full frontend suite) — 110 files, 1604 tests passed
- `prek -a` — all hooks pass, including `tsc (frontend)` and `eslint (frontend)`

No screenshots: the diff touches only a `.test.tsx` file, which `tools/frontend/check_pr_screenshots.py` excludes from the visual-evidence requirement.

Closes #2010

### DIFF
--- a/frontend/src/components/app-detail/execution-section.test.tsx
+++ b/frontend/src/components/app-detail/execution-section.test.tsx
@@ -9,11 +9,15 @@ vi.mock("../shared/execution-table", async () => {
   const actual = await vi.importActual<typeof import("../shared/execution-table")>("../shared/execution-table");
   return {
     ...actual,
-    ExecutionTable: vi.fn((props: { records: ExecutionRecord[]; tableId: string }) => (
-      <div data-testid="mock-execution-table" data-record-count={props.records.length} data-table-id={props.tableId}>
-        execution table
-      </div>
-    )),
+    ExecutionTable: vi.fn(
+      (
+        props: Pick<ComponentProps<typeof import("../shared/execution-table").ExecutionTable>, "records" | "tableId">,
+      ) => (
+        <div data-testid="mock-execution-table" data-record-count={props.records.length} data-table-id={props.tableId}>
+          execution table
+        </div>
+      ),
+    ),
   };
 });
 
@@ -70,19 +74,15 @@ describe("ExecutionSection", () => {
   });
 
   it("passes the correct props to ExecutionTable", () => {
-    render(
-      <ExecutionSection
-        heading="Executions"
-        records={[SUCCESS_EXECUTION]}
-        kind="job"
-        tableId="table-42"
-        loading={false}
-        appKey="my_app"
-        handlerKind="listener"
-        handlerId={7}
-        instanceQs="?instance=0"
-      />,
-    );
+    renderSection({
+      records: [SUCCESS_EXECUTION],
+      kind: "job",
+      tableId: "table-42",
+      appKey: "my_app",
+      handlerKind: "listener",
+      handlerId: 7,
+      instanceQs: "?instance=0",
+    });
 
     // React calls function components with only a props argument (no Preact-style second
     // "context" argument), so the second call arg is undefined here.


NodeJSmith:
@coderabbitai review

PR #2086, opened 2026-09-09
refactor(frontend): share the idle-status helper and roving-tabindex type across handler components

### LINKED ISSUE #2013: Extract the shared idle-status derivation and roving-tabindex type used by handler-health-card
## Description

Automated clean-code scan of `handler-health-card.tsx`. Two checkers (llm-checker, lazy-checker) returned clean — the file has no comment bloat, no dead code, and no deferred-debt markers. The nitpicker raised 11 items; 6 were dropped after checking them against the codebase (see "Checked and dismissed" below). The 2 that survive are both single-source-of-truth duplications where the file restates a concept that already has, or clearly wants, one canonical home. Both are mechanical, no behavior change.

### 1. The "mute status means idle" derivation is duplicated across two files under two names

`handler-health-card.tsx:73` computes `const idle = item.statusKind === "mute";` and `unified-handler-row.tsx:68` computes the byte-identical `const isIdle = item.statusKind === "mute";`. Both then drive the same `opacity-60` idle treatment.

This is the one derivation of its kind that *wasn't* extracted. `overview-tab-helpers.ts:12` already exports the exact sibling shape:

```ts
export function isFailing(item: UnifiedItem): boolean {
  return item.statusKind === "err";
}
```

and `handler-health-card.tsx` already imports and uses it (line 60, `const failing = isFailing(item)`). So the precedent for the fix lives one function away in a module this file already depends on.

**Fix:** add `isIdle(item: UnifiedItem)` to `overview-tab-helpers.ts` next to `isFailing`, and have both `handler-health-card.tsx` and `unified-handler-row.tsx` import it instead of restating the comparison. Pick one of the two local names while doing it so the concept reads the same in both files.

Note: this is *not* covered by #1907 (the `unified-handler-row.tsx` scan). That issue mentions `isIdle` only in passing as a boolean-naming reference point and never flags the cross-file duplication or the missing helper.

### 2. The roving-tabindex type `0 | -1` is written out inline in three places

`handler-health-card.tsx:30` (`HandlerHealthCardProps.tabIndex`), `handler-health-card.tsx:40` (`StatTooltipProps.tabIndex`), and `hooks/use-roving-tab-index.ts:54` (the `getTabIndex` return type) each independently spell out `0 | -1`. The hook is the producer of the value and the natural owner of the type, but it exports no name for it, so every consumer re-derives the convention by hand.

**Fix:** export `type RovingTabIndex = 0 | -1;` from `hooks/use-roving-tab-index.ts`, use it as `getTabIndex`'s return type, and import it for both prop declarations in this file.

## Checked and dismissed

Recorded so a future scan doesn't re-raise them:

- **`../../` relative imports should use the `@/` alias** — the project's own lint rule draws the line elsewhere. `eslint.config.js` `no-restricted-imports` uses the pattern `^\.\.\/\.\.\/\.\.\/.+` with the message "Use the @/ path alias instead of deep relative imports (3+ levels)." This file's imports are 2 levels, explicitly permitted. Repo-wide, 74 files use `../../` vs 8 using `@/hooks`/`@/utils`.
- **Inline `color-mix(... 40% ...)` percentage is a magic number** — inline percentages in `color-mix()` are the established convention: `handlers-rows.tsx:88`, `unified-handler-row.tsx:115`, `components/ui/button.tsx:22,24,27`, `log-detail-drawer.tsx:252,271`, `log-table-view.tsx:90` all do the same.
- **`opacity-60` / `opacity-100` are untokenized magic values** — the sibling `unified-handler-row.tsx:80` uses the identical bare utility, and `unified-handler-row.test.tsx:184,190` assert on the literal `"opacity-60"` string. (Adding a token is a reasonable idea, but it's a design-system decision spanning both files, already noted in #1907, not a defect in this one.)
- **Comparing `statusKind` against the bare string `"mute"` could silently desync on a rename** — `StatusKind` is a closed union (`utils/status.ts:175`), so TypeScript rejects a comparison against a literal outside it. A rename fails the type check rather than passing silently. The duplication angle survives as finding 1; the desync rationale does not.
- **`instanceQs` abbreviates "query string" inconsistently** — it's the established project-wide name, used in `overview-tab-helpers.ts:6`'s own signature, `pages/app-detail.tsx:257`, and five test files.
- **`HandlerHealthCard` is ~90 lines; extract a `useHandlerHealthCardData(item)` hook** — the derived block (lines 58-73) is flat straight-line assignment with no branching or state. Extracting it adds an indirection layer without removing branches or hidden state, which is the trade the repo's reader-load guidance argues against.

## Acceptance Criteria

- [ ] `isIdle(item: UnifiedItem)` exists in `overview-tab-helpers.ts` alongside `isFailing`
- [ ] `handler-health-card.tsx` and `unified-handler-row.tsx` both import it instead of inlining `item.statusKind === "mute"`, using the same local name in both
- [ ] `RovingTabIndex` (or equivalent) is exported from `hooks/use-roving-tab-index.ts` and used as `getTabIndex`'s return type
- [ ] Both `tabIndex` prop declarations in `handler-health-card.tsx` reference that type instead of restating `0 | -1`
- [ ] `npm run build`, `npx tsc --noEmit`, and the existing `handler-health-card.test.tsx` / `unified-handler-row.test.tsx` suites pass

NodeJSmith:
## Summary

Two single-source-of-truth duplications in the app-detail handler components, both mechanical with no behavior change.

**1. The "mute status means idle" derivation lived in two files under two names.** `handler-health-card.tsx` computed `const idle = item.statusKind === "mute"`, and `unified-handler-row.tsx` computed the byte-identical `const isIdle = item.statusKind === "mute"`. Both drove the same `opacity-60` idle treatment. `overview-tab-helpers.ts` already exported the exact sibling shape (`isFailing`), which `handler-health-card.tsx` already imported — so the precedent for the fix sat one function away in a module the file already depended on.

This adds `isIdle(item: UnifiedItem)` next to `isFailing` and has both components import it, settling on `idle` as the local name in both so the concept reads the same either place.

**2. The roving-tabindex value type `0 | -1` was spelled out at three sites** — both `tabIndex` prop declarations in `handler-health-card.tsx` and `getTabIndex`'s return type in `use-roving-tab-index.ts`. The hook produces the value and is the natural owner of the type, but exported no name for it, so every consumer re-derived the convention by hand. This exports `RovingTabIndex` from the hook and uses it at all three sites.

Also whitelists `docs/pr-evidence/*.png` in `.gitignore` (the repo ignores `*.png` by default) and excludes `pr-evidence/**` from the mkdocs build so the evidence images below don't land on the docs site.

## Testing

- `npm test -- --run` (frontend) — 1604 passed across 110 files, including the existing `handler-health-card.test.tsx` and `unified-handler-row.test.tsx` suites
- `prek -a` — all hooks pass, including `tsc (frontend)`, `eslint (frontend)`, and `prettier (frontend)`

## Screenshots

This is a refactor with no intended visual change; the captures below are evidence that the idle and failing treatments still render identically after routing both through the shared helper.

App detail overview — handler health cards:

App detail overview

App handlers list — unified handler rows:

App handlers list

Closes #2013


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2089, opened 2026-09-09
test: consolidate handler-health-grid test blocks and share section testids

### LINKED ISSUE #2016: Clean up test hygiene in handler-health-grid.test.tsx
Quality-scanner pass over `frontend/src/components/app-detail/handler-health-grid.test.tsx`. Three checkers ran (llm-checker, lazy-checker, nitpicker); llm-checker and lazy-checker returned CLEAN. The surviving findings below are all test-hygiene, none affect production code.

## 1. Misleading `describe` block with an inert prop override

`describe("HandlerHealthGrid — passes props to cards")` (line 91) contains exactly one test, `"renders correct number of cards for given items"`, which asserts only `expect(cards).toHaveLength(3)`. It passes `{ instanceQs: "?instance=1" }` as an override (line 98) but never asserts that `instanceQs` reaches the cards — the override changes nothing observable in the test.

The block name promises prop-passing coverage the test does not provide, so a future reader trusts a guarantee that isn't there. Actual `instanceQs` propagation through the grid is already covered in `overview-tab.test.tsx:335` (`"card navigation includes instanceQs"`, asserting the navigate URL), so this is a naming/dead-weight problem, not a coverage hole.

**Proposed fix:** fold the card-count test into the existing `describe("HandlerHealthGrid — with items")` block and drop the inert `instanceQs: "?instance=1"` override, leaving the default props. Do not add a duplicate propagation assertion here — `overview-tab.test.tsx` already owns that.

## 2. Testid literals hardcoded instead of living in the shared helper module

`overview-health-empty` and `overview-health-grid` are hand-typed string literals at lines 42, 47, and 67, and again at `overview-tab.test.tsx:221`, `:262`, and `:270`. The sibling module `handler-health.test-helpers.ts` exists for exactly this purpose but only covers the card testids (`CARD_TESTID_PREFIX`, `CARD_SELECTOR`, `cardTestId`), so callers hand-type the section-level ones. A rename of either testid in `handler-health-grid.tsx` would need six separate edits with no compiler help.

**Proposed fix:** add `HEALTH_GRID_TESTID = "overview-health-grid"` and `HEALTH_EMPTY_TESTID = "overview-health-empty"` to `handler-health.test-helpers.ts` and use them at all six call sites across both test files.

## 3. Repeated inline type expression

`Partial<typeof DEFAULT_GRID_PROPS>` is written out verbatim in both `renderGrid` (line 21) and `renderGridWithAppState` (line 25).

**Proposed fix:** extract a local `type GridPropOverrides = Partial<typeof DEFAULT_GRID_PROPS>` and use it in both signatures. Minor; bundled here rather than filed separately.

## Assessed and dropped

The nitpicker also flagged the two-level relative imports (`../../test/factories`, `../../test/mock-wouter`, `../../test/render-helpers`, lines 4-6) as import-hygiene violations that should use the configured `@/` alias. Dropped as likely invalid: the `@/` alias exists, but `frontend/src` currently has 28 files importing test helpers via the `../../test/` relative form against 8 using `@/test/`, so the relative form is the dominant existing convention. Singling out this one file would increase inconsistency rather than reduce it. Converging on one style is a separate, repo-wide decision.

NodeJSmith:
## Summary

Test-hygiene cleanup in `handler-health-grid.test.tsx`, addressing the three findings from the quality-scanner pass in #2016. No production code changes — this touches only test files and a test-helper module.

## What changed

**1. Removed a misleading `describe` block with an inert prop override**

`describe("HandlerHealthGrid — passes props to cards")` held a single test that asserted only `expect(cards).toHaveLength(3)`. It passed `{ instanceQs: "?instance=1" }` as an override but never asserted that `instanceQs` reached the cards, so the block name promised prop-passing coverage the test did not provide.

The card-count test is now folded into the existing `describe("HandlerHealthGrid — with items")` block with the inert override dropped. No duplicate propagation assertion was added — `overview-tab.test.tsx` ("card navigation includes instanceQs") already owns that coverage by asserting the navigate URL.

**2. Moved section-level testid literals into the shared helper module**

`overview-health-grid` and `overview-health-empty` were hand-typed string literals at six call sites across two test files, while the sibling `handler-health.test-helpers.ts` already existed for this purpose but only covered the card testids. Renaming either testid in `handler-health-grid.tsx` would have required six separate edits with no compiler help.

Both are now exported as `HEALTH_GRID_TESTID` and `HEALTH_EMPTY_TESTID` and used at all six call sites.

**3. Extracted a repeated inline type expression**

`Partial<typeof DEFAULT_GRID_PROPS>` was written out verbatim in both `renderGrid` and `renderGridWithAppState`; it is now a local `type GridPropOverrides`.

## Not done, deliberately

The scanner also flagged the two-level relative imports (`../../test/factories` and friends) as import-hygiene violations that should use the `@/` alias. Left alone, as the issue itself recommends: `frontend/src` currently has 28 files using the relative form against 8 using `@/test/`, so the relative form is the dominant convention here and converging on one style is a separate repo-wide decision.

## Testing

- `npx vitest run` on the two changed files — 47 passed
- `npx vitest run` (full frontend suite) — 110 files, 1604 tests passed
- `prek -a` — all hooks passed (includes prettier, eslint, stylelint, tsc)

No screenshots: the diff touches only `*.test.tsx` and a `.ts` helper, both excluded from the rendering-file set that `check_pr_screenshots.py` guards.

Closes #2016

### DIFF
--- a/frontend/src/components/app-detail/handler-health-grid.test.tsx
+++ b/frontend/src/components/app-detail/handler-health-grid.test.tsx
@@ -4,7 +4,7 @@ import { describe, expect, it, vi } from "vitest";
 import { createJob, createListener } from "../../test/factories";
 import { createWouterMock } from "../../test/mock-wouter";
 import { renderWithAppState } from "../../test/render-helpers";
-import { CARD_SELECTOR, cardTestId } from "./handler-health.test-helpers";
+import { CARD_SELECTOR, cardTestId, HEALTH_EMPTY_TESTID, HEALTH_GRID_TESTID } from "./handler-health.test-helpers";
 import { HandlerHealthGrid } from "./handler-health-grid";
 import { buildItems } from "./handler-list";
 import type { UnifiedItem } from "./unified-handler-row";
@@ -18,11 +18,13 @@ vi.mock("wouter", () =>
 
 const DEFAULT_GRID_PROPS = { appKey: "test_app", instanceQs: "" };
 
-function renderGrid(items: UnifiedItem[], overrides: Partial<typeof DEFAULT_GRID_PROPS> = {}) {
+type GridPropOverrides = Partial<typeof DEFAULT_GRID_PROPS>;
+
+function renderGrid(items: UnifiedItem[], overrides: GridPropOverrides = {}) {
   return render(<HandlerHealthGrid items={items} {...DEFAULT_GRID_PROPS} {...overrides} />);
 }
 
-function renderGridWithAppState(items: UnifiedItem[], overrides: Partial<typeof DEFAULT_GRID_PROPS> = {}) {
+function renderGridWithAppState(items: UnifiedItem[], overrides: GridPropOverrides = {}) {
   return renderWithAppState(<HandlerHealthGrid items={items} {...DEFAULT_GRID_PROPS} {...overrides} />);
 }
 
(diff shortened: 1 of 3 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2092, opened 2026-09-09
test: collapse render boilerplate and name fixtures in handler-list tests

### LINKED ISSUE #2021: Clean up repeated render boilerplate and opaque ID assertions in handler-list.test.tsx
Surfaced by an automated quality scan of `frontend/src/components/app-detail/handler-list.test.tsx`. Both findings are pre-existing, test-only, and mechanical. Two of the three checkers (llm-checker, lazy-checker) returned CLEAN — no comment bloat, no deferred debt, correct reuse of shared factories and `unified-row.test-helpers`.

## Finding 1 — Nine near-identical `render(<HandlerList ... />)` call sites

Every one of the nine tests hand-rolls the same four-prop render call (lines 31, 37-39, 45, 51-53, 60, 68-70, 77, 83-85, 100-102), varying only one or two props. As part of that, `onSelect={() => {}}` is re-declared as a fresh inline no-op nine separate times.

The repetition also forces awkward line wrapping — several call sites wrap onto three lines purely to stay under the 120-char limit, which is noise rather than signal about what each test varies.

**Proposed fix:** add a local `renderHandlerList(overrides = {})` helper at the top of the file defaulting `listeners: []`, `jobs: []`, `selectedId: null`, `onSelect: noop` (with a single module-scope `const noop = () => {};`). Each test then reads as a one-line override stating only the prop under test. The sibling `handlers-tab.test-helpers.ts` is precedent for this shape in the same directory.

## Finding 2 — Ordering assertion uses bare numeric IDs while the fixtures carry semantic names

In `renders issues first while preserving source order within each health group` (lines 90-111), the fixtures are named (`healthy_listener`, `failing_listener`, `healthy_job`, `failing_job`, `idle_job`) but the expected-order array (lines 104-110) refers to them only by number: `rowTestId("listener", 2)`, `rowTestId("job", 6)`, and so on. A reader has to map five numbers back to five names by eye to check that the assertion actually encodes "issues first, source order preserved within group" — which is the entire point of the test.

**Proposed fix:** bind the IDs to named constants in the test (e.g. `const FAILING_LISTENER = 2;`) or, better, drive the expected array off the fixture objects so the assertion reads in terms of health group rather than integers.

## Filtered out during validity assessment

- **"listener"/"job" string literals should be constants** — `rowTestId(kind: UnifiedItemKind, id: number)` already constrains these at compile time via the `UnifiedItemKind` union; a typo fails typecheck. Constants would add indirection with no safety gain.
- **`onSelect: () => void` in the mock's props type is a dead field** — it is not dead. The annotation types the props object `HandlerList` actually passes to the mocked row; dropping it would make the mock's contract inaccurate. Not destructuring an unused prop is normal.
- **`../../test/factories` climbs two directory levels instead of using the `@/` alias** — the `@/` alias exists, but the relative form is used by ~18 test files across `src/` versus ~8 using `@/test/factories`. This file matches the more common existing pattern; normalizing is a repo-wide consistency question, not a defect here.

NodeJSmith:
## Summary

Test-only cleanup of `frontend/src/components/app-detail/handler-list.test.tsx`, addressing both findings from the automated quality scan in #2021. No production code or component behavior changes.

## What changed

**1. Collapsed nine near-identical render call sites into a local helper.**

Every one of the nine tests hand-rolled the same four-prop `HandlerList` render call, re-declaring an inline `onSelect={() => {}}` no-op each time. Several wrapped onto three lines purely to stay under the 120-char limit, so the wrapping communicated line-length pressure rather than what the test actually varied.

A local `renderHandlerList(overrides)` helper now supplies the empty/unselected defaults (`listeners: []`, `jobs: []`, `selectedId: null`, `onSelect: noop`), with a single module-scope `noop`. Each test now states only the prop under test, and most fit on one line. `Partial<ComponentProps<typeof HandlerList>>` types the overrides, so the helper stays in sync with the component's props automatically. This matches the sibling helper shape already used in this directory.

**2. Named the fixtures in the ordering assertion.**

The `renders issues first while preserving source order within each health group` test built five semantically-named fixtures (`healthy_listener`, `failing_listener`, `healthy_job`, `failing_job`, `idle_job`) but asserted the expected order using bare numeric IDs — `rowTestId("listener", 2)`, `rowTestId("job", 6)`, and so on. Verifying that the assertion actually encodes "issues first, source order preserved within group" meant mapping five integers back to five names by eye, which is the whole point of the test.

Each fixture is now bound to a named variable and the expected array derives its IDs from those objects, so the assertion reads in terms of health group. This is the second, preferred option the issue proposed.

## Testing

- `prek -a` — all hooks pass (ruff, eslint, prettier, stylelint, tsc, house-lint, and the rest)
- `npx vitest run src/components/app-detail/handler-list.test.tsx` — 9 passed

The nine tests pass unchanged in name and count; only their construction changed. No screenshots apply — `check_pr_screenshots.py` excludes `*.test.tsx`, and this PR touches no rendered frontend file.

Closes #2021

### DIFF
--- a/frontend/src/components/app-detail/handler-list.test.tsx
+++ b/frontend/src/components/app-detail/handler-list.test.tsx
@@ -1,4 +1,5 @@
 import { render } from "@testing-library/react";
+import type { ComponentProps } from "react";
 import { describe, expect, it, vi } from "vitest";
 
 import { createJob, createListener } from "../../test/factories";
@@ -7,6 +8,8 @@ import type { UnifiedItemKind } from "./unified-handler-row";
 import { ROW_TESTID_PREFIX, rowTestId } from "./unified-row.test-helpers";
 
 const HANDLER_LIST_TEST_ID = "handler-list";
+
+const noop = () => {};
 
 // Mock UnifiedHandlerRow to isolate HandlerList behavior — the row component
 // calls query hooks which require the Zustand app store (useAppStore) and MSW.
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #2094, opened 2026-09-09
test: extract shared two-owner Router arrange block in test_router.py

### LINKED ISSUE #2025: Extract the shared two-owner Router arrange block in test_router.py
## Description

The `duplicate-code` CI job (PMD CPD via `tools/check_duplicate_code.py`) flags a repeated arrange block in `tests/unit/bus/test_router.py`: build a fresh `Router()`, create two listeners on the same topic owned by `"owner1"` and `"owner2"`, and `add_route` both. Only the assertion tail differs between copies — each test then exercises a different `Router` method (`get_topic_listeners`, `remove_listener_by_id`, `clear_owner`, `get_listeners_by_owner`) against the same two-owner fixture.

This file is a known gap rather than a deliberate exception. `design/specs/097-dedupe-bus-test-scaffolding/design.md` names `test_router.py` explicitly as one of the files that dedupe pass never targeted, and `tests/unit/bus/CLAUDE.md` already documents the established shape for this fix: a module-scope helper in the test file itself, since the pattern repeats only within this one file (`.claude/rules/test-conventions.md` reserves `tests/support/` placement for patterns reused across 3+ files).

## Goal

Extract the shared `Router` arrange block into a module-local helper (or a small pytest fixture) in `test_router.py`, returning the router and its listeners, and use it at every call site with that shape. Then document the helper in `tests/unit/bus/CLAUDE.md`'s "File-local helpers" section, matching how `test_invocation.py`, `test_handler_invoker.py`, `test_duration_hold.py`, and `test_duration_timer.py` are already listed there.

Near-identical variants in the same file are in scope for the same helper family:

- Two listeners, same topic, different owners (the flagged cluster).
- Two listeners, same owner, different topics — `test_clear_owner_removes_all_owner_listeners` and `test_get_listeners_by_owner_returns_added_listeners`.
- The single-listener setup (`router = Router()` / `create_listener(topic=...)` / `router.add_route(...)`), which repeats roughly ten more times across the file.

Whether the single-listener case is worth its own helper is a judgment call for whoever picks this up — collapse it if it reads better, leave it if the explicit three lines are clearer than a call. Follow the precedent set in `test_duration_hold.py`: prefer two narrow helpers over one flag-branching helper, and leave a test's setup explicit when the setup itself is part of what the test proves.

## Where

- `tests/unit/bus/test_router.py:49-54` — `test_add_route_multiple_listeners_same_topic`
- `tests/unit/bus/test_router.py:135-140` — `test_remove_listener_by_id_leaves_other_listeners`
- `tests/unit/bus/test_router.py:187-192` — `test_clear_owner_leaves_other_owners`
- `tests/unit/bus/test_router.py:251-256` — `test_get_listeners_by_owner_does_not_return_other_owners`

These line ranges are supporting evidence, not a boundary. Any other instance of the same shape found while doing the work is in scope too, including the variants listed above and anything the reader spots that these ranges missed.

## Acceptance Criteria

- [ ] A module-local helper (or fixture) in `test_router.py` builds the two-owner `Router` fixture, and all four flagged tests use it.
- [ ] The same-owner/different-topic variant is either covered by a sibling helper or explicitly left alone with the reason clear from the code.
- [ ] The helper is listed in `tests/unit/bus/CLAUDE.md` under "File-local helpers".
- [ ] `uv run pytest tests/unit/bus/test_router.py` passes with no change in what each test asserts.
- [ ] The PMD cluster for these locations no longer appears in `tools/check_duplicate_code.py` output.

NodeJSmith:
## Summary

`tools/check_duplicate_code.py` (PMD CPD) flagged a repeated arrange block in `tests/unit/bus/test_router.py`: build a fresh `Router()`, create two listeners on the same topic owned by `"owner1"` and `"owner2"`, then `add_route` both. Four tests shared that setup verbatim and differed only in which `Router` method they exercised afterward.

This extracts the block into two module-scope helpers and updates every call site with that shape.

## What changed

**`tests/unit/bus/test_router.py`** — two module-local helpers:

- `router_with_two_owners(topic="state_changed")` — two listeners on one topic, owned by `"owner1"` and `"owner2"`. Used by the four flagged tests (`test_add_route_multiple_listeners_same_topic`, `test_remove_listener_by_id_leaves_other_listeners`, `test_clear_owner_leaves_other_owners`, `test_get_listeners_by_owner_does_not_return_other_owners`).
- `router_with_two_topics(topic1="state_changed", topic2="call_service", owner_id="owner1")` — the near-identical variant with one owner across two topics. Used by `test_clear_owner_removes_all_owner_listeners` and `test_get_listeners_by_owner_returns_added_listeners`.

Both return `(router, l1, l2)`.

**`tests/unit/bus/CLAUDE.md`** — both helpers documented under "File-local helpers", matching the existing entries for `test_invocation.py`, `test_handler_invoker.py`, `test_duration_hold.py`, and `test_duration_timer.py`.

## Design choices

Two narrow helpers instead of one flag-branching helper, following the precedent set in `test_duration_hold.py` (`arm_duration_timer` / `arm_remaining_duration_timer`).

The single-listener setup (`Router()` / `create_listener` / `add_route`, repeated ~10 more times) is deliberately left inline. The issue flagged this as a judgment call. The topic and listener kwargs vary per test — glob patterns, priorities, owner index — and those varying arguments are part of what each test proves, so the explicit three lines read better than a call with three overrides. The reasoning is recorded in the `CLAUDE.md` entry so the next reader does not re-litigate it.

No assertion was changed; this is a pure arrange-block extraction.

## Testing

- `prek -a` — all hooks pass (frontend hooks required `cd frontend && npm install` first, per `.claude/rules/frontend-worktree.md`)
- `uv run python tools/check_duplicate_code.py` — `test_router.py` no longer appears anywhere in the output. CPD still scans the file's directory (9 other `tests/unit/bus/*.py` files appear in the remaining known-failing baseline), so the absence reflects the fix rather than the file being excluded from analysis.

Closes #2025


NodeJSmith:
@coderabbitai review

PR #2099, opened 2026-09-09
test: centralize handlers-tab app-key/URL constants and clean up comment noise

### LINKED ISSUE #2033: Clean up comment noise and repeated URL literal in handlers-tab.navigation.test.tsx
## Description

Quality scan of `frontend/src/components/app-detail/handlers-tab.navigation.test.tsx` surfaced four pre-existing hygiene findings. All predate the current work and are mechanical, single-file cleanups. None require human judgment — each has one obvious fix.

### 1. Orphaned task-tracker comment (`:41`)

`// URL-driven selection tests (T03)` references a task ID from an archived orchestration run. Task files are ephemeral and get removed by `cfl archive`, so `T03` is meaningless to any future reader. The remaining content ("URL-driven selection tests") is already carried by the `describe` block and the test titles directly below it.

**Fix:** delete the comment.

Three sibling files carry the same orphaned pattern (`(T07)`) and could be cleaned in the same pass:
- `frontend/src/hooks/use-query-params.test.ts:10`
- `frontend/src/hooks/use-correct-url.test.ts:6`
- `frontend/src/components/shared/log-table/use-log-filters.test.ts:14`

### 2. Comments restating the adjacent code (`:72`, `:74`)

```ts
it("does not call correctUrl when data is empty (loading guard)", () => {
  // Empty arrays = loading state / no data — should not correct URL
  renderHandlersTab([], [], "listener/999");
  // The empty-state branch renders, no correctUrl call
  expect(mockCorrectUrl).not.toHaveBeenCalled();
});
```

Line 72 restates the `it(...)` title one line above it; line 74 restates the assertion it precedes. Neither adds information.

**Fix:** delete both comments.

### 3. Repeated URL literal (`:30`, `:62`, `:68`, `:83`, `:91`, `:103`, `:109`, `:147`)

`"/apps/test_app/handlers"` (and its `"test_app"` component) is hand-typed at eight sites — once inside the `vi.mock("wouter", ...)` factory that establishes the location, then re-typed at every assertion. If the mocked location ever changes, each assertion has to be updated independently, and a missed one fails in a way that reads as a routing bug rather than a stale literal.

**Fix:** extract module-level constants near the top of the file, e.g.

```ts
const APP_KEY = "test_app";
const HANDLERS_URL = `/apps/${APP_KEY}/handlers`;
```

and derive the mock location, the `appKey` props, and the assertion strings from them.

### 4. Mock callback name does not match the prop it stands in for (`:124`, `:142`)

The local spy is `onSwitch`, but it is passed as `onSwitchToCode` (`:136`) and the test title asserts "calls onSwitchToCode..." (`:122`). Reading the assertion at line 142 requires cross-referencing the prop declaration to confirm which callback fired.

**Fix:** rename the local to `onSwitchToCode`.

## Assessed and excluded

Three findings from the scan were checked against the codebase and dropped:

- **`../../test/...` relative imports should use the `@/` alias** — relative imports are the dominant convention here: 28 files import from `../../test/`, 29 from `../test/`, only 8 from `@/test/`. Converting this one file would move it away from the majority pattern, not toward it.
- **`uptimeSeconds: 120` is an unnamed magic number** — it appears 14 times across the frontend test suite as an established store-override idiom. A per-file constant would not fix the repo-wide pattern and would make this file inconsistent with its siblings.
- **`instanceIndex: 1` and `"?instance=1"` are coupled magic numbers** — that coupling is the assertion. The test exists to prove the `instanceIndex` prop produces the matching query string; deriving one from the other would make the assertion self-referential and stop it from catching a regression.

## Acceptance Criteria

- [ ] `// URL-driven selection tests (T03)` removed from `handlers-tab.navigation.test.tsx:41` (optionally the three `(T07)` siblings too)
- [ ] The two restating comments at lines 72 and 74 removed
- [ ] `"test_app"` / `"/apps/test_app/handlers"` extracted to module-level constants and used at all eight sites
- [ ] Local spy renamed from `onSwitch` to `onSwitchToCode`
- [ ] `cd frontend && npx vitest run src/components/app-detail/handlers-tab.navigation.test.tsx` passes

NodeJSmith:
## Summary

Mechanical hygiene cleanup of `frontend/src/components/app-detail/handlers-tab.navigation.test.tsx`. Test-only change — no source or behavior is touched.

Four findings from the quality scan in #2033:

1. **Removed the orphaned task-tracker comment** — `// URL-driven selection tests (T03)` referenced a task ID from an orchestration run that `cfl archive` has since removed, so `T03` no longer resolves to anything. The remaining content was already carried by the `describe` block and the test titles below it.
2. **Removed two comments that restated adjacent code** — one repeated the `it(...)` title one line above it, the other repeated the assertion it preceded.
3. **Extracted the repeated URL literal** — `"test_app"` and `"/apps/test_app/handlers"` were hand-typed at eight sites: once in the `vi.mock("wouter", ...)` factory that establishes the mocked location, then re-typed at every assertion and `appKey` prop. They now derive from two module-level constants:
   ```ts
   const APP_KEY = "test_app";
   const HANDLERS_URL = `/apps/${APP_KEY}/handlers`;
   ```
   Previously, changing the mocked location meant updating each assertion independently, and a missed one failed in a way that read as a routing bug rather than a stale literal.
4. **Renamed the mock callback to match its prop** — the local spy was `onSwitch` but was passed as `onSwitchToCode`, so reading the assertion required cross-referencing the prop declaration to confirm which callback fired.

The three sibling files carrying the same orphaned `(T07)` comment are listed as optional in the issue and are left alone here to keep this diff to a single file.

## Testing

- `npx vitest run src/components/app-detail/handlers-tab.navigation.test.tsx` — 12 passed / 12 (the acceptance-criteria check)
- `uv run nox -s dev` — 7737 passed, 1 skipped, 2 xfailed, 0 failed

## Note on draft status

This is opened as a draft only because the automated shipping run found the work uncommitted rather than already committed, so it routed to the draft path. Nothing is failing and every acceptance criterion in #2033 is met — it should be ready to mark for review as-is.

Partial progress on #2033 — needs human review

Closes #2033

---

## Review follow-up (2026-09-09)

All six reviewers (code, integration, WTF, llm-checker, lazy-checker, nitpicker) converged on the
same finding: `APP_KEY`/`HANDLERS_URL` were declared inside the navigation test while
`renderHandlersTab` in `handlers-tab.test-helpers.ts` kept its own hardcoded `appKey: "test_app"`.
The two agreed only by convention, so the extraction named the literal without actually removing
the duplication — most tests received their app key from the helper, not from the constant, and a
future divergence would have surfaced as a failing URL assertion rather than a type error.

Fixed at the root: both constants are now exported from `handlers-tab.test-helpers.ts`, the helper
derives its `appKey` from `APP_KEY`, and the whole `handlers-tab.*.test.tsx` family (navigation,
job, listener, rendering) derives its wouter mock locations, `appKey` props, and store-event
`app_key` fixtures from them. `"test_app"` is now written exactly once in the directory.

Not changed: the two removed "loading guard" comments. Two reviewers noted the removal arguably
drops the *why* (empty arrays mean loading, not an empty state), but the deletion is explicitly
what #2033 asked for and the test title carries `(loading guard)`.

Verification: `npx vitest run src/components/app-detail/` — 291 passed / 18 files; `npx tsc
--noEmit` clean; `prek -a` all hooks passed.

### DIFF
--- a/frontend/src/components/app-detail/handlers-tab.job.test.tsx
+++ b/frontend/src/components/app-detail/handlers-tab.job.test.tsx
@@ -8,7 +8,7 @@ import { useAppStore } from "../../state/store";
 import { createJob } from "../../test/factories";
 import { createWouterMock } from "../../test/mock-wouter";
 import { server } from "../../test/server";
-import { renderHandlersTab } from "./handlers-tab.test-helpers";
+import { APP_KEY, HANDLERS_URL, renderHandlersTab } from "./handlers-tab.test-helpers";
 
 vi.mock("sonner", async (importOriginal) => {
   const actual = await importOriginal<typeof import("sonner")>();
@@ -41,7 +41,7 @@ const SCHEDULE_STATUSES = ["scheduled", "waiting", "completed", "manual"] as con
 /** Past RUN_NOW_FEEDBACK_TIMEOUT_MS (8000ms in job-detail.tsx) so the timeout fallback fires. */
 const PAST_RUN_NOW_FEEDBACK_TIMEOUT_MS = 10000;
 
-vi.mock("wouter", () => createWouterMock({ useLocation: () => ["/apps/test_app/handlers", mockNavigate] }));
+vi.mock("wouter", () => createWouterMock({ useLocation: () => [HANDLERS_URL, mockNavigate] }));
 
 vi.mock("../../hooks/use-correct-url", () => ({
   useCorrectUrl: () => mockCorrectUrl,
@@ -373,7 +373,7 @@ describe("HandlersTab job detail", () => {
               {
                 kind: "job",
                 job_id: jobId,
-                app_key: "test_app",
+                app_key: APP_KEY,
                 instance_index: 0,
                 status: "success",
                 duration_ms: 10,
@@ -427,7 +427,7 @@ describe("HandlersTab job detail", () => {
             {
               kind: "job",
               job_id: 999,
-              app_key: "test_app",
+              app_key: APP_KEY,
               instance_index: 0,
               status: "success",
               duration_ms: 10,
@@ -477,7 +477,7 @@ describe("HandlersTab job detail", () => {
             {
               kind: "job",
               job_id: jobId,
-              app_key: "test_app",
+              app_key: APP_KEY,
               instance_index: 0,
               status: "success",
               duration_ms: 10,
--- a/frontend/src/components/app-detail/handlers-tab.listener.test.tsx
+++ b/frontend/src/components/app-detail/handlers-tab.listener.test.tsx
@@ -4,7 +4,7 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import { createListener } from "../../test/factories";
 import { createWouterMock } from "../../test/mock-wouter";
-import { renderHandlersTab } from "./handlers-tab.test-helpers";
+import { HANDLERS_URL, renderHandlersTab } from "./handlers-tab.test-helpers";
 
 // Mock child components that make API calls
 vi.mock("../shared/execution-table", () => ({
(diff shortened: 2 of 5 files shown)


NodeJSmith:
@coderabbitai review

PR #2101, opened 2026-09-09
test: derive mobile-breakpoint trigger width from BREAKPOINT_MOBILE

### LINKED ISSUE #2035: Import BREAKPOINT_MOBILE in handlers-tab rendering test instead of duplicating 768 in a comment
## Description

A quality scan of `frontend/src/components/app-detail/handlers-tab.rendering.test.tsx` surfaced one finding that survived validity assessment. It predates the current work.

## Finding: mobile-breakpoint value is duplicated as a comment-only literal instead of importing `BREAKPOINT_MOBILE`

`handlers-tab.rendering.test.tsx:129`:

```tsx
observer?.trigger(400); // narrower than BREAKPOINT_MOBILE (768)
```

The component under test (`handlers-tab.tsx:99`) compares `entry.contentRect.width < BREAKPOINT_MOBILE`, and `BREAKPOINT_MOBILE` is a real exported constant (`src/hooks/use-media-query.ts:7`) already imported directly by `handlers-tab.tsx`. This test file does not mock that module, so it can import the constant too — but instead it hardcodes `400` and restates the constant's value (`768`) in a comment.

That makes the comment a second source of truth for a value that lives in code. If `BREAKPOINT_MOBILE` is ever lowered below `400`, the comment becomes wrong and the test silently stops exercising the mobile branch — the assertion on `back-to-list` would simply fail with no indication that the magic number, not the component, was the problem.

**Proposed fix:** import the constant and derive the trigger width from it, dropping the comment as redundant:

```tsx
import { BREAKPOINT_MOBILE } from "../../hooks/use-media-query";
// ...
observer?.trigger(BREAKPOINT_MOBILE - 1);
```

## Acceptance Criteria

- [ ] `handlers-tab.rendering.test.tsx` imports `BREAKPOINT_MOBILE` from `../../hooks/use-media-query` rather than hardcoding a width and restating `768` in a comment
- [ ] The `observer?.trigger(...)` call derives its width from `BREAKPOINT_MOBILE`
- [ ] The now-redundant `// narrower than BREAKPOINT_MOBILE (768)` comment is removed
- [ ] `npx vitest run src/components/app-detail/handlers-tab.rendering.test.tsx` passes

## Findings assessed and excluded

For the record, these were raised by the scanners but dismissed with evidence:

- **`uptimeSeconds: 120` is a magic number** — 120 is an established repo-wide test convention, set in the shared `renderHandlersTab` helper and used identically in `overview-tab.test.tsx`, `handlers-tab.navigation.test.tsx`, and `recent-activity-section.rerender.test.tsx`. Not a file-local drift risk.
- **`"test_app"` and testid literals repeated** — test IDs are intentionally literal in tests so the assertion mirrors the DOM; 2–4 occurrences in one file is not indirection-worthy.
- **The last test reconstructs `QueryClientProvider` manually instead of using `renderHandlersTab`** — `renderWithAppState` (`src/test/render-helpers.tsx`) wraps the tree inline rather than via RTL's `wrapper` option, so `rerender()` would drop the provider. The manual path is required for this test's rerender, not a shortcut.
- **`cb` constructor param abbreviation** — a private callback field on a 25-line test double; no consequence.
- **`// Mock child components that make API calls`** — states why the mocks exist, not just what the code does.
- **Issue `#1785` in the test name** — a legitimate traceability pointer, not history narration.

NodeJSmith:
## Summary

The `HandlersTab` mobile-layout test triggered its resize observer with a hardcoded `400`
and restated the real breakpoint value in a trailing comment:

```tsx
observer?.trigger(400); // narrower than BREAKPOINT_MOBILE (768)
```

The component under test (`handlers-tab.tsx`) compares `entry.contentRect.width < BREAKPOINT_MOBILE`
against the exported constant in `src/hooks/use-media-query.ts`. This test file does not mock that
module, so it can simply import the constant — the comment was acting as a second source of truth
for a value that already lives in code.

That duplication is a silent-failure risk rather than a cosmetic one. If `BREAKPOINT_MOBILE` were
ever lowered below `400`, the comment would become wrong and the test would stop exercising the
mobile branch entirely. The `back-to-list` assertion would fail with nothing pointing at the magic
number as the cause.

## What changed

- Import `BREAKPOINT_MOBILE` from `../../hooks/use-media-query` in
  `frontend/src/components/app-detail/handlers-tab.rendering.test.tsx`
- Derive the trigger width as `BREAKPOINT_MOBILE - 1` instead of hardcoding `400`
- Drop the now-redundant `// narrower than BREAKPOINT_MOBILE (768)` comment

Test-only change — no production code or rendered output is touched, so this is typed `test:`
and stays out of the changelog.

## Testing

- `npx vitest run src/components/app-detail/handlers-tab.rendering.test.tsx` — 5 passed
- Full frontend suite: `npx vitest run` — 110 files, 1604 tests passed
- `prek -a` — all hooks pass (ruff, eslint, tsc, stylelint, breakpoint-drift guard, house lints)

The breakpoint-drift guard (`tools/frontend/check_breakpoint_drift.py`) still passes, confirming
the JS constant this test now imports remains in sync with the Tailwind screen registration in
`global.css`.

Closes #2035

### DIFF
--- a/frontend/src/components/app-detail/handlers-tab.rendering.test.tsx
+++ b/frontend/src/components/app-detail/handlers-tab.rendering.test.tsx
@@ -2,6 +2,7 @@ import { QueryClientProvider } from "@tanstack/react-query";
 import { act, render, waitFor } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
+import { BREAKPOINT_MOBILE } from "../../hooks/use-media-query";
 import { useAppStore } from "../../state/store";
 import { createListener } from "../../test/factories";
 import { createWouterMock } from "../../test/mock-wouter";
@@ -126,7 +127,7 @@ describe("HandlersTab rendering", () => {
     expect(observer?.element).toBeDefined();
 
     act(() => {
-      observer?.trigger(400); // narrower than BREAKPOINT_MOBILE (768)
+      observer?.trigger(BREAKPOINT_MOBILE - 1);
     });
 
     await waitFor(() => expect(getByTestId("back-to-list")).toBeDefined());


NodeJSmith:
@coderabbitai review

NodeJSmith:
Automated autofix review (`/review-autofixes`, scoped to this PR).

**CodeRabbit:** 0 unresolved threads — its review was rate-limited (`Review rate limited`), so nothing was posted to triage.

**Review battery (6 reviewers on `git diff origin/main...HEAD`):**

| Reviewer | Verdict |
|---|---|
| integration-reviewer | PASS — 0 findings on the diff; 1 LOW pre-existing, out of scope |
| wtf-reviewer | 1 LOW (skipped, see below) |
| llm-checker | 1 finding (skipped — premise disproven, see below) |
| lazy-checker | CLEAN |
| nitpicker | CLEAN |

**No code changes were needed.** The diff matches all four acceptance criteria in #2035 exactly, and `npx vitest run src/components/app-detail/handlers-tab.rendering.test.tsx` passes locally (5/5).

**Findings skipped, with reasons:**

- *llm-checker (test/implementation coupling):* argued that `BREAKPOINT_MOBILE - 1` is tautologically `< BREAKPOINT_MOBILE`, so the test can no longer catch a wrong change to the constant — whereas the old hardcoded `400` would have. The premise is wrong: `frontend/src/hooks/use-media-query.test.ts:7` directly asserts `expect(BREAKPOINT_MOBILE).toBe(768)`, and `tools/frontend/check_breakpoint_drift.py` (prek-enforced, verified passing) pins the constant against the CSS/Tailwind registrations. Those are the dedicated guards. The old coupling was incidental, and it would have turned a *legitimate* breakpoint change into a spurious failure in an unrelated ResizeObserver regression test.
- *wtf-reviewer (LOW, restore an explanatory comment):* declined — AC #3 of #2035 explicitly calls for removing that comment, and `BREAKPOINT_MOBILE - 1` already reads as "one pixel under the mobile breakpoint."

**Pre-existing issue filed:** #2105 — four `log-table` test files still hardcode `768`/`1024` inside `vi.mock("@/hooks/use-media-query", ...)` factories (one with a `// keep in sync with ...` comment). Verified unchanged vs `origin/main`, so out of scope here; it needs `vi.importActual`/`vi.hoisted()` rather than a one-line import.

**CI:** green except `duplicate-code` and `file-sizes`, which are the known `continue-on-error` baseline failures on `main` per CLAUDE.md.

Auto-merge (squash) enabled.

PR #2104, opened 2026-09-09
test: extract shared recording-listener setup in listener dispatch tests

### LINKED ISSUE #2037: Extract the repeated recording-listener setup in listener dispatch tests
## Description

`tests/integration/test_listeners_dispatch.py` repeats the same "recording listener" setup in nearly every test: an empty `calls` list, a tiny handler that appends `event.data` to it, and a `create_listener(handler, task_bucket=bucket, owner_id="test", topic="t", ...)` call that differs only in the rate-limiting kwarg (`debounce=`, `throttle=`, `once=`, or nothing). PMD's `duplicate-code` job flags this as one duplicated block across 7 locations in the file.

A second, smaller shape rides along with it: the `for i in range(3): async def invoke_fn(val=str(i + 1)): ... ; await listener.invoker.dispatch(invoke_fn)` loop, copied verbatim in three `TestListenerDispatchAndCancel` tests.

Neither shape is a deliberate exception — this family of tests already has a shared helper module (`tests/integration/listener_helpers.py`, "Shared helpers for listener integration tests"), which is the natural home for the extraction, and the repo already has a test-factory convention (`tests/support/factories.py` / `tests/support/helpers.py`, enforced by `tools/check_test_factories.py`).

## Goal

Extract the shared recording-listener setup into one helper (e.g. a `make_recording_listener(bucket, **listener_kwargs) -> (listener, calls)` in `tests/integration/listener_helpers.py`, or a fixture if that reads better) and use it at every call site with this shape, so each test only expresses what is actually distinct about it — the rate-limiting kwarg and the assertion. Do the same for the repeated three-shot `dispatch()` loop.

Verify handler identity is not asserted anywhere before extracting: `create_listener` derives `identity.handler_name` / `short_name` from the handler object, so moving the handler into a helper changes those values. Nothing in `tests/integration/test_listeners_*.py` asserts on them today, but confirm before and after.

## Where

PMD flagged these locations as one block:

- `tests/integration/test_listeners_dispatch.py:17-22`
- `tests/integration/test_listeners_dispatch.py:65-74`
- `tests/integration/test_listeners_dispatch.py:138-143`
- `tests/integration/test_listeners_dispatch.py:174-179`
- `tests/integration/test_listeners_dispatch.py:189-194`
- `tests/integration/test_listeners_dispatch.py:207-212`
- `tests/integration/test_listeners_dispatch.py:225-230`

Other copies of the same shape found while reading, also in scope:

- `tests/integration/test_listeners_dispatch.py:26-35`, `37-46`, `48-57`, `97-113`, `241-249`, `256-260`, `262-270` — the same `calls` + handler + `create_listener(...)` preamble, including the `lambda _e: None` variants in the cancel tests
- `tests/integration/test_listeners_dispatch.py:196-201`, `214-219`, `232-237` — the repeated `for i in range(3)` dispatch loop
- `tests/integration/test_listeners_validation.py:19-89`, `96-121` — the same `create_listener(handler, task_bucket=bucket, owner_id="test", topic="t")` preamble with a recording list; the handler bodies differ (dependency-injection annotations), so judge case by case whether the helper fits

The flagged line ranges are a sample, not a boundary. Any other instance of the same shape found while doing the work is in scope too.

## Acceptance Criteria

- [ ] A single shared helper (or fixture) covers the recording-handler + `create_listener` setup, living in `tests/integration/listener_helpers.py` or an appropriate `tests/support/` module
- [ ] Every call site in `tests/integration/test_listeners_dispatch.py` with this shape uses it
- [ ] The repeated three-shot `dispatch()` loop is extracted or collapsed
- [ ] Sibling occurrences in `tests/integration/test_listeners_validation.py` are either migrated or explicitly left alone with a reason
- [ ] No test loses coverage or changes what it asserts; `uv run pytest tests/integration/test_listeners_dispatch.py tests/integration/test_listeners_validation.py` passes
- [ ] The `duplicate-code` job no longer reports this block

NodeJSmith:
## Summary

`tests/integration/test_listeners_dispatch.py` repeated the same "recording listener" preamble at
nearly every call site — an empty `calls` list, a handler appending `event.data` to it, and a
`create_listener(...)` call that differed only in which rate-limiting kwarg it passed. Three tests
additionally carried a verbatim three-shot `dispatch()` loop. PMD CPD flagged both shapes as
duplicate blocks.

This extracts that setup into three helpers in `tests/integration/listener_helpers.py`, the module
that already exists for this test family:

- `make_recording_listener(bucket, **kw) -> (listener, calls)` — the recording-listener preamble,
  forwarding the rate-limiting option under test (`debounce=`, `throttle=`, `once=`) to
  `create_listener`.
- `dispatch_values(listener, *values)` — the repeated dispatch loop.
- `invoke_handler(bucket, handler)` — for the sync/async × event/no-event matrix in
  `TestListenerInvoke`, where the handler itself is the subject under test and so stays spelled out
  at the call site.

Net effect on the test file is 98 lines removed against 66 added, with no test changing what it
asserts.

`tests/integration/test_listeners_validation.py` keeps its local setup. Its handlers carry the
dependency-injection annotations the tests exist to exercise, and its remaining duplicate clusters
are `pytest.raises` blocks asserting that `create_listener` rejects an option combination — there
is no listener or `calls` list for a helper to return there.

The helper module docstring records this boundary so the next reader does not re-litigate it:
`make_recording_listener` covers the "handler appends `event.data` to a list" shape only.

## Testing

- `uv run nox -s dev` — green, 7737 passed, 1 skipped, 2 xfailed.
- `prek -a` — all hooks pass.
- Confirmed all 16 `test_listeners_dispatch.py` occurrences previously reported by
  `tools/check_duplicate_code.py` are gone, and the new helpers module introduces none.
- Verified before and after that nothing in `tests/integration/test_listeners_*.py` asserts on
  `identity.handler_name` or `short_name`, which `create_listener` derives from the handler object —
  so routing handler construction through a helper cannot silently change an assertion.

Closes #2037


NodeJSmith:
@coderabbitai review

PR #2109, opened 2026-09-09
test: enforce every ExecutionStatus member against the executions.status CHECK constraint

### LINKED ISSUE #2039: Enforce ExecutionStatus members match the executions.status CHECK constraint
## Description

Surfaced by a review of PR #2015, which trimmed the `ExecutionStatus` docstring down to the single durable invariant: *"Must stay in sync with the ``executions.status`` CHECK constraint."* That claim is accurate, but nothing enforces it — the gap predates PR #2015 (verified against `main`; the trimmed docstring's previous wording narrated migration history rather than enforcing anything either).

Today, `tests/unit/test_schema_migration.py` probes the CHECK constraint with two hardcoded literals only:

- `test_check_constraints_accept_skipped_status` — inserts `status="skipped"`
- `test_check_constraints_reject_invalid_status` — inserts `status="invalid"`

No test iterates `list(ExecutionStatus)` and asserts every member is accepted by the live constraint. A contributor who adds a sixth `ExecutionStatus` member without updating `src/hassette/migrations_sql/009.sql` gets no signal until a raw insert fails at runtime.

The wire-model side already has the right pattern to copy: `tests/unit/test_ws_models.py::test_every_execution_status_round_trips_through_wire_model` parametrizes over every enum member.

## Key Items

- `src/hassette/types/types.py` — `ExecutionStatus` declares five members and now explicitly claims sync with the SQL CHECK constraint.
- `src/hassette/migrations_sql/009.sql:23` — `CHECK (status IN ('success', 'error', 'cancelled', 'timed_out', 'skipped'))`, the constraint that must match.
- `tests/unit/test_schema_migration.py:88-142` — existing constraint tests, literal-based rather than enum-driven.
- `tests/unit/test_ws_models.py:262` — the parametrize-over-every-member pattern to mirror.

## Acceptance Criteria

- [ ] A test parametrized over `list(ExecutionStatus)` inserts an execution row for every member against a migrated DB and asserts each is accepted by the CHECK constraint.
- [ ] The test fails if a new `ExecutionStatus` member is added without a corresponding migration updating the constraint.
- [ ] Existing literal-based constraint tests (`reject_invalid_status`, `accept_skipped_status`) still pass or are subsumed by the new test.

NodeJSmith:
## Summary

`ExecutionStatus` documents itself as staying in sync with the `executions.status` CHECK constraint, but nothing enforced that claim. The schema tests probed the constraint with two hardcoded string literals (`'skipped'` and `'invalid'`), so adding a new enum member without a matching migration to widen the CHECK would pass the whole suite and only surface later as a raw insert failure against a real database.

This adds `test_check_constraints_accept_every_execution_status`, parametrized over `list(ExecutionStatus)`, which inserts one execution row per member against a freshly migrated database and asserts it round-trips. It mirrors the existing `test_every_execution_status_round_trips_through_wire_model` pattern, so the two tests now cover both halves of the enum's contract: the wire model and the live DB constraint.

Because the parametrization is derived from the enum itself rather than a hardcoded list, a new member added without a migration fails here immediately instead of in production.

## Changes

- `tests/unit/test_schema_migration.py` — new parametrized test over every `ExecutionStatus` member; import `ExecutionStatus` alongside the existing `SourceTier`.

Test-only change. No source, schema, or migration files are touched.

## Testing

- `uv run nox -s dev` — 7742 passed, 1 skipped, 2 xfailed, 0 failures.
- All 5 parametrizations of the new test pass (`success`, `error`, `timed_out`, `cancelled`, `skipped`), confirming the current CHECK constraint already accepts every member — the test documents and locks in existing correct behavior rather than fixing a live bug.
- `prek -a` — all hooks pass.

Closes #2039


NodeJSmith:
@coderabbitai review

PR #2114, opened 2026-09-09
chore(frontend): show backpressure chip for any non-default policy

### LINKED ISSUE #2045: Show backpressure chip for any non-default policy in listener detail
Surfaced by the automated quality scanner (llm-checker / lazy-checker / nitpicker) against `frontend/src/components/app-detail/listener-detail.tsx`. Pre-existing, not introduced by any in-flight branch.

## Finding: backpressure chip tests for a specific non-default policy instead of "not the default"

**Where:** `frontend/src/components/app-detail/listener-detail.tsx:30`

```tsx
if (listener.backpressure === "drop_newest") chips.push({ label: "backpressure", value: "drop_newest" });
```

**What's wrong**

`ModifierChips` exists to surface settings that deviate from the default — every other line in the function is a truthiness check on an optional field. This line breaks that pattern by hardcoding one specific enum member on both sides of the expression:

- The condition enumerates the non-default value (`=== "drop_newest"`) rather than excluding the default.
- The rendered `value` re-states the same literal instead of reading it off `listener.backpressure`, so the string appears twice on one line.

`BackpressurePolicy` (`src/hassette/types/enums.py:76`) is `BLOCK | DROP_NEWEST`, with `BLOCK` documented as "the default for all listeners". The generated frontend type is `"block" | "drop_newest"` (`frontend/src/api/generated-types.ts:939`).

Today the two forms are behaviorally identical. The problem is what happens when the backend enum grows: adding a third policy widens the generated union, this line still only matches `"drop_newest"`, and a listener configured with the new non-default policy silently renders no chip at all. Nothing in the type system flags it — the comparison stays valid against a wider union, so it fails silently in the UI rather than at build time.

**Proposed fix**

Invert the test to exclude the default and render the actual value:

```tsx
if (listener.backpressure !== "block") chips.push({ label: "backpressure", value: listener.backpressure });
```

This removes the duplicated literal, matches the "show what deviates from the default" intent of the surrounding function, and keeps working unchanged if `BackpressurePolicy` gains members. Behavior is identical for the two policies that exist today, so no visual change and no screenshot needed.

## Acceptance Criteria

- [ ] `listener-detail.tsx:30` tests against the default policy rather than enumerating a specific non-default one
- [ ] The chip's displayed value is read from `listener.backpressure`, not re-typed as a literal
- [ ] Existing listener chip tests (`frontend/src/components/app-detail/handlers-tab.listener.test.tsx`) still pass

## Notes on findings that were assessed and dropped

Three other scanner findings did not survive validity assessment and are recorded here only so they are not re-filed by a later scan:

- **Split `import type` / `import` from `../../api/endpoints` (lines 1-2).** This is the repo-wide convention — 30 files under `frontend/src` do the same, and `simple-import-sort/imports` (the enforced rule in `eslint.config.js`) sorts imports without merging type and value specifiers. Merging them here would make this file the outlier.
- **`isPending: loading` rename breaks boolean naming convention.** The identical rename appears in `job-detail.tsx:183` and `recent-activity-section.tsx:186`, and `loading` is the prop name `ExecutionSection` expects. Consistent, intentional.
- **Chip label strings should be extracted to shared constants.** There is no second consumer — `job-detail.tsx` builds a completely different chip set (`jitter`, `group`). Extracting display copy with one call site would be speculative abstraction.

NodeJSmith:
## Summary

`ModifierChips` in the listener detail panel decided whether to show the backpressure chip by
enumerating one specific non-default policy:

```tsx
if (listener.backpressure === "drop_newest") chips.push({ label: "backpressure", value: "drop_newest" });
```

This inverts the test to exclude the default instead, and reads the displayed value off the field
rather than repeating the literal:

```tsx
if (listener.backpressure !== "block") chips.push({ label: "backpressure", value: listener.backpressure });
```

## Why

Every other line in `ModifierChips` surfaces a setting that deviates from its default. This one
broke the pattern by hardcoding the same enum member on both sides of the expression.

`BackpressurePolicy` is `block | drop_newest` today, so the two forms are behaviorally identical
and there is no user-visible change. The problem is latent: if the backend enum gains a third
policy, the generated TypeScript union widens, the old comparison stays type-valid, and a listener
configured with the new policy silently renders no chip at all. Nothing fails at build time — it
just quietly stops showing information in the UI.

## Review follow-ups

The review battery (code / integration / WTF / llm / lazy / nitpicker) returned no blocking
findings. Two non-blocking items were addressed:

- **Coverage gap.** Nothing asserted the modifier chip's backpressure behavior — before this
  branch or after it. `handlers-tab.listener.test.tsx` touched `backpressure` only to exercise the
  "Backpressure Dropped" stats cell. Added two render assertions: a chip naming the policy for
  `drop_newest`, no chip for the default `block`. Verified they are meaningful by deleting the
  line under test and confirming the `drop_newest` case fails.
- **Unexplained literal.** `"block"` had nothing near it saying it is the default, which is the
  only reason it is the excluded value. Added a two-line comment stating that invariant and noting
  the chip renders the literal policy name users write in their registration call.

The `mkdocs.yml` `exclude_docs: pr-evidence/**` line in this diff already landed on `main`
independently; the merge is a no-op for that file.

## Testing

| Check | Command | Result |
|---|---|---|
| Frontend suite | `npx vitest run` | 1606 passed across 110 files |
| Listener chip tests | `npx vitest run src/components/app-detail/handlers-tab.listener.test.tsx` | 20 passed |
| Type check + build | `npm run build` (`tsc --noEmit && vite build`) | clean |
| Lint / hooks | `prek -a` | all hooks passed |

## Screenshots

Captured against the live demo stack at 1280px, dark theme. They confirm the two existing policies
render as expected — no chip for the default, chip with the policy name otherwise.

`block` (default) — no backpressure chip on `on_toggle_changed`:

listener detail with block backpressure policy

`drop_newest` — chip renders with the policy name:

listener detail with drop_newest backpressure policy

## CI is blocked by a `main` regression (#2177), not by this diff

`main` is currently red and GitHub runs PR checks against the head-merged-with-base ref, so this
PR inherits the failure no matter where its branch point sits. `stat-cell-builders.test.ts` imports
`COMMON_STAT_CELL_COUNT` from `stat-cell-builders.ts`, where it is declared `const` without
`export`:

```
src/components/app-detail/stat-cell-builders.test.ts(4,32): error TS2459: Module
'"./stat-cell-builders"' declares 'COMMON_STAT_CELL_COUNT' locally, but it is not exported.
```

That fails `tsc --noEmit`, which takes `frontend`, `build`, and every `e2e` job down with it. It is
tracked as #2177. On this branch's own base the same suite is 1606/1606 green and `npm run build`
is clean, and merging current `main` into an otherwise untouched tree reproduces all four
`stat-cell-builders` failures — so the regression is entirely upstream of this change.

Auto-merge (squash) is enabled. Once #2177 lands, these checks likely need a re-run (a base-branch
fix does not by itself re-trigger `pull_request` workflows) before auto-merge can fire.

Closes #2045


NodeJSmith:
@coderabbitai review

NodeJSmith:
Reviewed under the autofix review pass. CodeRabbit had no threads on this PR (auto-review is disabled and the earlier `@coderabbitai review` request produced no review). The full battery — code-reviewer, integration-reviewer, wtf-reviewer, llm-checker, lazy-checker, nitpicker — returned no blocking findings; the two non-blocking ones (no test coverage for the chip, unexplained `"block"` literal) are addressed in 092851e2.

CI is red from #2177, not from this diff: `stat-cell-builders.test.ts` imports `COMMON_STAT_CELL_COUNT`, which `stat-cell-builders.ts` declares `const` without `export`, so `tsc --noEmit` fails on the merge ref. This branch's own base is 1606/1606 green with a clean `npm run build` and `prek -a`. Auto-merge (squash) is enabled; the checks will likely need a re-run once #2177 lands.

NodeJSmith:
Second autofix review pass (unattended). No CodeRabbit threads to triage — auto-review is disabled and the earlier `@coderabbitai review` request still produced no review, so there was nothing to verify or resolve.

Re-ran the full battery on the current head. No CRITICAL or HIGH findings from any reviewer; code-reviewer and integration-reviewer both returned PASS, llm-checker returned CLEAN.

**Addressed in 244dbd9d** (three LOW items, no behavior change):

- The comment's second line restated what the code plainly does (renders the raw policy string) rather than explaining why the condition excludes the default instead of matching a named policy. Replaced with that rationale — the part a reader can't derive from the line itself.
- The two new tests used `listener_id` 31/32, outside the local id range of the tests they sit between (3 and 8). Moved to the free slots 4 and 7.
- Dropped the quotes around the policy name in one test title so the pair reads consistently.

**Verified rather than assumed:**

- Mutation-tested both new assertions. Deleting the `chips.push` line fails the `drop_newest` test; making the guard unconditionally true fails the `block` test. Both are load-bearing, in both directions.
- `BackpressurePolicy` is a required, non-nullable `"block" | "drop_newest"` (`generated-types.ts:939`), so `!== "block"` has no path to rendering a chip with an empty value.
- Frontend suite 1606/1606 across 110 files; `tsc --noEmit` clean; `prek -a` all hooks pass.

**Deliberately not changed:**

- *A test that would catch a revert to the old `=== "drop_newest"` form.* Not possible while the union has two members — the two forms are behaviorally identical for every value the type system can produce today, which #2045 says outright. The suggested compile-time exhaustiveness guard would be counterproductive here: `!== "block"` already handles a new member automatically, so a guard forcing the chip logic to be revisited on every new policy would exist only to be deleted.
- *A `DEFAULT_BACKPRESSURE_POLICY` constant* for the `"block"` literal shared by `listener-detail.tsx` and the test factory. In the tests the literal is the point — asserting "no chip for `DEFAULT_BACKPRESSURE_POLICY`" is circular and reads worse than the concrete value.
- *The `docs/pr-evidence/` screenshots.* Established convention, not this PR's invention: `origin/main` already carries 10 PNGs there from five prior merged PRs, and the `pr-evidence/**` exclusion is already in `mkdocs.yml` upstream. This branch's merge-base predates that merge, which is the only reason the three-dot diff shows the line as new; a two-dot diff against `origin/main` for `mkdocs.yml` is empty.

**CI is still blocked by #2177, not by this diff.** `stat-cell-builders.ts` still declares `COMMON_STAT_CELL_COUNT` without `export` on `origin/main` as of this pass, so `tsc --noEmit` fails on the head-merged-with-base ref and takes `frontend`, `build`, and all four `e2e` jobs with it. On this branch's own tree `tsc --noEmit` is clean, which is the direct confirmation that the failure is upstream. `file-sizes` and `duplicate-code` are the known informational baseline failures (`continue-on-error`).

Left the branch un-merged with `main` on purpose: merging would not change any check outcome while #2177 is unfixed, and the squash merge computes its diff against `main` at merge time anyway. Auto-merge (squash) is still enabled and now points at 244dbd9d; the checks will likely need a manual re-run once #2177 lands.

NodeJSmith:
Third autofix review pass (unattended). No code changes — nothing found that warranted one.

**CodeRabbit:** still no threads to triage (auto-review disabled; the earlier `@coderabbitai review` request never produced a review).

**Battery re-run on 244dbd9d:** code-reviewer PASS, integration-reviewer PASS (0 findings across all 11 dimensions), wtf-reviewer PASS, llm-checker CLEAN, lazy-checker CLEAN. nitpicker returned 3 findings, all one root cause — see below.

**nitpicker's `DEFAULT_BACKPRESSURE_POLICY` finding — skipped, and this pass verified the reason rather than inheriting it.** The claim is that `"block"` is duplicated across `listener-detail.tsx`, the new test, and `src/test/factories.ts` with no shared source of truth. The authoritative default is not in the frontend at all: it is `BackpressurePolicy.BLOCK` in `src/hassette/types/enums.py:109`, applied at `bus.py:609` and `listeners.py:112`. The generated OpenAPI union carries the *members* (`"block" | "drop_newest"`, `generated-types.ts:939`) but not which one is the default, so a frontend constant would be a third mirror of a Python value, not a source of truth — it relocates the drift risk instead of removing it. In the tests the literal is the assertion's subject, so naming it would be circular.

**Verified, not assumed:**

- Mutation-tested both new assertions again on this pass. Replacing the `chips.push` line fails the `drop_newest` test; forcing the guard to `true` fails the `block` test. Both directions are load-bearing. Working tree restored clean afterward.
- Frontend suite 1606/1606 across 110 files; `tsc --noEmit` clean on this branch's tree.
- `mkdocs.yml` two-dot diff against `origin/main` is empty — the `pr-evidence/**` line is already upstream, and `origin/main` already carries 10 PNGs under `docs/pr-evidence/`. The three-dot diff only shows it because of the merge base.

**CI is still blocked by #2177, which is still open.** Pulled the failing `frontend` job log directly: the sole error is `stat-cell-builders.test.ts(4,32): error TS2459: Module './stat-cell-builders' declares 'COMMON_STAT_CELL_COUNT' locally, but it is not exported` — a file this PR does not touch, and `origin/main` still declares that `const` without `export` as of this pass. That one `tsc` failure takes `frontend`, `build (3.13)`, all four `e2e` jobs, and `lint-complete` down with it. `file-sizes` and `duplicate-code` are the known informational baseline failures (`continue-on-error`).

Auto-merge (squash) remains enabled and points at 244dbd9d. Once #2177 lands the checks will likely need a manual re-run, since a base-branch fix does not re-trigger `pull_request` workflows on its own.

NodeJSmith:

PR #2119, opened 2026-09-09
test: extract the repeated recording-handler factory in test_listeners_timing

### LINKED ISSUE #2048: Extract the repeated recording-handler factory in test_listeners_timing.py
## Description

`tests/integration/test_listeners_timing.py` repeats the same "recording handler factory" setup in
six separate tests. Each copy is the identical five-line shape: a `calls: list[str]` accumulator,
a local `make_handler(label)` closure returning a no-arg async handler that appends its label to
that list, and then a `RateLimiter(...)` built against the test's `bucket` fixture.

```python
calls: list[str] = []

def make_handler(label: str):
    async def handler():
        calls.append(label)

    return handler
```

Flagged by the `duplicate-code` CI job (PMD CPD via `tools/check_duplicate_code.py`). This is not
a designed-parallel-structure case — it is a test helper that was never extracted, and
`.claude/rules/test-conventions.md` already points this kind of repeated shape at a shared
factory.

## Goal

Extract the shared recorder-and-labeled-handler setup into one helper and use it at every call
site with this shape. Something like a `make_recording_handler_factory()` returning the `calls`
list plus the factory, so a test reads:

```python
calls, make_handler = make_recording_handler_factory()
```

Where the helper lives is a judgment call for whoever picks this up: a module-level function in
`test_listeners_timing.py` is enough if it stays local, or `tests/support/helpers.py` alongside
the existing `make_controlled_clock` if it is likely to be reused by other rate-limiter tests.
Either is fine — the point is one definition instead of six.

## Where

The three locations PMD flagged as one block:

- `tests/integration/test_listeners_timing.py:16-24` (`test_debounce_delays_execution`)
- `tests/integration/test_listeners_timing.py:52-60` (`test_debounce_cancels_previous_calls`)
- `tests/integration/test_listeners_timing.py:104-112` (`test_debounce_reset_cancellation_is_silent`)

Three further byte-identical copies of the same factory found while reading, in the same file:

- `tests/integration/test_listeners_timing.py:175` (`test_throttle_limits_execution_frequency`)
- `tests/integration/test_listeners_timing.py:207` (`test_throttle_allows_first_call_with_zero_origin_clock`)
- `tests/integration/test_listeners_timing.py:248` (`test_throttle_tracks_time_correctly`)

There are also degenerate single-label variants that inline `async def handler(): calls.append("fired")`
without the factory (around lines 134 and 156). Fold those in only if the helper covers them
cleanly; do not contort the helper to absorb them. `test_throttle_with_no_args` (around line 220)
deliberately mutates a captured `label` between calls to prove the handler is re-read each time —
that is a different shape on purpose and should be left alone.

Any other instance of this same shape found while doing the work is in scope too, not just the
locations listed above.

## Acceptance Criteria

- [ ] A single shared helper provides the `calls` accumulator plus the labeled-handler factory.
- [ ] All six `def make_handler(label: str)` definitions in `tests/integration/test_listeners_timing.py` are replaced by calls to it.
- [ ] Any other occurrence of the same shape found during the work is converted too.
- [ ] `uv run pytest tests/integration/test_listeners_timing.py` passes.
- [ ] The `duplicate-code` job no longer reports this cluster.

NodeJSmith:
## Summary

`tests/integration/test_listeners_timing.py` contained six byte-identical copies of the same
"recording handler factory" setup — a `calls: list[str]` accumulator plus a local
`make_handler(label)` closure returning a no-arg async handler that appends its label. The
`duplicate-code` CI job (PMD CPD via `tools/check_duplicate_code.py`) flagged three of them as one
cluster; the other three turned up while reading the file.

This replaces all six with a single module-level `make_recording_handler_factory()` helper, so each
call site reads:

```python
calls, make_handler = make_recording_handler_factory()
```

Net effect on the file: 24 lines added, 42 removed.

## Why here and not `tests/support/helpers.py`

The issue left placement as a judgment call. The helper stays module-local because every consumer
lives in this one file and nothing else currently needs it. If a second rate-limiter test file grows
the same shape, promoting it to `tests/support/helpers.py` alongside `make_controlled_clock` is a
one-line move. The name does not shadow anything in the shared factory registry — the
`check_test_factories` hook passes.

- `test_throttle_with_no_args` mutates a captured `label` between calls to prove the handler is
  re-read each time. That is a different shape on purpose, per the issue, and is untouched.
- Three degenerate single-label variants that inline `calls.append("fired")` without a factory. The
  issue said to fold these in only if the helper covers them cleanly; it does not without contorting
  the signature, so they stay as-is.

## Testing

- `uv run nox -s dev` — 7736 passed, 1 skipped, 2 xfailed. One unrelated failure in
  `tests/integration/test_file_watcher.py::test_event_emitted_on_file_change`
  ("file_event_received was never set after 2 attempts"), which passes on rerun and whose file is
  byte-identical to `origin/main` — this branch touches only `test_listeners_timing.py`.
- `uv run pytest tests/integration/test_listeners_timing.py` — 14 passed.
- `prek -a` — all 29 hooks pass. `prek pyright -a --stage pre-push` — passes.
- `tools/check_duplicate_code.py` reports no remaining `test_listeners_timing` cluster.

Closes #2048


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2122, opened 2026-09-09
test: extract the repeated inline-app bring-up preamble in system app lifecycle tests

### LINKED ISSUE #2051: Extract the repeated inline-app bring-up preamble in test_app_lifecycle.py
## Description

`tests/system/test_app_lifecycle.py` repeats the same **inline-app bring-up preamble** in every test that defines its own app on the fly: make `tmp_path / "apps"` and `mkdir(exist_ok=True)` it, write an `app_code` string into it as a `.py` module, build the config with `make_system_config(ha_container, tmp_path)`, flip on autodetect for that directory with `enable_autodetect(config, apps_dir)`, enter `async with startup_context(config) as hassette:`, and then look the app instance back up by class name with `find_app(hassette, "SomeApp")`.

Only the app source string, the module filename, and the class name differ between copies. Everything around them is restated verbatim.

Worth knowing while doing this: `make_system_config()` (`tests/system/conftest.py:262`) already builds and `mkdir`s `tmp_path / "apps"` itself, so the per-test `apps_dir = tmp_path / "apps"` / `apps_dir.mkdir(exist_ok=True)` pair is re-deriving a path the config helper owns.

## Goal

Extract the shared write-app-then-start-Hassette sequence into a single helper (or fixture) — something along the lines of an async context manager `start_with_inline_app(ha_container, tmp_path, filename=..., source=...)` that writes the module, builds the autodetect-enabled config, enters `startup_context`, and yields the running `hassette` — and use it at every call site with this shape. A variant that accepts more than one module covers `test_multiple_apps_isolation`, which writes two.

The app source strings themselves should stay inline in each test: they *are* the subject of each test, and moving them to on-disk fixtures under `tests/system/apps/` would separate each test from the behavior it is asserting about. The recommended direction is to keep the source literal at the call site and extract only the plumbing around it.

Any other instance of this same shape found while doing the work is in scope too, not just the locations listed below.

## Where

- `tests/system/test_app_lifecycle.py:60-66`
- `tests/system/test_app_lifecycle.py:112-118`
- `tests/system/test_app_lifecycle.py:141-147`
- `tests/system/test_app_lifecycle.py:170-176`
- `tests/system/test_app_lifecycle.py:210-216`

Each flagged block is the tail of a longer copy that starts a few lines earlier at the `apps_dir = tmp_path / "apps"` / `apps_dir.mkdir(exist_ok=True)` pair — lines 49, 97, 130, 156, and 184 respectively. The full preamble, not just the flagged tail, is what should collapse into the helper.

`test_trivial_app_initializes` and `test_app_bus_handler_fires` use the same config/`startup_context`/`find_app` tail but source their apps from the pre-written `system_app_dir` fixture instead of writing one. They may or may not fit the same helper — decide deliberately rather than forcing them through it.

## Acceptance Criteria

- [ ] A shared helper (or fixture) for "write an inline app module and start Hassette with autodetect pointed at it" exists in `tests/system/conftest.py` (or `test_app_lifecycle.py` if it stays local to this file)
- [ ] Every test in `tests/system/test_app_lifecycle.py` that writes its own app module uses it instead of restating the preamble, including the two-module case in `test_multiple_apps_isolation`
- [ ] Per-test `apps_dir = tmp_path / "apps"` / `mkdir(exist_ok=True)` restatements are gone, with the directory owned in one place
- [ ] App source strings remain inline at each call site
- [ ] `uv run pytest tests/system/test_app_lifecycle.py` passes (requires Docker; CI's `nox -s system_with_coverage` also covers it)
- [ ] The `duplicate-code` job no longer reports this block

NodeJSmith:
## Summary

`tests/system/test_app_lifecycle.py` restated the same six-step inline-app bring-up preamble in
every test that defined its app on the fly: derive `tmp_path / "apps"`, `mkdir` it, write the app
source as a module, build the config with `make_system_config()`, flip on autodetect for that
directory, then enter `startup_context`. Only the source string, the module filename, and the
class name varied between copies.

This extracts the plumbing into a single async context manager, `start_with_inline_app()`, and
routes all five inline-app tests through it. Net effect on the file: 33 insertions, 42 deletions.

## What changed

- **New helper `start_with_inline_app(ha_container, tmp_path, **modules)`** — writes each module
  into the config's apps directory as `<stem>.py`, builds the autodetect-enabled config, enters
  `startup_context`, and yields the running `Hassette`. Modules are passed as keyword arguments
  (stem → source), so the two-module case in `test_multiple_apps_isolation` uses the same call
  shape as the single-module cases rather than needing a separate variant.
- **Dropped the per-test `apps_dir` restatements** — `make_system_config()` (`tests/system/conftest.py`)
  already builds *and* creates `tmp_path / "apps"`, so every test was re-deriving a path the config
  helper owns. The helper now reads `config.apps.directory` instead, leaving one owner for that path.
- **App source strings stay inline at each call site.** They are the subject of each test; moving
  them to on-disk fixtures would separate each test from the behavior it asserts about. Only the
  plumbing around them was extracted.

### Deliberately left alone

`test_trivial_app_initializes` and `test_app_bus_handler_fires` keep their existing
config/`startup_context` tail. They source their apps from the pre-written `system_app_dir` fixture
and write no modules of their own, so routing them through the helper would mean adding an
apps-directory override parameter used by nobody else — more surface than the duplication it removes.

## Testing

- `uv run nox -s dev` — **7737 passed**, 1 skipped, 2 xfailed.
- `prek -a` — all hooks pass (required `npm install` in the worktree first; worktrees do not share
  `frontend/node_modules`).
- `uv run pytest tests/system/test_app_lifecycle.py -m system --collect-only` — all 7 tests collect.

The changed file is a system test, which `nox -s dev` deselects by marker, so the local suite does
not execute it. CI's `nox -s system_with_coverage` runs it against a real Home Assistant container
and is the gate that actually exercises this change.

### Note on a flaky test seen during verification

The first `nox -s dev` run had one failure:
`tests/integration/test_scheduler_entity_time.py::test_waiting_job_removed_from_heap_but_stays_registered`
("waiting job must not be on the heap"). It is unrelated to this branch — the full branch diff
against `origin/main` is exactly one file, `tests/system/test_app_lifecycle.py`, and that file is
deselected by the `dev` session's marker filter, so every line of code the suite executed is
identical to `main`.

It passes in isolation and the full suite passed clean on re-run. It also sits in the same file as
two of the four tests catalogued in the (now closed) #1571 as timing-fragile, and shares their
failure mode: `change_alarm()` waits for bus dispatch to go idle, which does not guarantee the
scheduler has finished updating its heap. Flagging rather than fixing, since it is outside this
change's scope — worth a follow-up if it recurs.

Closes #2051


NodeJSmith:
@coderabbitai review

PR #2132, opened 2026-09-09
test: name the app-key constants in the recent-activity-section rerender test

### LINKED ISSUE #2055: Name the app-key constants in recent-activity-section.rerender.test.tsx
## Description

A quality scan of `frontend/src/components/app-detail/recent-activity-section.rerender.test.tsx` surfaced one pre-existing readability finding. Two of the three checkers (llm-checker, lazy-checker) returned clean; the file is otherwise tight and well-scoped.

### Load-bearing app-key coupling is expressed only through repeated string literals

The entire point of this test file is the distinction between "an execution belonging to the app under test" and "an execution belonging to some other app" — that is the seam `RecentActivitySection`'s store selector is being asserted against. But that distinction is carried purely by matching string literals scattered across three sites:

- `recent-activity-section.rerender.test.tsx:37` — `<RecentActivitySection appKey="test_app" ... />` sets the app under test.
- `recent-activity-section.rerender.test.tsx:65` — `makeExecution("test_app", kind)` must match line 37 for the "re-renders on its own app" assertion to mean anything.
- `recent-activity-section.rerender.test.tsx:54` — `makeExecution("other_app", ...)` (twice) must *not* match line 37 for the "does not re-render on an unrelated app" assertion to mean anything.

Nothing in the file states that lines 37 and 65 are required to be equal, or that line 54 is required to differ. A reader has to infer the invariant by string-matching across a 30-line span, and someone editing the `appKey` prop on line 37 without touching line 65 would silently turn the second test into a tautology (it would assert no re-render happens while claiming to assert one does) rather than failing loudly.

## Proposed Fix

Hoist two named constants at the top of the file and use them at all three sites, so the equal/not-equal relationship is stated rather than inferred:

```ts
const APP_KEY = "test_app";
const OTHER_APP_KEY = "other_app";
```

Then `appKey={APP_KEY}` on line 37, `makeExecution(APP_KEY, kind)` on line 65, and `makeExecution(OTHER_APP_KEY, ...)` on line 54.

## Acceptance Criteria

- [ ] `APP_KEY` and `OTHER_APP_KEY` constants are declared once at module scope in `recent-activity-section.rerender.test.tsx`.
- [ ] All three call sites (lines 37, 54, 65) reference the constants instead of inline string literals.
- [ ] `npx vitest run src/components/app-detail/recent-activity-section.rerender.test.tsx` still passes.

## Notes on findings that were assessed and excluded

For the record, several raised findings did not survive validity assessment and should not be actioned:

- **Import grouping** — claimed the external and local import tiers are not blank-line separated. Line 5 is a blank line, separating externals (lines 1-4) from locals (lines 6-13). The claim conflicts with the file.
- **`uptimeSeconds: 120` as a magic number** — the value is an established suite-wide idiom (`pages/app-detail.header.test.tsx:50`, `pages/app-detail.tabs.test.tsx:64`, `components/app-detail/handlers-tab.test-helpers.ts:20`, `pages/app-detail.rerender.test.tsx:62`), and non-null is load-bearing because `use-scoped-query` gates fetches on it for the since-restart preset. A file-local constant would diverge from five sibling files.
- **MSW route string `/api/telemetry/app/:app_key/activity`** — appears ~10 times across the suite, with `test/handlers.ts:104` already owning the canonical default handler. A file-local constant fixes nothing; centralizing it is a suite-wide change outside this file's scope.
- **`resolvedInstanceIndex={0}`, the single-use `overview-activity-empty` test-ID and `recent-activity` Profiler id, and renaming `counter` to `renderCounter`** — all single-use literals or already-clear names (`counter.renders` reads unambiguously at every use site). Extracting them adds indirection without reducing reader load.

This issue does not need human judgment — the surviving finding has one obvious mechanical fix.

NodeJSmith:
## Summary

The `recent-activity-section.rerender.test.tsx` suite asserts two opposite things about
`RecentActivitySection`'s store selector: it must re-render when an execution arrives for
the app under test, and must *not* re-render when one arrives for a different app. That
equal/not-equal relationship was carried entirely by matching string literals at three
separate call sites, with nothing in the file stating the invariant.

The failure mode this closes is a silent one. Changing the `appKey` prop without also
changing the matching literal in the "re-renders on its own app" case would turn that test
into a tautology — it would assert that no re-render happens while claiming to assert that
one does, and it would still pass.

## What changed

Two module-scope constants now carry the relationship explicitly:

```ts
const APP_KEY = "test_app";
const OTHER_APP_KEY = "other_app";
```

All three call sites reference them — the `appKey` prop on the rendered component, the
matching-app execution, and the two unrelated-app executions. No behavior change; the same
strings reach the same places.

Scope note: this is deliberately limited to the one finding that survived validity
assessment in #2055. The other candidates raised by the quality scan (import grouping, the
`uptimeSeconds: 120` idiom, the MSW route string, `resolvedInstanceIndex`) were assessed and
excluded there — several are suite-wide idioms where a file-local constant would diverge
from sibling files rather than help.

## Testing

- `npx vitest run src/components/app-detail/recent-activity-section.rerender.test.tsx` — 3 passed
- `prek -a` — all hooks passed (includes eslint, tsc, prettier for the frontend)

No screenshots: the only changed file is a `*.test.tsx`, which `check_pr_screenshots.py`
excludes from the rendered-frontend visual evidence requirement.

Closes #2055

### DIFF
--- a/frontend/src/components/app-detail/recent-activity-section.rerender.test.tsx
+++ b/frontend/src/components/app-detail/recent-activity-section.rerender.test.tsx
@@ -14,6 +14,9 @@ import { RecentActivitySection } from "./recent-activity-section";
 
 type ActivityFeedEntry = components["schemas"]["ActivityFeedEntry"];
 
+const APP_KEY = "test_app";
+const OTHER_APP_KEY = "other_app";
+
 function makeExecution(appKey: string, kind: ExecutionKind): WsExecutionCompletedPayload {
   return createExecutionCompletedPayload({ kind, app_key: appKey });
 }
@@ -34,7 +37,7 @@ async function renderSettled() {
         counter.renders += 1;
       }}
     >
-      <RecentActivitySection appKey="test_app" resolvedInstanceIndex={0} />
+      <RecentActivitySection appKey={APP_KEY} resolvedInstanceIndex={0} />
     </Profiler>,
     { storeOverrides: { uptimeSeconds: 120 } },
   );
@@ -51,7 +54,7 @@ describe("RecentActivitySection store subscriptions", () => {
     act(() => {
       useAppStore
         .getState()
-        .setExecutionCompleted([makeExecution("other_app", "handler"), makeExecution("other_app", "job")]);
+        .setExecutionCompleted([makeExecution(OTHER_APP_KEY, "handler"), makeExecution(OTHER_APP_KEY, "job")]);
     });
 
     expect(counter.renders).toBe(before);
@@ -62,7 +65,7 @@ describe("RecentActivitySection store subscriptions", () => {
     const before = counter.renders;
 
     act(() => {
-      useAppStore.getState().setExecutionCompleted([makeExecution("test_app", kind)]);
+      useAppStore.getState().setExecutionCompleted([makeExecution(APP_KEY, kind)]);
     });
 
     await waitFor(() => expect(counter.renders).toBeGreaterThan(before));


NodeJSmith:
@coderabbitai review

PR #2135, opened 2026-09-09
refactor: extract repeated manifest-lookup-and-skip guard in AppLifecycleService

### LINKED ISSUE #2056: Extract repeated manifest-lookup-and-skip guard in AppLifecycleService
`AppLifecycleService` repeats the same "look up the app manifest, log-and-skip if it's gone" guard in seven places. Every copy is byte-identical:

```python
app_manifest = self.registry.get_manifest(app_key)
if not app_manifest:
    self.logger.debug("Skipping disabled or unknown app %s", app_key)
    return
```

This is a single convention — *a missing or disabled manifest is a silent no-op skip, logged at DEBUG with one fixed message* — that currently lives as seven copies of the same four lines, including seven copies of the log string. Nothing distinguishes the call sites: the guard is uniform whether it runs before taking the app-key lock, after re-fetching under the lock, or inside an `_unlocked` helper.

## Goal

Extract the shared manifest-lookup-and-skip guard into one helper on `AppLifecycleService` and use it at every call site with this shape, so the "missing manifest is a skip" decision and its log message have a single definition. A resolver that logs and returns `None` keeps the caller's `return` local (the call sites are a mix of public methods and `_unlocked` helpers, so control flow has to stay with the caller):

```python
def _resolve_manifest(self, app_key: str) -> "AppManifest | None":
    """Fetch ``app_key``'s manifest, logging the standard skip message if it is absent."""
    app_manifest = self.registry.get_manifest(app_key)
    if app_manifest is None:
        self.logger.debug("Skipping disabled or unknown app %s", app_key)
    return app_manifest
```

Call sites then read:

```python
app_manifest = self._resolve_manifest(app_key)
if not app_manifest:
    return
```

Behavior must not change — same DEBUG message, same skip semantics, same placement relative to lock acquisition and to the post-lock re-fetch race guards (`start_app`, `start_instance`, and `reload_app` deliberately re-check the manifest *after* acquiring the lock because `_admit_start()` can park indefinitely; the re-fetch is load-bearing and must survive the refactor).

## Where

All in `src/hassette/core/app_lifecycle_service.py`:

- `start_app` — pre-lock guard (~494) and post-lock re-fetch guard (~507)
- `reload_app` — post-lock guard (~664)
- `_reload_instance_unlocked` (~832)
- `start_instance` — pre-lock guard (~884) and post-lock re-fetch guard (~894)
- `reload_changed_instances` (~1141)

PMD flagged four of these (494, 507, 664, 884); the other three surfaced while reading. **Any other instance of this same shape found while doing the work is in scope too** — the list above is evidence, not a boundary. In particular, sweep for other `registry.get_manifest(...)` call sites that log-and-skip on a missing manifest, including in `src/hassette/web/routes/apps.py`, where the equivalent check exists but maps to an HTTP response rather than a debug-log skip (leave those alone if the response shape differs — the goal is one definition of the *skip* convention, not forcing the HTTP paths through it).

## Acceptance Criteria

- [ ] A single manifest-resolution helper exists on `AppLifecycleService` and owns the skip log message.
- [ ] All seven listed call sites use it; no remaining copy of the inline four-line guard.
- [ ] The log string `"Skipping disabled or unknown app %s"` appears exactly once in the file.
- [ ] Post-lock re-fetch guards still re-fetch under the lock (the stale-manifest race guard is preserved).
- [ ] `prek -a` clean and the existing app-lifecycle tests pass with no behavior change.

NodeJSmith:
## Summary

`AppLifecycleService` repeated the same manifest-lookup-and-skip guard in seven places, byte-for-byte identical each time — including seven copies of the DEBUG log string:

```python
app_manifest = self.registry.get_manifest(app_key)
if not app_manifest:
    self.logger.debug("Skipping disabled or unknown app %s", app_key)
    return
```

That is one convention — *a missing or disabled manifest is a silent no-op skip, logged at DEBUG with one fixed message* — scattered across seven definitions. This extracts it into a single `_resolve_manifest()` helper that owns the lookup and the log message, and returns `None` when the manifest is absent.

## What changed

Added `AppLifecycleService._resolve_manifest(app_key)` and routed all seven call sites through it:

- `start_app` (pre-lock guard and post-lock re-fetch)
- `reload_app`
- `restart_instance`
- `start_instance` (pre-lock guard and post-lock re-fetch)
- `_reload_changed_instances`

Each call site keeps its own `return`, so control flow stays local. That matters here because the sites are a mix of public methods and `_unlocked` helpers — a helper that returned early on the caller's behalf would not work for both.

Net: 19 insertions, 14 deletions in `src/hassette/core/app_lifecycle_service.py`.

## Behavior is unchanged

This is a pure refactor. Specifically preserved:

- **Same DEBUG message and same skip semantics** at every site.
- **The post-lock re-fetch race guards are intact.** `start_app`, `start_instance`, and `reload_app` deliberately re-check the manifest *after* acquiring the app-key lock, because `_admit_start()` can park indefinitely and reconciliation may remove or replace the manifest while that wait is parked. Those re-fetches still happen under the lock — they now just call `_resolve_manifest()` instead of inlining the lookup. The load-bearing part of the guard survives the refactor.
- **Guard placement relative to lock acquisition is unchanged** at every site.

## Testing

- `uv run nox -s dev` — **7737 passed**, 1 skipped, 2 xfailed. This includes `TestStartAppStaleManifestRace::test_manifest_removed_during_admission_wait_is_not_used`, the regression test that directly covers the post-lock re-fetch guard this refactor touches.
- `prek -a` — all 29 hooks pass (ruff check, ruff format, house-lint, eslint, stylelint, tsc, and the project's custom checks).

No docs or frontend changes: this is internal framework plumbing with no user-facing behavior change, no new API surface, and no UI-visible data.

Closes #2056


NodeJSmith:
@coderabbitai review

PR #2139, opened 2026-09-09
test: use the exported cell-count constant and parameterize tone tests in stat-cell-builders

### LINKED ISSUE #2063: Clean up magic index and duplicated tone tests in stat-cell-builders.test.ts
Surfaced by the automated quality scanner (llm-checker + lazy-checker + nitpicker) against `frontend/src/components/app-detail/stat-cell-builders.test.ts`. Both the LLM-bias and deferred-debt passes came back clean; these two findings come from the style/hygiene pass and survived validity assessment. Neither is introduced by an open PR — they are pre-existing.

## Finding 1 — `cells.slice(5)` hardcodes a count the module already exports

**Where:** `frontend/src/components/app-detail/stat-cell-builders.test.ts:114`, `:129`, `:141`, `:154`

The literal `5` appears four times to mean "index where the conditional-cell zone starts." The module under test already exports exactly this value for exactly this purpose:

```ts
// stat-cell-builders.ts:5
/** Number of fixed cells `buildCommonStatCells` always produces, before any conditional cells. */
export const COMMON_STAT_CELL_COUNT = 5;
```

The constant currently has no consumer outside its own module (`resolveExtraCellIndex` at `stat-cell-builders.ts:38`) — it is exported but unused. If a fixed cell is ever added or removed, the source updates itself through the constant while all four test sites silently keep slicing at the wrong offset and start asserting against the wrong window.

**Proposed fix:** import `COMMON_STAT_CELL_COUNT` alongside `buildCommonStatCells` and replace the four `cells.slice(5)` calls with `cells.slice(COMMON_STAT_CELL_COUNT)`.

## Finding 2 — five near-identical conditional-cell tests

**Where:** `frontend/src/components/app-detail/stat-cell-builders.test.ts:68-100`

Five test blocks (`Timed Out`, `Cancelled`, `Thread Leaked`, `Suppressed`, `Dropped`) are structurally identical — set one input field, `find` the cell by label, assert `{ label, value, tone }`. Only four values vary per block (input field name, expected label, expected value, expected tone). Adding a sixth conditional cell means copy-pasting a block and remembering to change all four, which is exactly where copy-paste drift shows up.

**Proposed fix:** collapse into a single `it.each` table keyed on `[inputField, label, value, tone]`, keeping a templated test name so each case still reports individually. The companion "omits conditional cells when their counts are zero" test (`:57-66`) can stay as-is — it asserts the aggregate absence, not per-field behavior.

## Findings assessed as likely invalid (not included above)

- **Repeated `"Skipped"` string literal (7 occurrences, `:125-160`).** These are literal test inputs and literal expected outputs. Hoisting them into a shared constant would make the assertions compare a constant against itself, weakening the tests rather than improving them — stating expectations literally is the point.
- **Repeated tone literals `"err"`/`"warn"`/`"cancel"`/`"mute"`.** Same reasoning: expected values in assertions should be spelled out literally, not referenced from a shared symbol.
- **Single-letter callback parameter `c`.** The repo's own style guidance permits short, contextually obvious names in local scope; `c` inside `cells.map((c) => c.label)` is unambiguous and idiomatic.

## Acceptance Criteria

- [ ] All four `cells.slice(5)` sites use the imported `COMMON_STAT_CELL_COUNT` constant.
- [ ] The five conditional-cell tone tests at `:68-100` are expressed as one parameterized table.
- [ ] `cd frontend && npm test -- stat-cell-builders` passes with the same set of behaviors covered as before.

NodeJSmith:
## Summary

Two mechanical cleanups in `stat-cell-builders.test.ts`, surfaced by the automated quality scanner. Both are test-only — no production code or behavior changes.

**Use the exported constant instead of a magic index.** Four assertions sliced the cell array at a hardcoded `5` to reach the conditional-cell zone. The module under test already exports `COMMON_STAT_CELL_COUNT = 5` for exactly this purpose, and until now it had no consumer outside its own file. If a fixed cell is ever added or removed, `stat-cell-builders.ts` updates itself through the constant while all four test sites would have silently kept slicing at the wrong offset and started asserting against the wrong window. The tests now import the constant and slice through it.

**Collapse five copy-pasted tone tests into one table.** The `Timed Out`, `Cancelled`, `Thread Leaked`, `Suppressed`, and `Dropped` tests were structurally identical — set one input field, `find` the cell by label, assert `{ label, value, tone }`. Only the input field and the three expected values varied. They are now a single `it.each` table with a templated name, so each case still reports individually and a sixth conditional cell is one row rather than a copy-pasted block with four things to remember to change.

Net effect is 20 insertions, 36 deletions.

## Testing

- `npx vitest run src/components/app-detail/stat-cell-builders.test.ts` — 16 passed. This is the same test count as `main` (11 `it()` blocks plus 5 generated by the `it.each` table), so the parameterization preserved coverage rather than collapsing cases.
- Confirmed `COMMON_STAT_CELL_COUNT` is `5`, matching the literal it replaced — the slice windows are unchanged.
- `uv run nox -s dev` — 7737 passed, 1 skipped, 2 xfailed.
- `prek -a` — all hooks passed, including `tsc`, `eslint`, and `prettier`.

No screenshots: the only changed file is a `.test.ts`, so nothing rendered by the frontend is affected.

Closes #2063

### DIFF
--- a/frontend/src/components/app-detail/stat-cell-builders.test.ts
+++ b/frontend/src/components/app-detail/stat-cell-builders.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "vitest";
 
-import { buildCommonStatCells, type CommonStatInput } from "./stat-cell-builders";
+import type { DetailStatsCell } from "../shared/detail-stats";
+import { buildCommonStatCells, COMMON_STAT_CELL_COUNT, type CommonStatInput } from "./stat-cell-builders";
 
 function baseInput(overrides: Partial<CommonStatInput> = {}): CommonStatInput {
   return {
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #2142, opened 2026-09-09
refactor(frontend): decouple stat cell labels from lookup literals

### LINKED ISSUE #2065: Clean up label-literal coupling and redundant comments in stat-cell-builders.ts
Surfaced by the automated quality scanner (llm-checker + lazy-checker + nitpicker) against `frontend/src/components/app-detail/stat-cell-builders.ts`. All findings are pre-existing and local to this one file; no call sites change.

## Finding 1 — Cell-label literals are duplicated between the lookup and the construction site

`resolveExtraCellIndex` locates the splice point by matching label text (`stat-cell-builders.ts:32,35`), while `buildCommonStatCells` writes those same label strings independently (`stat-cell-builders.ts:53,54`):

```ts
const cancelledIndex = cells.findIndex((cell) => cell.label === "Cancelled");  // :32
const timedOutIndex  = cells.findIndex((cell) => cell.label === "Timed Out");  // :35
...
if (input.timedOut > 0)  cells.push({ label: "Timed Out", value: ..., tone: "warn" });   // :53
if (input.cancelled > 0) cells.push({ label: "Cancelled", value: ..., tone: "cancel" }); // :54
```

Nothing ties the two pairs together but naming discipline. Renaming a user-visible label in the `cells.push` call silently breaks `extraCell` placement — the `findIndex` returns `-1`, the function falls through to `COMMON_STAT_CELL_COUNT`, and the "Skipped" cell lands in the wrong position with no compiler error and no test failure unless a test happens to cover that exact ordering.

**Proposed fix:** hoist `const LABEL_TIMED_OUT = "Timed Out"` and `const LABEL_CANCELLED = "Cancelled"` to module scope and reference them from both the `findIndex` predicates and the `cells.push` calls.

## Finding 2 — `input.failed > 0 ? "err" : undefined` is computed twice

The identical derived tone appears on both the "Failed" cell and the "Err %" cell (`stat-cell-builders.ts:44,48`). The two are meant to move together — an "Err %" cell in a different tone than its own "Failed" count would be a bug — but nothing enforces that today.

**Proposed fix:** compute it once as a local (`const failedTone = input.failed > 0 ? "err" : undefined;`) before the array literal and reference it in both cells.

## Finding 3 — `COMMON_STAT_CELL_COUNT` is exported but has no external consumer

`stat-cell-builders.ts:5` exports the constant, but its only reference anywhere in `frontend/src` is line 38 inside this same module. The two real callers (`job-detail.tsx`, `listener-detail.tsx`) and the co-located test (`stat-cell-builders.test.ts`) all import only `buildCommonStatCells` and `CommonStatInput`.

**Proposed fix:** drop the `export` keyword and keep it module-private.

## Finding 4 — Redundant and decaying comments around the `extraCell` placement rule

Two problems in the comment block spanning `stat-cell-builders.ts:20-30`:

- The docstring on `resolveExtraCellIndex` (`:30`) states the rule is "per the rule documented on `CommonStatInput.extraCell`" — a pure back-reference that adds no information the six-line function body below it doesn't already make obvious. The interface JSDoc is the right home for the rule (it is what a caller reads); this second docstring just points at it.
- The JSDoc's closing sentence, "Only `JobDetail`'s 'Skipped' cell needs this today" (`:24-25`), is a point-in-time usage claim rather than a durable constraint. It is already the kind of statement that goes stale the moment a second caller passes `extraCell`, and a reader has no way to tell whether it is still true without grepping.

**Proposed fix:** keep the placement rule on `CommonStatInput.extraCell` (it is genuinely non-obvious and belongs on the public surface), drop the trailing "only JobDetail needs this today" sentence, and reduce the `resolveExtraCellIndex` docstring to nothing or a single clause that is not a pointer back to the interface.

## Considered and rejected

The `"warn"` tone literal repeated at `:53,56,58` was raised as a magic-string finding but is **not** included here. `DetailStatsCell.tone` is typed `StatusKind` (`components/shared/detail-stats.tsx:8`), a closed union, so a typo is a compile error rather than a silent break — unlike the label strings in Finding 1, which are free-form text. Extracting constants for a compile-checked union would add indirection without removing risk.

## Acceptance Criteria

- [ ] `"Timed Out"` and `"Cancelled"` each appear exactly once as a literal, referenced by shared constants from both the `findIndex` predicates and the `cells.push` calls
- [ ] The failed-tone ternary is computed once and reused by the "Failed" and "Err %" cells
- [ ] `COMMON_STAT_CELL_COUNT` is no longer exported
- [ ] The `resolveExtraCellIndex` back-reference docstring and the "only `JobDetail` needs this today" sentence are gone; the placement rule remains documented on `CommonStatInput.extraCell`
- [ ] `stat-cell-builders.test.ts` still passes unchanged

NodeJSmith:
## Summary

Removes the implicit coupling between the cell labels `buildCommonStatCells` writes and the label literals `resolveExtraCellIndex` searches for in `frontend/src/components/app-detail/stat-cell-builders.ts`.

`resolveExtraCellIndex` located the `extraCell` splice point by matching label text (`cell.label === "Cancelled"`), while `buildCommonStatCells` wrote those same strings independently. Nothing but naming discipline tied the two together: renaming a user-visible label at the construction site would make `findIndex` return `-1`, silently falling through to the `COMMON_STAT_CELL_COUNT` default and placing the extra cell in the wrong position — no compile error, no test failure.

## What changed

- **Hoisted `LABEL_TIMED_OUT` and `LABEL_CANCELLED` to module constants.** Each label now appears exactly once as a literal and is referenced by both the `findIndex` predicates and the `cells.push` calls, so the lookup and the construction site share one source of truth.
- **Computed the failed tone once.** `input.failed > 0 ? "err" : undefined` was duplicated across the "Failed" and "Err %" cells; it is now a single `failedTone` local reused by both. An "Err %" cell toned differently from its own "Failed" count would be a bug, and this makes the two structurally unable to diverge.
- **Un-exported `COMMON_STAT_CELL_COUNT`.** Its only reference anywhere in `frontend/src` was inside this same module — confirmed by grep across the repo and by a clean `tsc --noEmit` after the change.
- **Trimmed two decaying comments.** Dropped the `resolveExtraCellIndex` docstring (a pure back-reference to the interface JSDoc that restated what the six-line body already shows) and the "Only `JobDetail`'s 'Skipped' cell needs this today" sentence (a point-in-time usage claim that goes stale the moment a second caller passes `extraCell`). The genuinely non-obvious placement rule stays documented on `CommonStatInput.extraCell`, where callers read it.

No behavior change and no call sites touched — `job-detail.tsx` and `listener-detail.tsx` import only `buildCommonStatCells` and `CommonStatInput`, both unchanged.

Per the issue's "Considered and rejected" note, the repeated `"warn"` tone literal was deliberately left alone: `DetailStatsCell.tone` is typed as the closed `StatusKind` union, so a typo there is already a compile error. Extracting constants for a compile-checked union adds indirection without removing risk.

## Testing

- `stat-cell-builders.test.ts` passes unchanged (16/16) — the test file is not modified in this branch, which is what makes it a meaningful regression check on the refactor.
- Full frontend suite: 110 files, 1604 tests passed.
- `npm run typecheck` (`tsc --noEmit`): clean. This is what verifies the un-export broke no importer.
- `uv run nox -s dev`: 7737 passed, 1 skipped, 2 xfailed.
- `prek -a`: all 29 hooks passed; `prek pyright -a --stage pre-push`: passed.

No screenshots: this is a `.ts` logic module with no rendered output change, and the `pr-screenshots` guard scopes to `frontend/src/**/*.{tsx,css}`.

Closes #2065

### DIFF
--- a/frontend/src/components/app-detail/stat-cell-builders.ts
+++ b/frontend/src/components/app-detail/stat-cell-builders.ts
@@ -2,7 +2,10 @@ import { formatDurationOrDash, formatRate } from "../../utils/format";
 import type { DetailStatsCell } from "../shared/detail-stats";
 
 /** Number of fixed cells `buildCommonStatCells` always produces, before any conditional cells. */
-export const COMMON_STAT_CELL_COUNT = 5;
+const COMMON_STAT_CELL_COUNT = 5;
+
+const LABEL_TIMED_OUT = "Timed Out";
+const LABEL_CANCELLED = "Cancelled";
 
 export interface CommonStatInput {
   totalLabel: string;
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2148, opened 2026-09-09
test: add stderr_for_malformed_response helper to CLI client test family

### LINKED ISSUE #2080: Extract the malformed-response stderr helper in CLI client tests
## Description

CLI client tests repeatedly hand-roll the same "malformed 2xx response" arrangement: build a config, build a `make_transport(200, {"unexpected": "shape"})` transport (valid JSON that doesn't match the requested model), construct a `HassetteCLIClient(config, json_mode=False, transport=transport, ...)`, then call `get_expecting_exit(client)` and assert on the human-mode stderr it produced. The only things that vary are the config used to build the client and the extra client flags (`debug_mode=True`, `verify_ssl_flag=False`, `server_url_flag=None`) — the three-line arrangement in between is identical every time.

`tests/unit/cli/test_client.py` already has a helper family for exactly this shape: `stderr_for_successful_get(config, **client_kwargs)` and `stderr_for_connect_error(config, **client_kwargs)`, each of which absorbs "build a client from this config with these flags, exercise this failure surface, hand back the stderr." The malformed-response surface — the third member of that family — never got a helper, so every test that needs it rebuilds the arrangement by hand. `tests/unit/cli/test_client_credentials.py` already imports both existing siblings from `test_client.py`, so it inherits the gap: its two TLS-warning-on-malformed-response tests hand-roll the arrangement while its network-error siblings right above them call `stderr_for_connect_error()`.

## Fix

Add the missing member of the `stderr_for_*` helper family — a `stderr_for_malformed_response(config, **client_kwargs)` alongside `stderr_for_successful_get` and `stderr_for_connect_error` at the top of `test_client.py` — and use it at every call site with this shape, in both files. Keep what carries test meaning at the call site: the config, the client flags, and the assertion on the returned stderr. Absorb only the transport construction and client wiring.

Some nearby call sites use the same wrong-shape transport but diverge in what they exercise (json-mode error envelopes via `get_json_error`, `client.post` instead of `client.get`, a 503 body under `tolerate_503=True`, or an exit-code assertion instead of a stderr one). Fold those in where the helper family can reach them cleanly and leave them alone where it can't — do not contort the helper to cover every variant.

Any other instance of the same shape found while doing the work is in scope too, not just the locations listed below.

## Where

PMD CPD flagged these as one duplicated block:

- `tests/unit/cli/test_client.py:308-312`
- `tests/unit/cli/test_client.py:326-330`
- `tests/unit/cli/test_client_credentials.py:320-324`

Other occurrences of the same shape found while reading:

- `tests/unit/cli/test_client.py:366-373` (`test_debug_mode_shows_url_and_body` — same arrangement plus `debug_mode=True`)
- `tests/unit/cli/test_client_credentials.py:326-332` (`test_flag_sourced_insecure_does_not_warn_on_malformed_response_path` — same arrangement plus `verify_ssl_flag`/`server_url_flag`)

Adjacent call sites sharing the wrong-shape transport but a different exercise surface (fold in only where it reads cleanly): `test_client.py:299-305`, `316-323`, `244-251`, `405-411`, `413-419`.

## Relationship to #2041 and #2078

All three clusters live in the same file and overlap at some call sites, but absorb different things and are separable. #2041 absorbs the `pytest.raises(SystemExit)` + exit-code assertion scaffolding (the call sites flagged here already route through `get_expecting_exit`, so #2041 does not cover them). #2078 absorbs raw-bytes transport construction (the call sites flagged here use the JSON `make_transport`, so #2078 does not cover them either). This one absorbs the config-to-stderr client wiring for the malformed-response surface. Worth sequencing all three together to avoid churn on the same lines.

## Acceptance Criteria

- [ ] A `stderr_for_malformed_response`-style helper lives alongside `stderr_for_successful_get` and `stderr_for_connect_error` at the top of `tests/unit/cli/test_client.py`, taking a config plus `**client_kwargs` like its siblings.
- [ ] Every call site in `test_client.py` and `test_client_credentials.py` that builds a wrong-shape-200 client purely to assert on its human-mode stderr uses the helper, with no remaining hand-rolled copies.
- [ ] Configs, client flags, and the stderr assertions stay visible at the call site — the helper does not hide what each test is actually asserting.
- [ ] `uv run pytest tests/unit/cli/test_client.py tests/unit/cli/test_client_credentials.py` passes.

NodeJSmith:
## Summary

CLI client tests had a `stderr_for_*` helper family — `stderr_for_successful_get` and
`stderr_for_connect_error` — that absorbs "build a client from this config with these flags,
exercise this failure surface, hand back the stderr." The malformed-response surface (a 200
carrying valid JSON that doesn't match the requested model) never got a member, so every test
needing it rebuilt the same three-line arrangement by hand: build the config, build a
`make_transport(200, {"unexpected": "shape"})` transport, construct the client, call
`get_expecting_exit`.

This adds the missing third sibling, `stderr_for_malformed_response(config, **client_kwargs)`,
and routes every call site with that shape through it across both files.

## What changed

- **`tests/unit/cli/test_client.py`** — new `stderr_for_malformed_response` helper alongside its
  two siblings at the top of the file. Three call sites converted
  (`test_valid_json_wrong_shape_200_prints_clean_error_human_mode`, `test_non_loopback_shows_target`,
  `test_debug_mode_shows_url_and_body`).
- **`tests/unit/cli/test_client_credentials.py`** — imports the new helper next to the two it
  already imported, and converts its two TLS-warning-on-malformed-response tests, which were
  hand-rolling the arrangement while their network-error siblings directly above already called
  `stderr_for_connect_error`.

Net: 20 insertions, 21 deletions across the two files.

The issue called for folding in adjacent call sites only where the helper reaches them cleanly,
without contorting it to cover every variant. These share the wrong-shape transport but exercise a
different surface, so they keep their explicit wiring:

- `test_503_with_wrong_shape_json_exits_instead_of_crashing` — 503 body under `tolerate_503=True`
- `test_valid_json_wrong_shape_200_exits_code_1` — asserts the exit code, not stderr
- `test_valid_json_wrong_shape_200_json_mode_error_envelope` — json-mode envelope via `get_json_error`
- `test_schema_mismatch_exits_instead_of_crashing` / `test_schema_mismatch_json_mode_writes_error_doc` — `client.post`, not `client.get`

Each test's config, client flags, and stderr assertion stay visible at the call site — the helper
absorbs only transport construction and client wiring, so what a test actually asserts is unchanged.

## Testing

- `uv run nox -s dev` — 7737 passed, 1 skipped, 2 xfailed (unit + integration)
- `prek -a` — all 29 hooks pass

No source changes, so no docs or frontend updates apply.

Closes #2080


NodeJSmith:
@coderabbitai review

PR #2154, opened 2026-09-09
chore: trim over-explained comments in the diagnostics Panel component

### LINKED ISSUE #2085: Trim over-explained comments in diagnostics Panel
Automated quality scan of `frontend/src/components/diagnostics/panel.tsx`. Three checkers ran (llm-checker, lazy-checker, nitpicker); the surviving findings all land on one theme: comment volume out of proportion to a 50-line file. Roughly a third of the file's lines are prose explaining mechanics a reader can see directly, and each comment's *content* is defensible — it's the length and placement that are the problem.

Panel and `PANEL_BANNER_CLASS` are genuinely shared (4 consumer files), so none of this is dead code or a speculative abstraction. lazy-checker returned clean.

## Finding 1 — 6-line JSDoc on a one-line class constant (`panel.tsx:9-15`)

`PANEL_BANNER_CLASS` is a single string. Above it sits a six-line block explaining how `cn()`/`twMerge` resolves conflicting classes, which specific classes it overrides on `AlertShell`, and why the text styles live there. That's a tutorial aside on Tailwind merge precedence, not something a maintainer of this file needs restated.

This comment has already cost maintenance once: commit `931a90bd` ("correct the inaccurate PANEL_BANNER_CLASS doc comment") existed solely to fix drift in it. Long explanatory prose about another component's defaults is exactly the kind of comment that goes stale when that component changes.

**Proposed fix:** collapse to one line stating the durable invariant only — that the constant flattens `AlertShell`'s standalone margin/radius for use inside a panel's flex column. Drop the `twMerge` mechanics walkthrough and the enumeration of the classes being overridden.

## Finding 2 — rationale comment attached to an import statement (`panel.tsx:3-5`)

Three lines above `import { cardVariants } from "@/components/ui/card"` explain why `cardVariants()` is applied to a `<section>` rather than rendering `<Card>`. The reasoning is real and worth keeping — losing the landmark semantics would be a genuine accessibility regression — but it is documenting a decision made at line 39, not at the import.

**Proposed fix:** condense to one sentence and move it to the `<section className={cn(cardVariants(...))}>` call site, where a reader considering "why not just use `<Card>`?" will actually encounter it. Keep the a11y reason; drop the expanded explanation.

## Finding 3 — 4-line JSDoc on the `headingAside` prop (`panel.tsx:25-29`)

The prop is `headingAside?: ReactNode`. Its doc comment spends four lines on an example, a claim about layout stability, and a statement that omitting the prop, `null`, and `undefined` are interchangeable — which is just how React renders optional children, visible from the three-line flex row below it.

**Proposed fix:** reduce to one line describing what it is and where it renders (e.g. a status chip beside the heading). Drop the absence-equivalence sentence and the layout-stability claim.

## Acceptance criteria

- [ ] `PANEL_BANNER_CLASS` doc reduced to a single line stating why the overrides exist, with no `twMerge`/precedence walkthrough
- [ ] The `<Card>`-vs-`<section>` a11y rationale survives, as one sentence, at the `<section>` call site rather than above the import
- [ ] `headingAside` doc reduced to a single line; absence-equivalence and layout-stability prose removed
- [ ] No behavior change — rendered markup, class strings, and exports identical

## Findings assessed as likely invalid

Three nitpicker findings were dropped with an evidence trail rather than filed:

- **"`PANEL_BANNER_CLASS` belongs near `AlertShell`, not in `panel.tsx`."** Claimed: the constant describes another component's styling and is misplaced. Actually: its only consumers are `logging-panel.tsx:18` and `telemetry-panel.tsx:24`, both siblings that import `Panel` from this same module, and its semantics are panel-specific — it exists to cancel `AlertShell`'s *standalone* spacing precisely because a `Panel`'s flex column already supplies the gap. Moving it beside `AlertShell` would separate it from the context that gives it meaning.
- **"`PANEL_BANNER_CLASS` mixes layout, shape, and color in one unstructured string."** and **"`PANEL_HEADING_CLASS` concatenates margin, font, size, weight, line-height, and color with no grouping."** Claimed: these strings should be visually grouped by concern. Actually: unbroken Tailwind strings combining exactly these axes are the repo-wide convention, mandated by CLAUDE.md ("Use Tailwind utilities directly in JSX", arbitrary values such as `text-[var(--handler-job)]` for project tokens) and used identically throughout this directory — `drop-counter-row.tsx:13`, `service-row.tsx:25`, `service-row.tsx:41`, `boot-issues-panel.tsx:32`. Adding grouping comments inside class strings would make this file diverge from every sibling.

## Notes for the fixer

This is a comments-only change and is mechanical, but the rationale in Findings 1 and 2 is load-bearing: the `<section>`-not-`<Card>` reason and the "panel supplies the gap" reason must survive the trim in some form. The goal is shorter comments, not fewer facts.

NodeJSmith:
## Summary

`frontend/src/components/diagnostics/panel.tsx` is a ~50-line file where roughly a third of the
lines were prose restating mechanics the code already shows. Each comment's *content* was
defensible — the length and placement were the problem. This trims them without dropping any
load-bearing rationale.

## What changed

- **`PANEL_BANNER_CLASS`** — dropped the `twMerge` precedence walkthrough and the enumeration of
  which `AlertShell` classes get overridden, keeping the one durable reason the overrides exist
  (a panel's flex column already supplies the gap). This comment had already drifted once and
  needed a dedicated correction commit (`931a90bd`), which is the usual fate of long prose about
  another component's defaults.
- **`<Card>`-vs-`<section>` a11y rationale** — moved from above the `cardVariants` import down to
  the `<section>` call site, where the decision it documents is actually made. The reason itself
  survives: rendering `<Card>` (a `<div>`) would cost the panel its landmark semantics for
  screen-reader users navigating by `aria-label`.
- **`headingAside` prop doc** — dropped the absence-equivalence and layout-stability sentences,
  which just restate how React renders an optional child.

Also adds `docs/pr-evidence/` for PR visual evidence: allowlisted in `.gitignore` (which otherwise
ignores `*.png`) and excluded from the built docs site via `mkdocs.yml`'s `exclude_docs`.

## Testing

- **No behavior change, verified mechanically:** compiled the pre- and post-change file with
  esbuild — byte-identical output. Rendered markup, class strings, and exports are unchanged.
- `uv run nox -s dev` — 7736 passed, 1 skipped, 2 xfailed. One unrelated flake in
  `tests/integration/test_file_watcher.py::test_event_emitted_on_file_change` (a timing-sensitive
  watcher test); it passes in isolation, and this branch changes no Python or test files at all
  (`git diff origin/main --name-only -- src/ tests/` is empty).
- `prek -a` — all hooks pass, including `tsc`, `eslint`, `prettier`, and `stylelint`.
- Screenshots below captured against the live demo stack to confirm the diagnostics page renders
  identically in both themes.

## Screenshots

Diagnostics page, 1280px, light theme:

Diagnostics page, light theme

Diagnostics page, 1280px, dark theme:

Diagnostics page, dark theme

Closes #2085


NodeJSmith:
@coderabbitai review

PR #2160, opened 2026-09-09
test: extract repeated RobotMode converter registration into a fixture

### LINKED ISSUE #2091: Extract the repeated RobotMode type-converter registration in test_custom_conversion.py
## Description

`tests/unit/conversion/test_custom_conversion.py` re-declares the same `str -> RobotMode` type
converter inside the body of nearly every test that needs one. The block is a verbatim copy-paste:

```python
@register_type_converter_fn
def str_to_robot_mode(value: str) -> RobotMode:
    return RobotMode(value.lower())
```

The shared shape is "register the RobotMode converter for the duration of this test." The
`_isolate_type_registry` autouse fixture already snapshots and restores `TypeRegistry.conversion_map`
around every test, so registration is the only per-test part left — and it is identical each time.

The fix is to extract the registration into a single opt-in pytest fixture (e.g.
`robot_mode_converter`) and have every test with this shape request it instead of redeclaring the
converter inline.

**Important constraint:** the fixture must be opt-in, *not* `autouse`. Two tests deliberately depend
on the converter being absent or on registering their own variant:

- `test_without_converter_uses_constructor_fallback` asserts `(str, RobotMode) not in TypeRegistry.conversion_map`
- `test_decorator_with_error_message` registers the same converter with an `error_message=` argument
- `test_converter_is_used_by_state_pipeline` registers a call-recording variant that appends to a local list

Those three keep their local registration (or get their own narrow fixture); the plain-vanilla copies
are what collapses into the shared fixture.

## Where

PMD flagged these three as one duplicated block:

- `tests/unit/conversion/test_custom_conversion.py:150-155`
- `tests/unit/conversion/test_custom_conversion.py:160-165`
- `tests/unit/conversion/test_custom_conversion.py:170-175`

Reading the file turns up additional copies of the same shape beyond PMD's sample:

- `tests/unit/conversion/test_custom_conversion.py:193-195` (`test_bare_decorator_registers_converter`)
- `tests/unit/conversion/test_custom_conversion.py:201-203` (`test_decorator_with_error_message` — `error_message=` variant)
- `tests/unit/conversion/test_custom_conversion.py:211-214` (`test_converter_is_used_by_state_pipeline` — call-recording variant)

Any other instance of the same shape found while doing the work is in scope too, not just the
locations listed here.

## Acceptance Criteria

- [ ] A single opt-in fixture registers the plain `str -> RobotMode` converter, and every test with
      that exact shape requests it rather than declaring the converter inline.
- [ ] The fixture is not `autouse` — `test_without_converter_uses_constructor_fallback` still asserts
      the converter is absent and still passes.
- [ ] The `error_message=` and call-recording variants remain distinguishable (local, or their own
      fixture), not folded into the shared one.
- [ ] `uv run pytest tests/unit/conversion/test_custom_conversion.py` passes.
- [ ] `tools/check_duplicate_code.py` no longer reports this cluster.

NodeJSmith:
## Summary

`tests/unit/conversion/test_custom_conversion.py` re-declared the same plain `str -> RobotMode`
converter verbatim inside four separate tests. This replaces those copies with a single opt-in
`robot_mode_converter` fixture.

The `_isolate_type_registry` autouse fixture already snapshots and restores
`TypeRegistry.conversion_map` around every test, so registration was the only per-test part left —
and it was identical each time.

## What changed

- Added a `robot_mode_converter` fixture that registers the plain `str -> RobotMode` converter for
  the duration of one test.
- Four tests now request the fixture instead of declaring the converter inline:
  `test_custom_enum_coercion_via_model_validate`, `test_custom_enum_coercion_via_try_convert_state`,
  `test_custom_enum_unknown_yields_none`, and `test_bare_decorator_registers_converter`.

Net: 17 insertions, 20 deletions in one test file. No source changes.

## Why the fixture is opt-in, not autouse

Three tests deliberately depend on the converter being absent or on registering their own variant,
so they keep their existing local behavior:

- `test_without_converter_uses_constructor_fallback` — asserts `(str, RobotMode)` is *not* in
  `TypeRegistry.conversion_map`; an autouse fixture would break it.
- `test_decorator_with_error_message` — registers the same converter with an `error_message=`
  argument.
- `test_converter_is_used_by_state_pipeline` — registers a call-recording variant that appends to a
  local list.

Folding those into the shared fixture would erase the distinction each one is testing.

## Testing

- `uv run pytest tests/unit/conversion/test_custom_conversion.py` — 21 passed, including
  `test_without_converter_uses_constructor_fallback`.
- `uv run nox -s dev` — 7741 passed, 1 skipped, 2 xfailed, 0 failures.
- `prek -a` — all hooks pass; `prek pyright -a --stage pre-push` passes.
- `uv run python tools/check_duplicate_code.py` — no longer reports any
  `test_custom_conversion.py` cluster (the tool's remaining output is the pre-existing baseline
  backlog unrelated to this change).

Closes #2091

### DIFF
--- a/tests/unit/conversion/test_custom_conversion.py
+++ b/tests/unit/conversion/test_custom_conversion.py
@@ -59,6 +59,19 @@ class RobotState(BaseState[RobotMode | None]):
     value_type: ClassVar[type[Any] | tuple[type[Any], ...]] = (RobotMode, type(None))
 
 
+@pytest.fixture
+def robot_mode_converter(_isolate_type_registry: None) -> None:
+    """Register the plain ``str -> RobotMode`` converter for the duration of one test.
+
+    Opt-in rather than autouse: some tests need the converter absent, and others register
+    their own variant of it.
+    """
+
+    @register_type_converter_fn
+    def str_to_robot_mode(value: str) -> RobotMode:
+        return RobotMode(value.lower())
+
+
 class TestCustomStateWithAttributes:
     """Custom state models with typed AttributesBase subclasses."""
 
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #2162, opened 2026-09-09
test: extract shared per-app-key lock assertion in AppLifecycleService tests

### LINKED ISSUE #2093: Extract the per-app-key lock acquisition assertion in AppLifecycleService tests
## Description

`duplicate-code` (PMD CPD) flagged a copy-pasted block in the `AppLifecycleService` unit tests. The shared shape is **the "this operation acquires the per-app-key lock exactly once" assertion**: grab the lock via `_get_app_key_lock("test_app")`, replace `lock.acquire` with an `AsyncMock(wraps=lock.acquire)`, run the operation under `asyncio.wait_for(..., timeout=1)`, then assert `lock.acquire.call_count == 1` and `not lock.locked()`.

```python
lock = lifecycle_service._get_app_key_lock("test_app")
lock.acquire = AsyncMock(wraps=lock.acquire)

await asyncio.wait_for(lifecycle_service.<operation>(...), timeout=1)

assert lock.acquire.call_count == 1
assert not lock.locked()
```

Only the operation under test and its per-test mock setup differ between copies. The `wait_for(..., timeout=1)` wrapper is load-bearing (it turns a non-reentrant-lock deadlock into a test failure instead of a hung run) and is exactly the kind of detail that silently drifts when it lives in four places.

## Goal

Extract the shared lock-acquisition assertion into a helper in `tests/unit/core/_fixtures_app_lifecycle.py` — something like `assert_acquires_app_key_lock_once(lifecycle_service, app_key, operation)` taking the awaitable/callable for the operation under test — and use it at every call site with this shape. That module is already the home for this kind of shared lifecycle-test helper (`assert_load_completed_count`, `set_registry_apps`, `stub_detected_changes`), so it needs no new file.

Keep the per-test mock setup (`mock_registry.get_manifest`, `mock_factory.*`, method stubs) in the tests — that is genuinely per-operation and not part of the duplicated shape.

## Where

PMD flagged three occurrences:

- `tests/unit/core/test_app_lifecycle_service_per_instance_ops.py:348-353` — `stop_instance`
- `tests/unit/core/test_app_lifecycle_service_per_instance_ops.py:371-376` — `start_instance`
- `tests/unit/core/test_app_lifecycle_service_reload.py:154-159` — `reload_app`

Reading the surrounding files turned up a fourth copy of the identical shape that PMD did not list:

- `tests/unit/core/test_app_lifecycle_service_per_instance_ops.py:328-335` — `reload_instance`

**Any other instance of the same shape found while doing the work is in scope too, not just what is listed here.**

While reading, a related-but-distinct sibling pattern also showed up — the "second caller is queued and blocked on the app-key lock" concurrency assertion (`await wait_for(lambda: bool(lock._waiters), ...)` followed by `assert lock.locked()` / `assert not task2.done()`), at `test_app_lifecycle_service_per_instance_ops.py:420`, `test_app_lifecycle_service_reload.py:206`, and `test_app_lifecycle_service_coverage.py:80`. It reaches into `asyncio.Lock._waiters` (private API) in three places and would benefit from the same treatment, but it is a different assertion than the one flagged here — fold it in only if it falls out naturally, and otherwise leave it alone.

## Acceptance Criteria

- [ ] A shared helper for the "acquires the per-app-key lock exactly once" assertion exists in `tests/unit/core/_fixtures_app_lifecycle.py`
- [ ] All four call sites above use the helper instead of an inline copy
- [ ] The `asyncio.wait_for(..., timeout=1)` deadlock guard is preserved inside the helper
- [ ] Each test still names which operation it covers, so a failure points at the right lifecycle method
- [ ] `uv run pytest tests/unit/core/test_app_lifecycle_service_per_instance_ops.py tests/unit/core/test_app_lifecycle_service_reload.py` passes

NodeJSmith:
## Summary

The "this operation acquires the per-app-key lock exactly once" assertion was copy-pasted across
four `AppLifecycleService` locking tests. Every copy did the same four things: grab the lock via
`_get_app_key_lock()`, wrap `lock.acquire` in an `AsyncMock`, run the operation under
`asyncio.wait_for(..., timeout=1)`, then assert `call_count == 1` and `not lock.locked()`.

That `wait_for` wrapper is load-bearing rather than incidental. Every public lifecycle entry point
that touches one `app_key` must take that key's lock once for its whole sequence instead of
delegating to another public, lock-acquiring method partway through — doing the latter deadlocks
forever on the non-reentrant `asyncio.Lock`. The timeout is what converts that deadlock into a test
failure instead of a hung run, and it is exactly the kind of detail that silently drifts when it
lives in four places.

This extracts the shared shape into `assert_acquires_app_key_lock_once(lifecycle_service, app_key,
operation)` in `tests/unit/core/_fixtures_app_lifecycle.py`, alongside the existing lifecycle-test
helpers (`assert_load_completed_count`, `set_registry_apps`, `stub_detected_changes`), so no new
module was needed.

## What changed

- **New helper** in `_fixtures_app_lifecycle.py`, re-exported through `tests/unit/core/conftest.py`.
  Its docstring explains why the `wait_for` guard exists, so the rationale now lives in one place
  instead of being restated (or lost) per copy.
- **Four call sites migrated** — `reload_instance`, `stop_instance`, `start_instance`
  (`test_app_lifecycle_service_per_instance_ops.py`) and `reload_app`
  (`test_app_lifecycle_service_reload.py`). PMD/CPD flagged three of these; the fourth
  (`reload_instance`) was the same shape and was folded in too.
- **Per-operation mock setup stays in each test.** Only the lock assertion is shared, so a failure
  still points at the specific lifecycle method under test.
- **Fixture inventory updated** in `tests/unit/core/CLAUDE.md`.

Test-only refactor — no production code is touched.

## Out of scope

A related-but-distinct sibling pattern exists — the "second caller is queued and blocked on the
app-key lock" concurrency assertion that reaches into `asyncio.Lock._waiters`. It appears three
more times, but it asserts something different from the one flagged here and did not fall out
naturally from this extraction, so it is deliberately left alone per the issue's guidance.

## Testing

- `uv run nox -s dev` — **7741 passed**, 1 skipped, 2 xfailed
- `prek -a` — all 29 hooks pass
- Verified no inline copies of the extracted assertion remain: the only surviving
  `_get_app_key_lock` call sites are the three sibling-pattern assertions noted above.

Closes #2093

### DIFF
--- a/tests/unit/core/CLAUDE.md
+++ b/tests/unit/core/CLAUDE.md
@@ -17,6 +17,7 @@ Fixtures below are defined in family-scoped `_fixtures_*.py` modules in this dir
 ## Shared helpers and constants (module-level, not fixtures)
 
 - `assert_load_completed_count(event_capture, expected)` — assert how many `APP_LOAD_COMPLETED` broadcasts fired (the signal a connected dashboard refetches on)
+- `assert_acquires_app_key_lock_once(lifecycle_service, app_key, operation)` — await `operation()` and assert it took the per-app-key lock exactly once and released it; the internal `asyncio.wait_for` turns a non-reentrant-lock deadlock into a test failure instead of a hung run
 - `stub_detected_changes(lifecycle_service, changes)` — make `detect_changes` report `changes` and stub `apply_changes`, for tests of `handle_change_event`'s own apply/defer/broadcast decisions
 - `set_registry_apps(registry, apps)` — configures a `mock_registry`'s `__contains__`, `app_keys()`, `get_running_apps()`, and `get()` from an `apps`-shaped dict (`dict[str, dict[int, App]]`); use instead of assigning `mock_registry.apps = ...` directly (that attribute no longer exists on the real `AppRegistry`)
 - `make_manifest_obj(app_key, **kw)` — `SimpleNamespace` AppManifest stand-in for `AppRegistry` tests; shared by `test_app_registry.py` and `test_app_registry_snapshot.py`
--- a/tests/unit/core/_fixtures_app_lifecycle.py
+++ b/tests/unit/core/_fixtures_app_lifecycle.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import logging
+from collections.abc import Awaitable, Callable
 from pathlib import Path
 from typing import Any
 from unittest.mock import AsyncMock, MagicMock, Mock, patch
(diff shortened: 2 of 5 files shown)


NodeJSmith:
@coderabbitai review

PR #2164, opened 2026-09-09
test: tighten alert-banner assertions and drop redundant store overrides

### LINKED ISSUE #2095: Clean up storeOverrides noise and a weak assertion in alert-banner.test.tsx
Surfaced by the automated quality scanner (llm-checker, lazy-checker, nitpicker). Both the LLM-bias and deferred-debt passes returned clean; the two findings below come from the style/hygiene pass and were verified against the store, the render helper, and the component under test before filing. Pre-existing — not introduced by any in-flight branch.

### 1. `TelemetryDegradedBanner` tests seed `storeOverrides` values that are already the store defaults

All four tests in the `TelemetryDegradedBanner` block pass `droppedOverflow: 0, droppedExhausted: 0` (lines 76, 83, 90, 97), but `createInitialState` in `frontend/src/state/store.ts:152-154` already sets `telemetryDegraded: false, droppedOverflow: 0, droppedExhausted: 0`, and `frontend/src/test-setup.ts:77` resets the store to that initial state between every test. Three of the four tests therefore restate defaults they never vary, and the first test (`renders nothing when telemetryDegraded is false`, lines 75-77) restates *every* default, making its whole `storeOverrides` object a no-op.

This is also inconsistent with the sibling convention: `frontend/src/components/layout/status-bar.test.tsx` overrides only the field under test (`{ connection: "disconnected" }`, `{ sidebarCollapsed: true }`) and omits `storeOverrides` entirely when nothing varies.

**Proposed fix:** drop `droppedOverflow: 0, droppedExhausted: 0` from the three tests that do not vary them (lines 76, 83, 97), leaving those keys only in the drop-count test at line 90 where `10`/`5` are the point of the test. Keep `telemetryDegraded` explicit in all four, since that flag is the axis each test is exercising.

### 2. `does not render error span when error_message is null` asserts on the separator string, not on the span

Line 52 reads:

```ts
expect(container.textContent).not.toContain(" — ");
```

The assertion is coupled to the component's cosmetic separator rather than to the thing the test name claims to check. `alert-banner.tsx:34` renders `<span className="text-muted-foreground"> — {app.error_message}</span>`; if that separator is ever restyled (to `": "`, a different dash, or markup that puts the dash outside the text node), this test keeps passing while the actual regression — the error span rendering for a null `error_message` — goes undetected. It is a false-negative-prone assertion, not merely a magic string.

**Proposed fix:** assert the span's absence structurally instead. `AppLink` renders only an `<a>` (`frontend/src/components/shared/app-link.tsx`), so the error span is the only `span` inside the `li`:

```ts
expect(container.querySelector("li span")).toBeNull();
```

## Acceptance Criteria

- [ ] `droppedOverflow: 0` / `droppedExhausted: 0` removed from the `storeOverrides` at lines 76, 83, and 97; the drop-count test at line 90 keeps its non-default values.
- [ ] The `does not render error span when error_message is null` test asserts the absence of the error `span` node rather than the absence of the `" — "` substring.
- [ ] `npm run test` passes for `alert-banner.test.tsx` with no behavior change to `alert-banner.tsx`.

## Considered and not filed

The style pass also proposed extracting an `APP_KEY` constant for the repeated `"my_app"` literal, a `makeFailedApp()` factory for the inline `failedApps` arrays, named constants for the `10 + 5 = 15` drop-count arithmetic, and switching `../../test/...` imports to the `@/test/...` alias. These were assessed and dropped: `"my_app"` and `"/apps/my_app"` are self-contained fixture values with no referent elsewhere in the codebase (nothing to keep in sync); the distinct `10`/`5` inputs are what let the test prove both fields are summed, which named constants would obscure; and the relative test-helper import is the codebase majority convention (28 files use `../../test/`, 8 use `@/test/`), so changing one file would increase inconsistency rather than reduce it.

NodeJSmith:
## Summary

Two hygiene fixes in `frontend/src/components/layout/alert-banner.test.tsx`, both surfaced by the quality scanner's style/hygiene pass. Test-only change — `alert-banner.tsx` is untouched.

### 1. Removed `storeOverrides` values that only restate store defaults

Three of the four `TelemetryDegradedBanner` tests passed `droppedOverflow: 0, droppedExhausted: 0` without ever varying them. `createInitialState` (`frontend/src/state/store.ts:153-154`) already sets both to `0`, and `frontend/src/test-setup.ts:77` resets the store to that initial state in an `afterEach`, so the overrides were no-ops that obscured which field each test actually exercises.

They're now kept only in the drop-count test, where `10` / `5` are the point of the assertion. `telemetryDegraded` stays explicit in all four tests since that flag is the axis each one varies. This also brings the file in line with the sibling convention in `status-bar.test.tsx`, which overrides only the field under test.

### 2. Asserted the error span's absence structurally instead of by separator string

`does not render error span when error_message is null` asserted `container.textContent` did not contain `" — "` — the component's cosmetic separator, not the thing the test name claims to check. Restyling that separator (to `": "`, a different dash, or markup placing the dash outside the text node) would keep the test green while the real regression — the error span rendering for a null `error_message` — went undetected.

It now asserts `container.querySelector("li span")` is `null`. `AppLink` renders only a wouter `<Link>` (an `<a>`, no span), so the error span is the only `span` inside the `li`, making the query an exact structural check for the span's absence.

## Testing

- `npx vitest run src/components/layout/alert-banner.test.tsx` — 11 passed
- `uv run nox -s dev` — 7741 passed, 1 skipped, 2 xfailed
- `prek -a` — all hooks pass

Verified the two load-bearing assumptions directly rather than taking them from the issue: `app-link.tsx` renders a bare `Link`/`<a>` with no `span`, and `test-setup.ts` resets the store to `initialState()` between tests.

No visual evidence needed — `check_pr_screenshots.py` excludes `*.test.tsx` from the rendered-frontend gate, and no rendered files changed.

Closes #2095

### DIFF
--- a/frontend/src/components/layout/alert-banner.test.tsx
+++ b/frontend/src/components/layout/alert-banner.test.tsx
@@ -49,7 +49,7 @@ describe("AlertBanner", () => {
 
   it("does not render error span when error_message is null", () => {
     const { container } = render(<AlertBanner failedApps={[{ app_key: "my_app", error_message: null }]} />);
-    expect(container.textContent).not.toContain(" — ");
+    expect(container.querySelector("li span")).toBeNull();
   });
 
   it("renders all failed apps in the list", () => {
@@ -73,14 +73,14 @@ describe("AlertBanner", () => {
 describe("TelemetryDegradedBanner", () => {
   it("renders nothing when telemetryDegraded is false", () => {
     const { container } = renderWithAppState(<TelemetryDegradedBanner />, {
-      storeOverrides: { telemetryDegraded: false, droppedOverflow: 0, droppedExhausted: 0 },
+      storeOverrides: { telemetryDegraded: false },
     });
     expect(container.firstChild).toBeNull();
   });
 
   it("renders banner when telemetryDegraded is true", () => {
     const { getByTestId } = renderWithAppState(<TelemetryDegradedBanner />, {
-      storeOverrides: { telemetryDegraded: true, droppedOverflow: 0, droppedExhausted: 0 },
+      storeOverrides: { telemetryDegraded: true },
     });
     expect(getByTestId("telemetry-degraded-banner")).toBeDefined();
   });
@@ -94,7 +94,7 @@ describe("TelemetryDegradedBanner", () => {
 
   it("has role=alert for screen reader announcement", () => {
     const { container } = renderWithAppState(<TelemetryDegradedBanner />, {
-      storeOverrides: { telemetryDegraded: true, droppedOverflow: 0, droppedExhausted: 0 },
+      storeOverrides: { telemetryDegraded: true },
     });
     const banner = container.querySelector("[role='alert']");
     expect(banner).not.toBeNull();


NodeJSmith:
@coderabbitai review

PR #2170, opened 2026-09-09
refactor: render TelemetryDegradedBanner through the shared AlertShell

### LINKED ISSUE #2096: Clean up alert-banner.tsx: reuse AlertShell and drop redundant styling
Surfaced by the automated quality scanner (llm-checker, lazy-checker, nitpicker). All findings are pre-existing and confined to `frontend/src/components/layout/alert-banner.tsx` plus one call into the existing shared `AlertShell`.

## Finding 1 — `TelemetryDegradedBanner` hand-rolls the shared `AlertShell` container

`alert-banner.tsx:57-61` builds its container inline:

```
mb-4 flex items-center gap-3 rounded-md border border-[var(--status-warning)] bg-[var(--status-warning-bg)] px-4 py-3 ...
```

`components/shared/alert-shell.tsx:38` renders exactly `mb-4 rounded-md border px-4 py-3` plus, for `tone="warning"`, `border-[var(--status-warning)] bg-[var(--status-warning-bg)]` — a character-for-character match on geometry and tone. `AlertShell` is already the established pattern at four call sites (`diagnostics/telemetry-panel.tsx`, `diagnostics/logging-panel.tsx`, `shared/error-banner.tsx`, `app-detail/app-detail-header.tsx`), including `telemetry-panel.tsx:24`, which renders the *same* telemetry-degraded concept with `<AlertShell tone="warning" role="alert" ...>`.

`AlertShell`'s own doc comment deliberately calls out `AlertBanner` as a non-fit (different geometry: `mx-8 mb-2 px-3 py-2`, full-strength `border-destructive`), so `AlertBanner` staying hand-rolled is intentional and should be left alone. `TelemetryDegradedBanner` is not covered by that carve-out and matches the shell exactly.

**Fix:** render `TelemetryDegradedBanner`'s container as `<AlertShell tone="warning" role="alert" data-testid="telemetry-degraded-banner" className="flex items-center gap-3 text-[var(--status-warning)]">`, following the `telemetry-panel.tsx` / `PANEL_BANNER_CLASS` pattern where content-level classes ride on `className` and `cn()`'s twMerge handles overrides. The existing tests in `alert-banner.test.tsx` key off `data-testid`, which `AlertShell` accepts as a prop.

## Finding 2 — `text-[length:var(--text-body)]` is redundant, twice

`alert-banner.tsx:58` sets `text-[length:var(--text-body)]` on the container and `:63` sets it again on the child `<span>`, which would inherit it anyway. Both are also redundant against `global.css:427`, where `--text-body` is already the base body `font-size`.

**Fix:** drop `text-[length:var(--text-body)]` from both lines. Keep `leading-[var(--text-body-leading)]` on the span only if the surrounding line-height actually differs from the inherited default; otherwise drop it too.

## Finding 3 — `cn()` called with a single static string

`alert-banner.tsx:21-23` wraps one static class literal in `cn(...)` with no conditional or dynamic segment. `cn()` exists for merging conditional class lists; here it is pure indirection.

**Fix:** use a plain `className="..."` string.

## Finding 4 — JSDoc on `TelemetryDegradedBanner` restates the code

`alert-banner.tsx:42-46` is a three-line block whose second sentence — "Reads `telemetryDegraded`, `droppedOverflow`, and `droppedExhausted` signals" — enumerates the three `useAppStore` calls on the very next lines, and will silently drift if a signal is added or removed. The first sentence largely restates the component name plus its one-line guard.

The only non-obvious content is the parenthetical explaining what upstream conditions "degraded" covers (queue overflow, backpressure, unreachable), which is not derivable from the component body.

**Fix:** collapse to a single line keeping only the non-obvious part, e.g. `/** Warns when telemetry is degraded — queue overflow, backpressure, or an unreachable database. */`

## Assessed and excluded

These were raised by the checkers but do not survive validity assessment:

- **"Mixed semantic Tailwind tokens vs. raw CSS-var classes"** (`border-destructive` in `AlertBanner` vs. `border-[var(--status-warning)]` in `TelemetryDegradedBanner`, and `border-destructive` alongside `bg-[var(--destructive-bg)]` on the same element). `global.css:29` registers `--color-destructive` in `@theme`, which is what generates the `*-destructive` utilities; there is no `--color-*` registration for warning, `--destructive-bg`, or `--destructive-border`. The arbitrary-value form is therefore the only available spelling for those roles, not a style choice.
- **"Import paths should use the `@/` alias consistently."** Relative imports are the house majority under `frontend/src/components` (233 relative vs. 109 aliased); `@/lib/utils` for `cn` is the shadcn convention used repo-wide. The file matches existing practice.
- **"Two unrelated components in one file / filename names only one of them."** Both exports are layout-shell banners mounted on adjacent lines in `app.tsx:250-251`. That is genuine cohesion; the singular filename is a naming preference with no defect behind it.
- **JSX attribute ordering (`role` before `data-testid` in one component, after in the other).** No lint rule enforces attribute order and there is no behavioral consequence.

## Acceptance Criteria

- [ ] `TelemetryDegradedBanner` renders through `AlertShell tone="warning"` instead of a hand-rolled container; `AlertBanner` is left hand-rolled per `AlertShell`'s documented carve-out
- [ ] Redundant `text-[length:var(--text-body)]` removed from both lines
- [ ] `cn()` single-static-argument call replaced with a plain string
- [ ] `TelemetryDegradedBanner` doc comment reduced to the non-obvious degradation causes
- [ ] `alert-banner.test.tsx` and `app.test.tsx` still pass without modification

NodeJSmith:
## Summary

`TelemetryDegradedBanner` hand-rolled a container whose geometry and warning-tone classes matched the shared `AlertShell` character-for-character. `AlertShell`'s doc comment carves out `AlertBanner` specifically (different geometry, full-strength `border-destructive`), but that carve-out never covered `TelemetryDegradedBanner` — so it now renders through the shell like the four existing call sites, including `diagnostics/telemetry-panel.tsx`, which already renders the same telemetry-degraded concept that way.

## What changed

All in `frontend/src/components/layout/alert-banner.tsx`:

- **`TelemetryDegradedBanner` renders through `AlertShell tone="warning"`.** Container geometry and tone classes come from the shell; content-level classes (`flex items-center gap-3 text-[var(--status-warning)]`) ride on `className`, following the `telemetry-panel.tsx` pattern. `AlertBanner` is deliberately left hand-rolled per the documented carve-out.
- **Dropped redundant `text-[length:var(--text-body)]`** from both the container and its child `<span>`. `--text-body` is already the base body `font-size` in `global.css`, and the span would have inherited it from the container regardless. `leading-[var(--text-body-leading)]` is kept on the span.
- **Replaced the single-static-argument `cn()` call** in `AlertBanner` with a plain string — no conditional or dynamic segment was being merged.
- **Collapsed the `TelemetryDegradedBanner` doc comment** to the one non-obvious part (what upstream conditions "degraded" covers). The removed sentence enumerated the three `useAppStore` calls on the next lines and would drift silently as signals change.

No behavior change is intended: this is a pure refactor.

## Testing

- Compared the rendered container class set against the old literal — identical apart from the deliberately removed font-size utility.
- `uv run nox -s dev` — 7741 passed, 1 skipped, 2 xfailed.
- `cd frontend && npm run test` — 1604 passed across 110 files. `alert-banner.test.tsx` and `app.test.tsx` pass unmodified; they key off `data-testid`, which `AlertShell` accepts as a prop.
- `prek -a` and `prek pyright -a --stage pre-push` — clean.
- Captured the degraded state against the demo stack. The healthy stack has no organic trigger for this banner, so the screenshots below are the visual evidence that the shell swap renders identically.

## Screenshots

Desktop, light (1280):

Telemetry degraded banner, desktop light

Desktop, dark (1280):

Telemetry degraded banner, desktop dark

Mobile, dark (375):

Telemetry degraded banner, mobile dark

Screenshots live under `docs/pr-evidence/`, which is unignored in `.gitignore` and excluded from the mkdocs build so it never renders as a docs page.

Closes #2096


NodeJSmith:
@coderabbitai review

NodeJSmith:
Automated review pass: the frontend `build`/`frontend`/`e2e` CI failures on this PR are not caused by its diff. They come from a pre-existing `tsc` break on `main` (`stat-cell-builders.test.ts` imports the now-unexported `COMMON_STAT_CELL_COUNT`), tracked in #2177. Auto-merge is queued; this PR will need a re-run or branch update once #2177 lands on `main`.

NodeJSmith:
Second automated review pass (CodeRabbit triage + code/integration/WTF + llm/lazy/nitpicker battery). CodeRabbit's check came back clean with zero threads, so there was nothing to triage. Three findings were valid and are fixed in e8e913f8:

- **The `AlertBanner` carve-out comment was inaccurate** (raised independently by the WTF and lazy reviewers). It justified staying hand-rolled on container geometry, but `AlertShell`'s own docstring — updated in this same PR — documents geometry as overridable through `className`, and `PANEL_BANNER_CLASS` already does exactly that. The difference the shell genuinely cannot express is the border token: `AlertBanner` uses `border-destructive` (full-strength `--err`) while the `danger` tone uses `--destructive-border` (`--err` at 30% alpha), and only `tone` selects it. The comment now states that.
- **Nothing pinned the twMerge override contract** the expanded docstring promises — the existing test only asserted a non-conflicting class, which would pass under naive concatenation. Added a test asserting `mb-0 rounded-sm` actually displaces the shell's `mb-4 rounded-md`.
- **Prop order** on the new `<AlertShell>` call now matches the other call sites (`tone, className, role, data-testid`).

Skipped, with reasons: the dropped `text-[length:var(--text-body)]` is a verified no-op (`global.css` sets it on `body`, and no ancestor between `<body>` and the banner overrides `font-size`) and was a deliberate, documented decision in this PR; the `docs/pr-evidence/` "new convention" flag is a stale-branch three-dot-diff artifact — the `.gitignore` and `mkdocs.yml` entries are already on `main` alongside 11 existing evidence PNGs.

Verification: 57 frontend tests across the three affected files, `tsc --noEmit`, eslint, prettier, and the full prek hook set all clean.

CI status is unchanged from the previous pass — the `frontend`/`build`/`e2e` failures are still the pre-existing `tsc` break on `main` (#2177), confirmed again by reading `origin/main`'s `stat-cell-builders.ts`, which still declares `COMMON_STAT_CELL_COUNT` without exporting it while its test imports it. This branch type-checks clean on its own; the failure only appears in CI's merge commit. #2177 is still open, so auto-merge (squash) stays queued and this PR needs a re-run or branch update once that lands. `file-sizes` and `duplicate-code` are the known informational baseline failures.

NodeJSmith:

PR #2181, opened 2026-09-10
refactor(ui): deduplicate live-status and in_current_config explanations in command palette items

### LINKED ISSUE #2103: Deduplicate repeated live-status and in_current_config explanations in palette-items.ts
Surfaced by an automated quality scan of `frontend/src/components/layout/palette-items.ts`. Both findings are duplication of a rule/invariant that already has a single authoritative home — neither changes behavior.

## Finding 1: The live-status staleness invariant is re-explained at four call sites

`appLiveStatus` already carries a ~19-line doc comment at its definition (`frontend/src/utils/app-data.ts:100-118`) covering exactly the overlay-vs-cached-manifest staleness semantics, and `instanceLiveStatus` has its own (`app-data.ts:148-151`). `palette-items.ts` then re-explains the same invariant four more times at the consumer side:

- `palette-items.ts:59-62` — a 4-line block explaining the internal Zustand/`useManifests()` invalidation mechanism to justify one `filter` call.
- `palette-items.ts:73-75` — restates the staleness rationale ("Same live-status derivation as reload-all above") after a legitimate note distinguishing this predicate from `isReloadableStatus`.
- `palette-items.ts:109` — "same staleness reasoning as buildActionItems above".
- `palette-items.ts:123` — "same reasoning as the app row above".

Three of the four are explicitly cross-referential, which is the tell that the explanation belongs at the shared boundary rather than at each consumer. `rules/common/reader-load.md` states this directly: "Name the invariant at the boundary, not in every consumer, so the reader learns it once." Here the boundary *already* names it, so the four consumer copies are pure duplication that will drift if the mechanism changes.

**Proposed fix:** Delete the three cross-referential repeats (`:73-75` staleness sentence only — keep the `isReloadableStatus` distinction, which is genuinely local and not documented elsewhere; `:109`; `:123`). Reduce `:59-62` to a single line pointing at the helper, e.g. `// Live status, not m.status — see appLiveStatus for why the cached manifest can be stale.` The Zustand/`useManifests()` mechanism detail is implementation trivia that isn't needed to trust the one-line call and is better placed with the helper if it's wanted at all.

## Finding 2: The `in_current_config` filter and its rationale are duplicated

The same business rule — exclude apps not in the current config from actionable palette results — is encoded independently in two functions, each with its own 2-line explanatory comment:

- `palette-items.ts:50-52` (`buildActionItems`)
- `palette-items.ts:100-102` (`buildAppItems`)

The comment at `:50-51` even acknowledges the coupling ("exclude them the same way buildAppItems does"). A future change to the rule (e.g. adding a second exclusion condition) has to be found and applied in both places, with no compiler help if one is missed.

**Proposed fix:** Extract a single module-local helper, e.g. `function activeManifests(manifests: AppManifest[]) { return manifests.filter((m) => m.in_current_config); }`, carrying the rationale comment once. `buildActionItems` calls it directly; `buildAppItems` calls it and then sorts.

## Not filed (assessed and dropped)

For the record, these were raised by the scanners and rejected with evidence:

- **`id` prefix duplicates the `kind` value** (e.g. `` id: `app-${m.app_key}` `` next to `kind: "app"`). Claimed consequence was a silent desync of the id namespace from `kind`. In practice `item.id` is only ever used as a React `key` and a `data-testid` suffix (`app.tsx:390,392`); grouping and filtering read `item.kind` directly (`app.tsx:358-360`), and nothing parses the id prefix. No desync is observable.
- **Short locals `m`, `l`, `inst`, and the `sub` field abbreviation.** `rules/common/coding-style.md` explicitly endorses short, contextually obvious names inside functions; these are all unambiguous in their local scope.
- **Separate `import type` / value import from `../../api/endpoints`.** Idiomatic TypeScript and the file's consistent pattern.
- **`DOCS_URL` as a hard-coded literal.** A documentation site link, not environment-varying config, and already a named top-level constant.
- **`../../` relative imports vs. the `@/` alias.** Real convention drift, but repo-wide (74 files use `../..`, 81 use `@/`, both styles coexist inside `components/layout/` and even within single files), so it isn't a defect in this file and can't be resolved by editing it.

## Acceptance Criteria

- [ ] The live-status staleness rationale appears once (at the `appLiveStatus`/`instanceLiveStatus` definitions), not at each `palette-items.ts` call site.
- [ ] The `isReloadableStatus`-vs-failure-status distinction at `:73-74` is preserved — only the repeated staleness sentence is removed.
- [ ] `m.in_current_config` filtering is expressed once in `palette-items.ts` and reused by both builders.
- [ ] `frontend/src/components/layout/palette-items.test.ts` still passes unchanged (no behavior change intended).

NodeJSmith:
## Summary

Removes duplicated explanatory comments in the command palette's item builders (`frontend/src/components/layout/palette-items.ts`). No behavior change.

- **`in_current_config` filter:** `buildActionItems` and `buildAppItems` each repeated the same `manifests.filter((m) => m.in_current_config)` along with their own copy of why removed apps are excluded. That now lives in one module-local `activeManifests()` helper, with a single doc comment covering both reasons (stop/reload against an unloaded app would 404 or no-op, and "jump to…" only lists apps a user can act on).
- **Live-status staleness:** four inline comments repeated why the palette reads live WS status instead of the cached `m.status`. `appLiveStatus` (`frontend/src/utils/app-data.ts`) already documents that invariant in full, and `instanceLiveStatus` documents its own overlay. The reload-all action keeps a one-line pointer to `appLiveStatus`; the other restatements are deleted.

`activeManifests()` returns a fresh array from `filter`, so the `.sort()` in `buildAppItems` still never mutates the caller's `manifests` (the old code copied with `[...manifests]` first).

## Testing

- `uv run nox -s dev`: 7758 passed, 1 skipped, 2 xfailed, 0 failures
- `npx vitest run src/components/layout`: 7 files, 107 tests passed (includes `palette-items.test.ts`)
- `prek -a`: all hooks pass except `tsc (frontend)`; see below

## Why this is a draft

`prek -a` (and the pre-push hook, so the branch was pushed with `--no-verify`) fails on one `tsc` error in a file this PR doesn't touch:

```
src/components/app-detail/stat-cell-builders.test.ts(4,32): error TS2459: Module '"./stat-cell-builders"' declares 'COMMON_STAT_CELL_COUNT' locally, but it is not exported.
```

This error is on `main`. This branch is based exactly on the freshly fetched `origin/main` tip, has no diff under `frontend/src/components/app-detail/`, and on `origin/main` `stat-cell-builders.ts` declares `const COMMON_STAT_CELL_COUNT = 5;` without exporting it, while the test imports it. The fix belongs in its own change (export the constant, or stop importing it in the test), not in this cleanup PR.

Partial progress on #2103 — needs human review

Closes #2103

### DIFF
--- a/frontend/src/components/layout/palette-items.ts
+++ b/frontend/src/components/layout/palette-items.ts
@@ -42,24 +42,26 @@ export function buildStaticPageItems(navigate: (path: string) => void): PaletteI
   }));
 }
 
+/** Removed apps (in_current_config: false) are historical/DB-only — the runtime doesn't know
+ *  about them, so they're excluded from every actionable palette result (stop/reload would 404
+ *  or no-op, and "jump to…" is for apps a user can actually act on). */
+function activeManifests(manifests: AppManifest[]): AppManifest[] {
+  return manifests.filter((m) => m.in_current_config);
+}
+
 export function buildActionItems(
   manifests: AppManifest[],
   appStatuses: Record<string, AppStatusEntry>,
   onClose: () => void,
 ): PaletteItem[] {
-  // Removed apps (in_current_config: false) aren't loaded — stop/reload would 404 or no-op
-  // against an app the runtime doesn't know about, so exclude them the same way buildAppItems does.
-  const active = manifests.filter((m) => m.in_current_config);
+  const active = activeManifests(manifests);
   return [
     {
       id: "action-reload-all",
       kind: "action",
       label: "Reload all apps",
       action: () => {
-        // Selection is derived from live per-instance WS status, not the cached manifest's
-        // m.status — app_status_changed updates only the Zustand status map and does not
-        // invalidate useManifests(), so a manifest can go stale (e.g. still "running" after an
-        // instance fails) for as long as no execution event happens to refetch the palette data.
+        // Live status, not m.status — see appLiveStatus for why the cached manifest can be stale.
         const reloadable = active.filter((m) => isReloadableStatus(appLiveStatus(appStatuses, m)));
         void Promise.allSettled(reloadable.map((m) => reloadApp(m.app_key)));
         onClose();
@@ -72,7 +74,6 @@ export function buildActionItems(
       action: () => {
         // Not isReloadableStatus's stop-side counterpart — this targets apps recovery should
         // stop (any failure status), not "is stop meaningful for this app" (which running is too).
-        // Same live-status derivation as reload-all above, for the same reason.
         const failing = active.filter((m) => isFailureStatus(appLiveStatus(appStatuses, m)));
         void Promise.allSettled(failing.map((m) => stopApp(m.app_key)));
         onClose();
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #2183, opened 2026-09-10
test: extract bus-derived listener and registry-key helpers in test_if_exists

### LINKED ISSUE #2110: Extract the bus-derived listener and registry-key setup in test_if_exists.py
## Description

`tests/unit/bus/test_if_exists.py` repeats the same "build a listener whose identity fields come off the `bus` fixture" block: `create_listener(handler=..., app_key=bus.parent.app_key, instance_index=bus.parent.index, name="test_listener", owner_id=bus.owner_id)`. PMD's `duplicate-code` job flags it across 3 locations; every copy is byte-identical except the local variable it is assigned to.

A second, closely related shape rides along in the same file: the registry-key tuple `key = (bus.parent.app_key, bus.parent.index, "my_listener", "test.topic")`, copied verbatim in 5 tests. It is the same idea — re-deriving the bus fixture's identity coordinates by hand at each call site — and it should be collapsed in the same pass.

Neither is a deliberate exception. This directory already has the convention and the vocabulary for exactly this: `tests/unit/bus/CLAUDE.md` documents a "File-local helpers" section listing module-scope helpers in `test_invocation.py`, `test_handler_invoker.py`, `test_duration_hold.py`, and `test_duration_timer.py` that each collapse a setup pattern repeating within one file, per `.claude/rules/test-conventions.md` (shared placement in `tests/support/` is reserved for patterns reused across 3+ files). This shape only appears in `test_if_exists.py`, so a file-local helper is the right home — not a new shared factory.

## Goal

Extract the shared bus-derived listener construction into one module-scope helper in `test_if_exists.py` (e.g. `make_bus_listener(bus, *, handler=handler_a, name="test_listener")`) and use it at every call site with this shape, so each test only expresses what is actually distinct about it. Do the same for the repeated registry-key tuple (e.g. `listener_key(bus, name="my_listener", topic="test.topic")`).

Then add both helpers to the "File-local helpers" list in `tests/unit/bus/CLAUDE.md`, matching how the existing entries there are described — that list is the discovery path for these, since a reader scanning `conftest.py` alone won't find them.

One thing to confirm while doing it: `create_listener` derives `identity.handler_name` / `short_name` from the handler object, so keep the handler a parameter of the helper rather than hardcoding `handler_a`, and check nothing in the file asserts on those derived names before and after.

## Where

PMD flagged these locations as one block:

- `tests/unit/bus/test_if_exists.py:439-445`
- `tests/unit/bus/test_if_exists.py:455-461`
- `tests/unit/bus/test_if_exists.py:464-470`

Other copies of the same shape found while reading, also in scope:

- `tests/unit/bus/test_if_exists.py:96`, `104`, `125`, `148`, `262` — the repeated `key = (bus.parent.app_key, bus.parent.index, "my_listener", "test.topic")` registry-key tuple

The flagged line ranges are a sample, not a boundary. Any other instance of the same shape found while doing the work is in scope too.

## Acceptance Criteria

- [ ] A module-scope helper in `test_if_exists.py` covers the bus-derived `create_listener(...)` construction, and all three call sites use it
- [ ] A module-scope helper covers the registry-key tuple, and all five call sites use it
- [ ] Both helpers are listed under "File-local helpers" in `tests/unit/bus/CLAUDE.md`
- [ ] `uv run pytest tests/unit/bus/test_if_exists.py` passes unchanged

NodeJSmith:
## Summary

`tests/unit/bus/test_if_exists.py` rebuilt the `bus` fixture's identity coordinates by hand in two repeated shapes. PMD's `duplicate-code` job flagged one of them. This PR moves each shape into a file-local helper:

- **`make_bus_listener(bus, *, handler=handler_a, name="test_listener")`** replaces three identical `create_listener(handler=..., app_key=bus.parent.app_key, instance_index=bus.parent.index, name=..., owner_id=bus.owner_id)` blocks in the `bus.add_listener(...)` tests. `handler` stays a parameter because `create_listener()` derives the listener's `handler_name`/`short_name` from it. No test in the file asserts on those derived names.
- **`listener_key(bus, *, name="my_listener", topic="test.topic")`** replaces five copies of the `(app_key, index, name, topic)` tuple that `bus._registered_listeners` uses as its key.

Both helpers are listed under "File-local helpers" in `tests/unit/bus/CLAUDE.md`, next to the existing entries for `test_duration_hold.py`, `test_router.py`, and the others. The helpers stay file-local because this shape only appears in `test_if_exists.py`. `.claude/rules/test-conventions.md` reserves `tests/support/` for shapes used in 3+ files.

This PR changes only tests and the test directory's CLAUDE.md. Runtime behavior is unchanged.

## Testing

- `uv run pytest tests/unit/bus/test_if_exists.py`: 26 passed
- `uv run pytest tests/unit/bus -n 4`: 697 passed
- `uv run pyright tests/unit/bus/test_if_exists.py`: 0 errors
- `uv run ruff check` / `ruff format --check`: clean
- `tools/check_test_factories.py`: no local factory shadows a shared factory name

The `tsc (frontend)` failure that originally held this PR in draft was a pre-existing `main` issue (`COMMON_STAT_CELL_COUNT` not exported from `stat-cell-builders.ts`), unrelated to this branch. It was fixed on main by #2189, which is merged in here.

The `duplicate-code` and `file-sizes` lint jobs fail on `main` as well and run with `continue-on-error: true` — expected baseline noise, not introduced by this PR.

Closes #2110


NodeJSmith:
@coderabbitai review

Issue #2177, opened 2026-09-10
Fix tsc failure on main from unexported COMMON_STAT_CELL_COUNT


NodeJSmith:
## Description

`main` currently fails `tsc --noEmit`, which fails the `frontend` job in `tests.yml` and the
`npm run build` step in `lint.yml`'s `frontend` job. Every open PR that touches frontend paths
is blocked behind it.

```
src/components/app-detail/stat-cell-builders.test.ts(4,32): error TS2459:
Module '"./stat-cell-builders"' declares 'COMMON_STAT_CELL_COUNT' locally, but it is not exported.
```

Reproduce on a clean checkout of `main`:

```bash
cd frontend && npm install && npm run typecheck
```

Latest `tests.yml` run on `main` (`84f392d6`) is a **failure** for this reason:
https://github.com/NodeJSmith/hassette/actions/runs/34406031457

## Cause

A semantic merge conflict between two PRs that were each green on their own base:

- **#2139** (`8f5e9f5c`) changed `stat-cell-builders.test.ts` to import `COMMON_STAT_CELL_COUNT`
  from `stat-cell-builders.ts` as an exported constant.
- **#2142** (`3b25fcce`) rewrote `stat-cell-builders.ts` and left `COMMON_STAT_CELL_COUNT`
  declared as a module-local `const` with no `export`.

Neither branch contained the other's change, so neither CI run saw the broken import. Git
merged both cleanly because they touch different files.

## Fix

One word, in `frontend/src/components/app-detail/stat-cell-builders.ts:5`:

```diff
-const COMMON_STAT_CELL_COUNT = 5;
+export const COMMON_STAT_CELL_COUNT = 5;
```

The test already imports it under that exact name, and `getCommonStatCellCount()` (line 39)
continues to return it, so no other call site changes.

## Acceptance Criteria

- [ ] `cd frontend && npm run typecheck` exits 0 on `main`
- [ ] The `frontend` job in `tests.yml` passes on `main`
- [ ] `stat-cell-builders.test.ts` still imports the constant rather than re-declaring a local
      copy of the magic number (the point of #2139)

NodeJSmith:
Autofix found nothing to change: COMMON_STAT_CELL_COUNT is already exported on main by #2189 (89a3c5f3), and `npm run typecheck` exits 0

PR #2185, opened 2026-09-10
test: parametrize clock-patched session cookie TTL and renewal tests

### LINKED ISSUE #2118: Parametrize the clock-patched session cookie TTL and renewal tests
## Description

`tools/check_duplicate_code.py` (PMD CPD) flagged the TTL cases in `TestSessionCookieTtl`
(`tests/unit/web/test_auth_session.py`) as a duplicated block. Reading the code, the repeated
shape is **"mint a session cookie under a patched clock at a fixed epoch, then call a session
predicate under the same clock patched forward by N seconds, and assert the outcome."**

The three TTL tests are byte-identical apart from two values: the elapsed offset (`+1000`,
`+3601`, `+3600`) and the expected result (`issued_at == 1_000_000` twice, `is None` once). Each
one repeats the full four-line `patch("hassette.web.auth.session._current_timestamp", ...)`
mint-then-verify scaffold, so the module path string and the `1_000_000` base epoch are written
out six times between them.

The sibling `TestShouldRenewSessionCookie` class (same file, lines 238-265) is the same shape with
the mint half dropped — five tests that each patch `_current_timestamp` to `1_000_000 + N` and
assert `should_renew_session_cookie(issued_at=1_000_000, session_ttl=3600)` is `True`/`False`,
differing only in `N` and the expected boolean. PMD didn't flag it (the bodies are short enough to
fall under the token threshold), but it is the same duplication and belongs in the same fix.

`pytest.mark.parametrize` is already the convention for case tables in this directory
(`test_auth_tokens.py`, `test_dependencies.py`, `test_ws_helpers.py`, `test_telemetry_helpers.py`).

## Goal

Extract the shared "run a session predicate under a clock frozen at `base + elapsed`" scaffold
into one helper (a small local helper or a shared fixture — see `.claude/rules/test-conventions.md`
for where shared test helpers live) and drive the varying cases through
`pytest.mark.parametrize` at every call site with this shape, so each case is a
`(elapsed, expected)` row rather than a copy of the whole clock-patching body. The base epoch and
the `_current_timestamp` patch target should be written once.

Preserve the coverage exactly. Every existing case must survive as its own parametrized case with
a readable id — in particular the three distinct TTL outcomes (inside the window, exactly at the
`session_ttl` boundary, one second past) and the three distinct renewal outcomes (just before
half-life, exactly at half-life, past full TTL). Collapsing or dropping any boundary case defeats
the point of these tests: the boundary behavior is the thing under test, and a parametrization
that quietly loses one stays green. The explanatory comments and docstrings that justify specific
offsets (e.g. the half-life arithmetic note, the "past full TTL" rationale) should carry over to
the parametrize table or the test body rather than being dropped.

## Where

Flagged by the scanner:

- `tests/unit/web/test_auth_session.py:156-161` — `test_cookie_within_ttl_is_accepted`
- `tests/unit/web/test_auth_session.py:166-171` — `test_cookie_past_ttl_is_rejected`
- `tests/unit/web/test_auth_session.py:176-180` — `test_cookie_exactly_at_ttl_boundary_is_accepted`

Found while reading, same shape, also in scope:

- `tests/unit/web/test_auth_session.py:238-265` — the five `TestShouldRenewSessionCookie` tests

These locations are supporting evidence, not a boundary. Any other instance of the same
clock-patched mint/verify shape found while doing the work is in scope too.

## Acceptance Criteria

- [ ] The `_current_timestamp` patch target and the `1_000_000` base epoch appear once, not once
      per test.
- [ ] `TestSessionCookieTtl`'s three cases are parametrized rows over `(elapsed, expected)`.
- [ ] `TestShouldRenewSessionCookie`'s five cases are parametrized the same way.
- [ ] Every boundary case that exists today still exists as a distinct case with a readable id.
- [ ] `uv run pytest tests/unit/web/test_auth_session.py` passes with the same number of collected
      cases as before.

NodeJSmith:
## Summary

CPD flagged `TestSessionCookieTtl` in `tests/unit/web/test_auth_session.py` as duplicated code. Its three TTL tests were identical except for the elapsed offset and the expected result. `TestShouldRenewSessionCookie` had the same pattern across five tests. Each test repeated the full `patch("hassette.web.auth.session._current_timestamp", ...)` setup, which wrote the patch target and the `1_000_000` base epoch out again every time.

This PR changes the following:

- Adds `CLOCK_PATCH_TARGET` and `BASE_EPOCH` module constants and a `frozen_clock(elapsed)` context manager, so the patch target and epoch appear once.
- Merges the TTL tests into one `test_cookie_ttl_enforcement`, parametrized over `(elapsed, expected_issued_at)`.
- Merges the renewal tests into one `test_renewal_window`, parametrized over `(elapsed, expected)`.

No boundary case was dropped or merged. Each one is its own `pytest.param` with a readable id:

| Class | Cases |
|---|---|
| `TestSessionCookieTtl` | `within_ttl_accepted`, `exactly_at_ttl_boundary_accepted`, `past_ttl_rejected` |
| `TestShouldRenewSessionCookie` | `freshly_minted_not_renewed`, `just_before_half_life_not_renewed`, `exactly_at_half_life_renewed`, `past_half_life_renewed`, `past_full_ttl_not_renewed` |

The comments explaining the 1000s/3601s offsets, the half-life math, and why `should_renew_session_cookie` has its own upper bound are still there, now next to the matching param rows.

The change is test-only. No production code changed.

## Testing

- `uv run nox -s dev`: 7758 passed, 1 skipped, 2 xfailed.
- `pytest --collect-only` on the two refactored classes collects all 8 cases under the ids above, and all 8 pass.
- `prek run pyright -a --stage pre-push`: passed.
- `prek -a`: every Python hook passed. The `tsc (frontend)` hook fails, but not because of this PR. It fails the same way on the current `main`: `frontend/src/components/app-detail/stat-cell-builders.test.ts` imports `COMMON_STAT_CELL_COUNT`, and `stat-cell-builders.ts` doesn't export it. This branch doesn't touch any frontend files.

Closes #2118

### DIFF
--- a/tests/unit/web/test_auth_session.py
+++ b/tests/unit/web/test_auth_session.py
@@ -24,6 +24,13 @@ CLOCK_PATCH_TARGET = "hassette.web.auth.session._current_timestamp"
 BASE_EPOCH = 1_000_000
 
 
+@contextmanager
+def frozen_clock(elapsed: int) -> Iterator[None]:
+    """Freeze the session module's clock at ``BASE_EPOCH + elapsed`` whole unix seconds."""
+    with patch(CLOCK_PATCH_TARGET, return_value=BASE_EPOCH + elapsed):
+        yield
+
+
 class TestCheckBearerToken:
     def test_correct_token_succeeds(self) -> None:
         assert check_bearer_token("the-real-token", "the-real-token") is True
@@ -252,10 +259,3 @@ class TestShouldRenewSessionCookie:
     def test_renewal_window(self, elapsed: int, expected: bool) -> None:
         with frozen_clock(elapsed):
             assert should_renew_session_cookie(issued_at=BASE_EPOCH, session_ttl=3600) is expected
-
-
-@contextmanager
-def frozen_clock(elapsed: int) -> Iterator[None]:
-    """Freeze the session module's clock at ``BASE_EPOCH + elapsed`` whole unix seconds."""
-    with patch(CLOCK_PATCH_TARGET, return_value=BASE_EPOCH + elapsed):
-        yield
--- a/tests/unit/web/test_auth_session.py
+++ b/tests/unit/web/test_auth_session.py
@@ -22,6 +22,8 @@ from hassette.web.auth.trusted_proxies import resolve_trusted_proxies
 
 CLOCK_PATCH_TARGET = "hassette.web.auth.session._current_timestamp"
 BASE_EPOCH = 1_000_000
+"""Unix timestamp every test treats as t=0, the moment a cookie under test was minted."""
+
 SESSION_TTL = 3600
 """Session lifetime every test in this file mints and verifies against."""
 
@@ -175,7 +177,7 @@ class TestSessionCookieTtl:
         ("elapsed", "expected_issued_at"),
         [
             pytest.param(1000, BASE_EPOCH, id="within_ttl_accepted"),
-            # The TTL bound is inclusive - expiry is `elapsed > session_ttl`, not `>=`.
+            # The TTL bound is inclusive — expiry is `elapsed > session_ttl`, not `>=`.
             pytest.param(SESSION_TTL, BASE_EPOCH, id="exactly_at_ttl_boundary_accepted"),
             pytest.param(SESSION_TTL + 1, None, id="past_ttl_rejected"),
         ],
--- a/tests/unit/web/test_auth_session.py
+++ b/tests/unit/web/test_auth_session.py
@@ -22,6 +22,11 @@ from hassette.web.auth.trusted_proxies import resolve_trusted_proxies
 
 CLOCK_PATCH_TARGET = "hassette.web.auth.session._current_timestamp"
 BASE_EPOCH = 1_000_000
+SESSION_TTL = 3600
+"""Session lifetime every test in this file mints and verifies against."""
+
+HALF_LIFE = SESSION_TTL // 2
+"""Age at which ``should_renew_session_cookie`` starts renewing, per its half-life rule."""
 
 
 @contextmanager
@@ -97,7 +102,7 @@ class TestSessionCookieMintAndVerify:
         token = "the-real-token"
 
         cookie_value = mint_session_cookie(token)
-        issued_at = verify_session_cookie(cookie_value, token, session_ttl=3600)
+        issued_at = verify_session_cookie(cookie_value, token, session_ttl=SESSION_TTL)
 
         assert issued_at is not None
 
@@ -113,15 +118,15 @@ class TestSessionCookieMintAndVerify:
         replayed_cookie_value = str(cookie_value)
         replayed_token = str(token)
 
-        assert verify_session_cookie(replayed_cookie_value, replayed_token, session_ttl=3600) is not None
+        assert verify_session_cookie(replayed_cookie_value, replayed_token, session_ttl=SESSION_TTL) is not None
 
     def test_cookie_minted_for_one_token_does_not_verify_against_another(self) -> None:
         cookie_value = mint_session_cookie("token-a")
 
-        assert verify_session_cookie(cookie_value, "token-b", session_ttl=3600) is None
+        assert verify_session_cookie(cookie_value, "token-b", session_ttl=SESSION_TTL) is None
 
     def test_malformed_cookie_value_does_not_verify(self) -> None:
-        assert verify_session_cookie("not-a-valid-cookie-shape", "the-real-token", session_ttl=3600) is None
+        assert verify_session_cookie("not-a-valid-cookie-shape", "the-real-token", session_ttl=SESSION_TTL) is None
 
     def test_tampered_signature_does_not_verify(self) -> None:
         token = "the-real-token"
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
Automated autofix review pass (code-reviewer, integration-reviewer, wtf-reviewer, llm-checker, lazy-checker, nitpicker).

**Refactor is faithful.** All 8 original scenarios (3 TTL, 5 renewal) survive the parametrize collapse 1:1 with matching inputs, expected values, and ids; the file still collects 37 cases, same as `main`. Boundary expectations were cross-checked against `session.py` (`age > session_ttl` for expiry, `session_ttl / 2 <= age <= session_ttl` for renewal). code-reviewer, integration-reviewer, and llm-checker all came back clean; `frozen_clock` correctly stays file-local (no shared clock helper exists in `tests/support/`, and the promotion threshold is 3+ files).

**Fixed in 3ed44b7b** (the readability/hygiene findings from wtf-reviewer and nitpicker):
- Finished the constant extraction the PR started — `SESSION_TTL` and `HALF_LIFE` now sit next to `BASE_EPOCH`, so boundary rows read `SESSION_TTL + 1` / `HALF_LIFE - 1` instead of bare `3601` / `1799` explained by a comment that only covered one of the three rows it applied to. `session_ttl=3600` is gone from all 11 call sites in the file.
- One calling convention for `frozen_clock` (`frozen_clock(0)`, not `frozen_clock(elapsed=0)` next to `frozen_clock(elapsed)`), plus a one-line note that the cookie is always minted at t=0.
- `expected` → `expected_renewed`, matching its `expected_issued_at` sibling.

Skipped: `assert issued_at == expected_issued_at` staying `==` rather than `is None` (inherent to table-driving a nullable return), and the "class now holds one method" note (still a meaningful behavior namespace).

Verified: `tests/unit/web` 210 passed, `prek -a` clean on all Python hooks, `prek pyright -a --stage pre-push` passed. (Frontend prek hooks fail locally only because worktrees don't share `node_modules/`.)

**Blocked, not by this PR:** `frontend` / `build` are red because `main` itself fails `tsc --noEmit` — `COMMON_STAT_CELL_COUNT` is declared but not exported in `stat-cell-builders.ts` while its test imports it. Tracked in #2177. `frontend` feeds `tests-complete`, which is a required check, so auto-merge cannot land this until #2177 is fixed on `main`. Confirmed pre-existing by reading `origin/main` directly; this PR touches only `tests/unit/web/test_auth_session.py`.

NodeJSmith:
@coderabbitai review

NodeJSmith:
Second automated autofix review pass (code-reviewer, integration-reviewer, wtf-reviewer, llm-checker, lazy-checker, nitpicker). CodeRabbit did not run — its check reports `Review rate limited`, and there are no review threads on this PR.

**Refactor independently re-verified as faithful.** Collected the `main` version of the file and this branch's side by side: both collect 37 cases, and all 8 original TTL/renewal scenarios map 1:1 onto the parametrized ids (`within_ttl_accepted`, `exactly_at_ttl_boundary_accepted`, `past_ttl_rejected`; `freshly_minted_not_renewed`, `just_before_half_life_not_renewed`, `exactly_at_half_life_renewed`, `past_half_life_renewed`, `past_full_ttl_not_renewed`) with matching inputs and expected values. Boundaries re-checked against `session.py` (`age > session_ttl` for expiry, `session_ttl / 2 <= age <= session_ttl` for renewal); both are asymmetric-sensitive, so the boundary rows still fail if the production comparison operators regress. code-reviewer, llm-checker, and lazy-checker all came back clean.

**Fixed in f5aed6fc** (two LOW findings):
- `BASE_EPOCH` was the one new constant without an attribute docstring, despite `1_000_000` being the least self-explanatory of the four values.
- The new TTL-boundary comment used a plain hyphen where the rest of the file uses an em dash (flagged independently by wtf-reviewer and nitpicker).

**Skipped, with reasons:**
- *Single vs. double backticks in the new `#` comments* (nitpicker). The claim that the file uses RST double backticks everywhere doesn't hold — the double-backtick usages are all in docstrings, and a pre-existing docstring on `main` already uses a single-backtick span. `#` comments aren't RST-rendered, so there's no convention being broken.
- *Name the bare `5` and `1000` elapsed offsets* (nitpicker). The asymmetry with their `HALF_LIFE ± 1` / `SESSION_TTL + 1` siblings is meaningful, not accidental: those are boundary-relative and the derivation is the point, while `5` and `1000` just mean "comfortably inside the window," which the param ids already say. A single-use `FRESH_OFFSET = 5` would add indirection without adding information.
- *Extend the `CLOCK_PATCH_TARGET` precedent to the `compare_digest` / `is_trusted_peer` patch strings* (nitpicker). Those lines are untouched by this diff; widening a test-only autofix PR into unrelated classes isn't worth it.
- *Promote `frozen_clock` into `tests/support/`* (integration-reviewer, advisory). Only one other site patches `hassette.web.auth.session._current_timestamp` — `tests/integration/web_api/test_auth.py:108`, with a different convention (seconds-ago-from-real-now rather than elapsed-since-base). That's 2 files, below the 3+ promotion threshold in `.claude/rules/test-conventions.md`, so filing an issue against the repo's own documented bar would be noise. Noted here instead.

Verified locally: `tests/unit/web/test_auth_session.py` 37 passed, `prek` clean on every applicable hook, `prek pyright -a --stage pre-push` passed.

**Still blocked, not by this PR:** `frontend` and `build (3.13)` are red because `main` itself fails `tsc --noEmit` — `stat-cell-builders.ts` declares `COMMON_STAT_CELL_COUNT` without exporting it while its test imports it. Confirmed by reading `origin/main` directly, and this branch touches only `tests/unit/web/test_auth_session.py`. Tracked in #2177, which is still open. `frontend` feeds the required `tests-complete` check, so auto-merge cannot land this until #2177 is fixed on `main`. `file-sizes` and `duplicate-code` are the usual `continue-on-error` baseline noise.

NodeJSmith:

NodeJSmith:

PR #2186, opened 2026-09-10
refactor: name the healthy sidebar group key and clarify group-def callback names

### LINKED ISSUE #2124: Clean up healthy-group literal and callback naming in use-group-open.ts
## Description

Quality scan of `frontend/src/components/layout/use-group-open.ts` (22 lines). Two of three checkers (llm-checker, lazy-checker) returned CLEAN — no comment bloat, no deferred debt, no dead code, no over-abstraction. The nitpicker surfaced two style/hygiene findings, both of which survive validity assessment. Both predate the current work.

### 1. Bare `"ok"` literal encodes the "healthy group" concept (`use-group-open.ts:13`)

`if (key === "ok" && allHealthy) return true;` hardcodes which `GroupKey` counts as the healthy group. The same concept is expressed in three other places in `sidebar-groups.ts` — the `GROUP_DEFS` entry (`:20`), `getGroupKey`'s fallback return (`:62`), and implicitly in `allHealthy`'s derivation (`:41-45`, which is defined as "err/blocked/warn/stopped are all empty", i.e. only `ok`/`disabled` remain). Nothing names that relationship.

Scoping note: this is a readability/intent finding, not a correctness risk. `GroupKey` is a string-literal union, so removing `"ok"` from it would make `key === "ok"` a no-overlap comparison and fail type checking. The nitpicker's stated rationale ("no compiler signal") overstates the risk; the finding stands on the weaker but real ground that a reader has to reconstruct the `allHealthy` ↔ `"ok"` coupling by reading two files.

**Proposed fix:** export a named constant from `sidebar-groups.ts` (e.g. `export const HEALTHY_GROUP_KEY: GroupKey = "ok";`), use it for the `GROUP_DEFS` entry key and `getGroupKey`'s fallback return, and import it here in place of the literal.

### 2. Single-letter callback parameter (`use-group-open.ts:6`)

`GROUP_DEFS.map((g) => [g.key, g.defaultOpen])` uses `g` outside a loop-index or math context. Self-consistent with `sidebar-groups.ts:33`, which uses the same name, but neither reads clearly.

**Proposed fix:** rename to `def` (or `groupDef`) here, and optionally at `sidebar-groups.ts:33` for consistency.

## Acceptance Criteria

- [ ] `"ok"` is referenced through a single exported named constant rather than repeated as a bare literal across `use-group-open.ts` and `sidebar-groups.ts`
- [ ] The `map` callback parameter in `use-group-open.ts:6` has a descriptive name
- [ ] `cd frontend && npm run build` and the existing `sidebar-groups.test.ts` suite pass unchanged (no behavior change intended)

## Human judgment

Not required. Both findings are mechanical, local, and behavior-preserving, with one clear fix each.

NodeJSmith:
## Summary

Addresses both nitpicker findings from the quality scan in #2124. There is no change in behavior.

- **Named the healthy sidebar group.** `sidebar-groups.ts` now exports `HEALTHY_GROUP_KEY: GroupKey = "ok"`. The `GROUP_DEFS` entry, the fallback return in `getGroupKey`, and the `allHealthy` auto-open check in `useGroupOpen` all use it. Before, the rule "`allHealthy` force-opens the `ok` group" was an unexplained `"ok"` literal, and a reader had to open both files to see the connection.
- **Renamed the single-letter callbacks.** In both `GROUP_DEFS.map(...)` callbacks (`sidebar-groups.ts` and `use-group-open.ts`), `g` is now `def`. This also settles an existing inconsistency: `sidebar.tsx` already used `def` for the same iteration.

The only `"ok"` literals left in these files are the `GroupKey` union that declares the value, the constant's own definition, and the unrelated `tone` union.

## Testing

- `npx tsc --noEmit`: clean
- `npm run build`: succeeds
- `npx vitest run src/components/layout`: 7 files, 107 tests passed, including `sidebar-groups.test.ts`

`sidebar-groups.test.ts` is deliberately left unchanged, still asserting the raw `"ok"` literal rather than importing the new constant. The test is the pin on that constant's value — importing it would make the assertion self-referential and stop it catching a change. The other five group keys are asserted as literals there too, so this also keeps the file internally consistent. #2124's acceptance criteria call for the suite to pass unchanged.

## Note on the earlier draft status

This PR was originally opened as a draft because `tsc (frontend)` failed on an unrelated `main` breakage (`COMMON_STAT_CELL_COUNT` not exported from `stat-cell-builders.ts`). That was fixed on `main` by #2189, which is merged into this branch. `tsc` now passes.

Closes #2124

### DIFF
--- a/frontend/src/components/layout/sidebar-groups.ts
+++ b/frontend/src/components/layout/sidebar-groups.ts
@@ -6,6 +6,8 @@ type AppManifest = components["schemas"]["AppManifestResponse"];
 
 export type GroupKey = "err" | "blocked" | "warn" | "ok" | "stopped" | "disabled";
 
+export const HEALTHY_GROUP_KEY: GroupKey = "ok";
+
 export interface GroupDef {
   key: GroupKey;
   label: string;
@@ -17,7 +19,7 @@ export const GROUP_DEFS: GroupDef[] = [
   { key: "err", label: "FAILING", tone: "err", defaultOpen: true },
   { key: "blocked", label: "BLOCKED", tone: "err", defaultOpen: true },
   { key: "warn", label: "SLOW", tone: "warn", defaultOpen: true },
-  { key: "ok", label: "RUNNING", tone: "ok", defaultOpen: false },
+  { key: HEALTHY_GROUP_KEY, label: "RUNNING", tone: "ok", defaultOpen: false },
   { key: "stopped", label: "STOPPED", tone: "mute", defaultOpen: true },
   { key: "disabled", label: "DISABLED", tone: "mute", defaultOpen: false },
 ];
@@ -30,7 +32,7 @@ export interface GroupedApps {
 }
 
 export function groupAndSortApps(manifests: AppManifest[], appStatuses: Record<string, AppStatusEntry>): GroupedApps {
-  const groups = new Map<GroupKey, AppManifest[]>(GROUP_DEFS.map((g) => [g.key, []]));
+  const groups = new Map<GroupKey, AppManifest[]>(GROUP_DEFS.map((def) => [def.key, []]));
   for (const m of manifests) {
     const key = getGroupKey(m, appStatuses);
     groups.get(key)!.push(m);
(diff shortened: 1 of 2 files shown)


NodeJSmith:
@coderabbitai review

PR #2187, opened 2026-09-10
test: split alert-shell tone test into one render per tone

### LINKED ISSUE #2130: Clean up tone test shape in alert-shell.test.tsx
Surfaced by the automated quality scanner (llm-checker + lazy-checker + nitpicker). Pre-existing; not introduced by any in-flight branch.

Two of three checkers returned CLEAN — no comment bloat, no deferred debt. One finding survived validity assessment.

## Finding: the tone test breaks the one-render-per-`it` shape used by every sibling primitive test

`alert-shell.test.tsx:18-31` bundles both tones into a single `it` block, renders `AlertShell` twice inside it, and aliases the destructured helper twice (`getByTestId: getDanger`, `getByTestId: getWarning`) to keep the two queries apart. Tests 1 and 3 in the same file do the opposite — one render, one plain `const el = getByTestId("b")`.

That divergence also drags in a second inconsistency: tests 1 and 3 use `data-testid="b"`, while test 2 switches to semantic ids (`"danger"`, `"warning"`) purely because two nodes coexist in the same block.

The sibling primitive tests establish the house idiom clearly:

- `frontend/src/components/shared/badge.test.tsx` — one `it` per variant, single render each, constant `data-testid="b"` throughout (21 occurrences)
- `frontend/src/components/shared/card.test.tsx` — same shape, constant `data-testid="c"` (10 occurrences)

`alert-shell.test.tsx` is the only one of the three that deviates.

### Proposed fix

Split `it("applies tone-specific background tokens")` into two single-render blocks matching `badge.test.tsx`, dropping the aliasing and the one-off semantic ids:

```tsx
it("applies the danger background token", () => {
  const { getByTestId } = render(
    <AlertShell tone="danger" data-testid="b">
      x
    </AlertShell>,
  );
  expect(getByTestId("b").className).toContain("bg-[var(--destructive-bg)]");
});

it("applies the warning background token", () => {
  const { getByTestId } = render(
    <AlertShell tone="warning" data-testid="b">
      x
    </AlertShell>,
  );
  expect(getByTestId("b").className).toContain("bg-[var(--status-warning-bg)]");
});
```

Net effect: every block in the file has the same shape, one `data-testid` value is used throughout, and a failure names which tone broke instead of reporting one bundled assertion.

## Acceptance criteria

- [ ] `it("applies tone-specific background tokens")` is split into one `it` per tone, each with a single `render` call
- [ ] No aliased `getByTestId:` destructuring remains in the file
- [ ] All `data-testid` values in the file are consistent, matching the `badge.test.tsx` / `card.test.tsx` pattern
- [ ] `cd frontend && npm test -- alert-shell` passes

## Rejected during validity assessment

Two nitpicker findings were dropped rather than filed:

- **"Extract the tone literals and expected class fragments into a shared `TONE_CLASS_MAP` fixture."** Claimed: repeating `"danger"` / `"warning"` and the `bg-[var(--...)]` strings across blocks is a magic-string problem. Actually: `badge.test.tsx` and `card.test.tsx` inline every expected value the same way (`toBe("success")`, `toBe("compact")`, etc.), and all 12 frontend test files that assert on classes use inline `className).toContain("literal")`. Why invalid: the proposed map would restate `TONE_CLASSES` from `alert-shell.tsx:13-16`, adding indirection to a two-entry table while making the assertions less direct — it conflicts with the established convention rather than correcting a deviation from it.
- **"Replace `data-testid="b"` with a named constant."** Claimed: the single-letter id is an unnamed magic string. Actually: it is the deliberate throwaway-id convention shared by `card.test.tsx` (`"c"`) and `badge.test.tsx` (`"b"`), neither of which uses a constant. Why invalid: the id's meaninglessness is the point — it marks the node under test in a single-render block. Only the file-internal *inconsistency* is a real problem, and that is covered by the surviving finding above.

NodeJSmith:
## Summary

Splits the bundled `it("applies tone-specific background tokens")` block in `frontend/src/components/shared/alert-shell.test.tsx` into two tests, one per tone. Each now renders once and uses a single `getByTestId`:

- `applies the danger background token`: asserts `bg-[var(--destructive-bg)]`
- `applies the warning background token`: asserts `bg-[var(--status-warning-bg)]`

This drops the aliased `getByTestId: getDanger` / `getByTestId: getWarning` destructuring and the one-off semantic test ids (`"danger"`, `"warning"`). Every block in the file now has the same shape and uses the same `data-testid="b"`, matching `badge.test.tsx` and `card.test.tsx`. When a tone breaks, the failing test's name now says which one.

The change is limited to this test file. No production code changes.

## Testing

- `npx vitest run src/components/shared/alert-shell.test.tsx`: 4/4 passed (the file had 3 tests before the split)
- prettier and eslint pre-commit hooks passed

`uv run nox -s dev` was not run, since this change touches no Python.

## Note on the earlier `tsc` failure

An earlier revision of this branch was blocked by a pre-existing `tsc` error on `main`
(`COMMON_STAT_CELL_COUNT` was not exported from `stat-cell-builders.ts`). That has since been
fixed on `main` and merged into this branch; the `frontend` check passes.

Closes #2130

### DIFF
--- a/frontend/src/components/shared/alert-shell.test.tsx
+++ b/frontend/src/components/shared/alert-shell.test.tsx
@@ -15,19 +15,22 @@ describe("AlertShell", () => {
     expect(el.className).toContain("rounded-md");
   });
 
-  it("applies tone-specific background tokens", () => {
-    const { getByTestId: getDanger } = render(
-      <AlertShell tone="danger" data-testid="danger">
+  it("applies the danger background token", () => {
+    const { getByTestId } = render(
+      <AlertShell tone="danger" data-testid="b">
         x
       </AlertShell>,
     );
-    const { getByTestId: getWarning } = render(
-      <AlertShell tone="warning" data-testid="warning">
+    expect(getByTestId("b").className).toContain("bg-[var(--destructive-bg)]");
+  });
+
+  it("applies the warning background token", () => {
+    const { getByTestId } = render(
+      <AlertShell tone="warning" data-testid="b">
         x
       </AlertShell>,
     );
-    expect(getDanger("danger").className).toContain("bg-[var(--destructive-bg)]");
-    expect(getWarning("warning").className).toContain("bg-[var(--status-warning-bg)]");
+    expect(getByTestId("b").className).toContain("bg-[var(--status-warning-bg)]");
   });
 
   it("forwards role and extra classes", () => {


NodeJSmith:
@coderabbitai review

PR #2188, opened 2026-09-10
test: extract cancel-listener construction and parametrize attribute checks in test_listeners.py

### LINKED ISSUE #2131: Extract the repeated cancel-listener construction in test_listeners.py
## Description

The `duplicate-code` CI job (PMD CPD) flagged a repeated block in `tests/unit/bus/test_listeners.py`. This predates the current work.

The shared shape is **constructing a cancel-listener fixture**: nearly every test in `TestCreateCancelListener` opens with the same four lines —

```python
tb = make_task_bucket()
listener = Listener.create_cancel_listener(
    task_bucket=tb,
    owner_id="test_owner",
    topic="hass.event.state_changed.light.kitchen",
    handler=lambda: None,
)
```

— followed by a single one-attribute assertion (`identity.source_tier`, `identity.owner_id`, `duration_config`, `invoker.rate_limiter`, `invoker.error_handler`, `topic`, `predicate`, `listener_id`). The `owner_id` and `topic` values vary arbitrarily between tests without the variation being what any of them are testing, so the construction is pure boilerplate around a one-line assertion.

There is no shared cancel-listener factory today. `tests/support/helpers.py` already provides `create_listener()` for ordinary listeners; the natural analogue is missing, so each test hand-rolls the construction.

## Goal

Extract the shared cancel-listener construction into a single helper — a `create_cancel_listener(**kw)` factory alongside the existing `create_listener()` in `tests/support/helpers.py` (or a module-local helper if it stays confined to this one file) — and use it at every call site with this shape. The attribute-assertion tests that differ only in which field they check are also good candidates for `pytest.mark.parametrize` over `(kwargs, attribute_path, expected)` once the construction is a one-liner.

Per `.claude/rules/test-conventions.md`, check `tests/support/factories.py` and `tests/support/helpers.py` before adding anything new, and prefer extending the shared module if the shape is needed outside this file.

## Where

PMD flagged these as one duplicated block:

- `tests/unit/bus/test_listeners.py:234-241`
- `tests/unit/bus/test_listeners.py:267-274`
- `tests/unit/bus/test_listeners.py:278-285`

Reading the file surfaced the same shape at six more call sites in the same class:

- `tests/unit/bus/test_listeners.py:245-252`
- `tests/unit/bus/test_listeners.py:256-263`
- `tests/unit/bus/test_listeners.py:289-296`
- `tests/unit/bus/test_listeners.py:300-308` (adds `predicate=`)
- `tests/unit/bus/test_listeners.py:312-319`
- `tests/unit/bus/test_listeners.py:325-332`

Any other instance of the same shape found while doing the work is in scope too, not just what is listed here.

## Acceptance Criteria

- [ ] A single cancel-listener construction helper exists, placed per the shared-factory rules in `.claude/rules/test-conventions.md`
- [ ] Every call site in `TestCreateCancelListener` uses it instead of constructing `Listener.create_cancel_listener(...)` inline
- [ ] Tests that differ only in the asserted attribute are parametrized rather than unrolled, where that does not obscure what each case covers
- [ ] `uv run pytest tests/unit/bus/test_listeners.py` passes
- [ ] `prek -a` is clean

NodeJSmith:
## Summary

PMD CPD flagged `TestCreateCancelListener` in `tests/unit/bus/test_listeners.py` as duplicate code. Nearly every test built a cancel-listener with the same `make_task_bucket()` + `Listener.create_cancel_listener(...)` block and then checked one attribute. The `owner_id`/`topic` values changed from test to test, but none of the tests were actually checking that variation.

Changes:

- Added a file-local `make_cancel_listener(*, owner_id=..., topic=..., predicate=None)` helper. It builds a cancel-listener with a fresh task bucket and the file's no-op `fn` handler. It stays module-local because nothing outside this file builds cancel-listeners. `tests/support/helpers.py` and `factories.py` have no equivalent, and `check_test_factories.py` passes.
- Replaced every inline `Listener.create_cancel_listener(...)` call in the class with the helper.
- Merged the four "defaults to `None`" assertions into one parametrized `test_unset_by_default`, whose cases are type-checkable `pytest.param(lambda listener: listener.<attr>, id=...)` extractors (`duration_config`, `invoker.rate_limiter`, `invoker.error_handler`, `predicate`).
- Kept `test_owner_id_is_preserved`, `test_topic_is_set`, and `test_predicate_can_be_set` as separate named tests, so the two string round-trips keep their `==` comparison and only the predicate asserts object identity.
- `test_source_tier_is_framework` and `test_works_without_bus_instance` stay as standalone tests, now one line each.
- Documented the new helper in `tests/unit/bus/CLAUDE.md` next to the other file-local bus test helpers.

Coverage is unchanged: each assertion from the old tests maps to one test or parametrized case. Net diff is -51 lines.

## Testing

- `uv run nox -s dev`: **7758 passed**, 1 skipped, 2 xfailed
- `prek run pyright -a --stage pre-push`: passed
- `prek -a`: every hook passed except **`tsc (frontend)`**, which fails with:
  ```
  src/components/app-detail/stat-cell-builders.test.ts(4,32): error TS2459: Module '"./stat-cell-builders"' declares 'COMMON_STAT_CELL_COUNT' locally, but it is not exported.
  ```
  This PR doesn't cause it. The branch changes only `tests/unit/bus/test_listeners.py` and `tests/unit/bus/CLAUDE.md`. The branch's merge-base is the current `origin/main` tip (`12739704`), and `git grep` shows `origin/main` has the same unexported `const COMMON_STAT_CELL_COUNT` in `stat-cell-builders.ts` alongside the test that imports it.

This PR is a draft only because the automated ship step opens a draft whenever `prek -a` fails. The work for #2131 is complete. Once the `tsc` failure on `main` is confirmed and tracked separately, this can be marked ready.

Needs human review.

Closes #2131


NodeJSmith:
@coderabbitai review

NodeJSmith:
Review note (posted after the fact — this PR auto-merged while its review battery was still running).

Six reviewers plus CodeRabbit and Codex ran against `09d3b3b`. CodeRabbit, Codex, and code-reviewer all reported zero findings. Two items did land, both now in #2195:

- `test_unset_by_default` parametrized four unrelated invariants over `lambda listener: listener.<attr>` extractors — wtf-reviewer (MEDIUM) and llm-checker independently flagged that this hides which attribute a case checks and leaves a plain pytest traceback unable to name the broken invariant. Split back into four named one-line tests; verified this does not reintroduce a CPD hit (`tools/check_duplicate_code.py` does not flag `test_listeners.py`; the tests are under the `MIN_TOKENS = 20` floor).
- The new `make_cancel_listener` entry in `tests/unit/bus/CLAUDE.md` was inserted mid-list rather than appended, breaking the append-only ordering the prior two entries established.

Also note: an automated review job pushed the fix commit to `autofix/issue-2131` after GitHub had already deleted it on merge, which re-created the branch. It carries only the #2195 change and can be deleted.

PR #2190, opened 2026-09-10
test: move stranded AppLink import into the import block in app-link tests

### LINKED ISSUE #2134: Move stranded component import in app-link.test.tsx back into the import block
Surfaced by the automated quality scanner (llm-checker / lazy-checker / nitpicker). Pre-existing, not introduced by any open branch.

Two of the three checkers returned CLEAN. One finding survived validity assessment.

## Finding 1 — Component import is stranded below `vi.mock()` with no explanation

`frontend/src/components/shared/app-link.test.tsx` splits its import block around an executable statement:

```ts
import { createWouterMock } from "../../test/mock-wouter";   // line 4

vi.mock("wouter", () => createWouterMock());                  // line 6

import { AppLink } from "./app-link";                         // line 8
```

There is a real hazard that justifies this layout in some files, and `mock-wouter.ts`'s `createWouterMock` docstring documents it precisely: a plain top-level import of the component under test is unsafe only when that import can sort *before* `../test/mock-wouter` under eslint's import sorter, which in practice happens only when both imports share the same relative depth (the `src/app.test.tsx` case, where `./app` and `./test/mock-wouter` can sort either way).

That condition does not hold here. `../../test/mock-wouter` is parent-relative and `./app-link` is same-directory, so the mock helper's import always sorts first — the documented hazard cannot fire, and the split buys nothing.

The result is a file that reads as if it is working around a problem it does not have, and that diverges from the dominant convention: of the ~15 test files calling `createWouterMock`, nearly all keep every import at the top and place `vi.mock` after them (`pages/apps-table-row.test.tsx`, `pages/logs.test.tsx`, `pages/apps.test.tsx`, `hooks/use-breadcrumbs.test.tsx`, and others). Where a file genuinely does need special handling, it says so — `hooks/use-correct-url.test.ts` carries an explanatory comment, and `src/app.test.tsx` uses a dynamic import inside the factory with a comment pointing at the docstring. This file has neither the comment nor the need.

### Proposed fix

Move line 8 up into the import block with the other local imports, leaving `vi.mock("wouter", () => createWouterMock());` as the first statement after the imports — matching `pages/apps-table-row.test.tsx`.

Verify by running the file (`cd frontend && npx vitest run src/components/shared/app-link.test.tsx`). This is a deterministic check: if the tests pass, the change is correct. In the unlikely event the "Cannot access before initialization" error does appear, keep the split and add a one-line comment pointing at `createWouterMock`'s import-order note, as `hooks/use-correct-url.test.ts` does.

`frontend/src/components/shared/log-table/log-table-view.test.tsx` has the same unexplained split (component import at line 24, after three `vi.mock` calls) and is worth the same treatment if touched, but it is out of scope for this scan.

## Not filed — assessed as likely invalid

**Extract repeated `"my_app"` / `"listener"` / `42` literals into named constants** (nitpicker, 3 overlapping findings).

- **Claimed:** the literal `"my_app"` appears ~10 times and `"/apps/my_app"` ~8 times, risking "a typo in one assertion silently testing the wrong string."
- **Actually:** in every `it()` block the literal appears twice — once as the input prop (`appKey="my_app"`) and once inside a hardcoded expected string (`expect(href).toBe("/apps/my_app?instance=2")`). A typo in either position makes the assertion mismatch and the test fail loudly. There is no path by which a mistyped literal produces a passing test.
- **Why invalid:** the stated failure mode cannot occur, and the proposed fix makes the tests weaker. Deriving the expected paths from shared constants moves them toward re-deriving `appDetailPath` / `handlerPath` — the exact logic under test — which is what hardcoded expected values exist to prevent. The inline-literal style is also the established norm for this literal in this directory: `components/shared/action-buttons.test.tsx` inlines `appKey="my_app"` 12+ times.

NodeJSmith:
## Summary

Moves the `AppLink` import in `frontend/src/components/shared/app-link.test.tsx` up into the import block, so `vi.mock("wouter", () => createWouterMock())` is the first statement after the imports. This matches `pages/apps-table-row.test.tsx` and most of the other `createWouterMock` test files.

The old split layout guards against a hazard that can't happen in this file. `createWouterMock`'s docstring says a top-level import of the component under test is only unsafe when it can sort before `../test/mock-wouter`. That only happens when both imports are at the same relative depth. Here, `../../test/mock-wouter` is parent-relative and `./app-link` is in the same directory, so the mock helper always sorts first.

This is a test-only change with no runtime behavior change.

## Testing

- `npx vitest run src/components/shared/app-link.test.tsx`: 9/9 passed. No "Cannot access before initialization" error, which confirms the import order is safe.
- `npx vitest run` (full frontend suite): 1604/1604 passed across 110 files.
- `npx eslint src/components/shared/app-link.test.tsx`: clean. The import sorter accepts the new order.
- `npx tsc --noEmit`: clean. The `COMMON_STAT_CELL_COUNT` type error that this branch originally hit was a `main` problem, fixed by #2189 and picked up here by merging `main`.
- `uv run nox -s dev` was not run. It covers the Python suite, and this PR touches no Python.

The `duplicate-code` and `file-sizes` lint jobs report FAILURE, as they do on `main`. Both run with `continue-on-error` and track a pre-existing backlog; neither is affected by this diff.

## Follow-up

`frontend/src/components/shared/log-table/log-table-view.test.tsx` has the same unexplained split and was out of scope for this change. Filed as #2194.

Closes #2134

### DIFF
--- a/frontend/src/components/shared/app-link.test.tsx
+++ b/frontend/src/components/shared/app-link.test.tsx
@@ -2,10 +2,9 @@ import { render } from "@testing-library/react";
 import { describe, expect, it, vi } from "vitest";
 
 import { createWouterMock } from "../../test/mock-wouter";
+import { AppLink } from "./app-link";
 
 vi.mock("wouter", () => createWouterMock());
-
-import { AppLink } from "./app-link";
 
 describe("AppLink — basic href", () => {
   it("renders a link to /apps/:key with no extras", () => {


NodeJSmith:
@coderabbitai review

PR #2193, opened 2026-09-11
chore: name the mobile visible-crumb count in breadcrumbs

### LINKED ISSUE #2141: Link the duplicated mobile-crumb-count literals in breadcrumbs.tsx
## Description

Quality-scanner pass over `frontend/src/components/shared/breadcrumbs.tsx`. One finding survived validity assessment; the rest were dropped as invalid (see bottom).

## Finding: the "keep 2 crumbs on mobile" decision is encoded in three unlinked literals

The mobile collapse behavior depends on the number `2` appearing in three separate places, with nothing tying them together:

- `breadcrumbs.tsx:30` — `items.length > 2` decides whether the ellipsis `<li>` renders at all
- `breadcrumbs.tsx:43` — `max-sidebar:[&:not(:nth-last-child(-n+2))]:sr-only` clips all but the last two crumbs
- `breadcrumbs.tsx:44` — `max-sidebar:[&:nth-last-child(2)>span]:hidden` hides the separator on the second-to-last crumb

Changing the design to "show the last 3 crumbs" requires finding and correctly updating all three, and the two CSS ones are buried inside Tailwind arbitrary-value pseudo-selector chains where the shared meaning of `2` is not visible. Getting one of the three wrong produces a silent visual bug (a stray separator, or an ellipsis that renders when nothing is actually clipped) rather than a type or lint error.

### Proposed fix

Tailwind's JIT extracts class names statically from source, so the arbitrary values on lines 43-44 cannot interpolate a JS constant — the literal must stay in the class string. Given that constraint:

1. Name the JS-side value: `const MOBILE_VISIBLE_CRUMBS = 2;` at the top of the file, used at line 30 (`items.length > MOBILE_VISIBLE_CRUMBS`).
2. Add a one-line comment directly above the `cn(...)` call noting that the `nth-last-child` literals must stay in sync with `MOBILE_VISIBLE_CRUMBS` and cannot be interpolated because Tailwind scans class names statically.

That makes the coupling discoverable from either end without fighting the build tooling.

## Acceptance Criteria

- [ ] The `2` at `breadcrumbs.tsx:30` is a named module-level constant
- [ ] A comment at the `cn(...)` call links the `nth-last-child` literals to that constant and states why they can't be interpolated
- [ ] No visual or behavioral change to the breadcrumb trail at any width

## Findings assessed as invalid (not filed)

- **Doc comment at lines 16-23 is verbose** — each sentence carries a distinct non-derivable fact (which crumbs are clipped and at what breakpoint; why clipping is needed at all; why clipped rather than removed). It is design rationale, which the house style asks for.
- **`hover:text-[var(--primary-hover)]` should be a named utility** — no such utility exists, and this is the established repo pattern (`app-link.tsx:28`, `log-table-view.tsx:90`, `log-detail-drawer.tsx:252,271`).
- **`interface Props` should be `BreadcrumbsProps`** — bare `Props` is used in 27 `.tsx` files in this codebase; renaming one file in isolation would increase inconsistency, not reduce it.
- **`"breadcrumbs"` default `data-testid` should be a constant** — single use as a default parameter value; extracting it adds indirection with no second consumer.

NodeJSmith:
## Summary

The breadcrumb trail keeps the last two crumbs visible below the sidebar breakpoint. That `2` was spread across three unlinked literals in `frontend/src/components/shared/breadcrumbs.tsx`: the JS check that decides whether the ellipsis renders, and two Tailwind `nth-last-child` selectors that clip the earlier crumbs and hide the separator. Changing the design to show three crumbs meant finding all three, and missing one would cause a quiet visual bug (a stray separator, or an ellipsis with nothing clipped behind it).

Changes:

- Added a module-level `MOBILE_VISIBLE_CRUMBS = 2` constant and used it in the ellipsis check (`items.length > MOBILE_VISIBLE_CRUMBS`).
- Added a comment above the `nth-last-child` class strings. It says the literals there must match `MOBILE_VISIBLE_CRUMBS` and can't be interpolated, because Tailwind reads class names statically from source.

This doesn't change behavior or rendering. The rendered values are identical.

## Testing

- `uv run nox -s dev`: 7758 passed, 1 skipped, 2 xfailed
- `prek -a`: all hooks pass
- Screenshots of the breadcrumb trail at desktop (1280px) and mobile (375px) widths, below.

## Screenshots

**Desktop (1280px):** full trail

Breadcrumbs at 1280px

**Mobile (375px):** collapsed to the last two crumbs with a leading ellipsis

Breadcrumbs at 375px

Closes #2141


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2195, opened 2026-09-11
test: split cancel-listener default-value parametrization into named tests

NodeJSmith:
## Summary

Follow-up to #2188, which auto-merged while its review battery was still running. These are the review findings that landed too late to fold into that PR.

- **Split `test_unset_by_default` into four named tests.** Three reviewers (wtf-reviewer MEDIUM, llm-checker, and the earlier attrgetter pass) independently flagged the same shape: parametrizing four unrelated invariants — no duration config, no rate limiter, no error handler, no predicate — over `lambda listener: listener.<attr>` extractors hid which attribute a case checked behind a generic `get_attr` parameter, and a plain pytest traceback no longer named the broken invariant (only the node id did). The four one-line tests each carry their own docstring again. The now-unused `collections.abc.Callable` import is dropped.
- **Append the `make_cancel_listener` entry to the end of `tests/unit/bus/CLAUDE.md`'s file-local helper list.** #2188 inserted it mid-list; the prior two entries established append-only ordering, so "last entry = most recently added helper" holds again.

The dedup this closes out — `make_cancel_listener()` itself — is unchanged and stays on `main`.

## Testing

- `uv run pytest tests/unit/bus/test_listeners.py -q`: **56 passed** (same count as before — the four parametrized cases become four tests)
- `uv run pyright tests/unit/bus/test_listeners.py`: 0 errors
- `prek run --files tests/unit/bus/test_listeners.py tests/unit/bus/CLAUDE.md`: all hooks passed
- `uv run python tools/check_duplicate_code.py`: `test_listeners.py` is not flagged — the four one-line tests stay under PMD CPD's `MIN_TOKENS = 20` floor, so splitting them back out does not reintroduce the duplication #2131 was filed to remove.

Reviewed by code-reviewer (PASS, 0 findings), integration-reviewer, wtf-reviewer, llm-checker, lazy-checker (CLEAN), and nitpicker. CodeRabbit and Codex both reported no findings on #2188.


NodeJSmith:
@coderabbitai review

PR #2197, opened 2026-09-11
test: name the repeated card test id literals as constants

### LINKED ISSUE #2146: Name the repeated test id literal in card.test.tsx
Surfaced by an automated quality scan of `frontend/src/components/shared/card.test.tsx`. Both the LLM-bias and deferred-debt reviews came back clean; the only surviving findings are naming hygiene around the test id literal.

## Finding 1 — the `data-testid` value is an unnamed single-character literal repeated 10 times

`"c"` appears as both the rendered `data-testid` and the `getByTestId` lookup in nearly every test (lines 10, 18, 27, 37, 47, 56, 64, 76, 92, 97). It communicates nothing about what is being queried, and it is re-declared inline at every call site rather than named once.

## Finding 2 — one block uses a different literal for the same purpose

The "pass-through attributes" block uses `"my-card"` (line 86) instead of `"c"`, with nothing distinguishing that case from the others. Same concept, two values, no stated reason.

## Proposed fix

Both findings collapse into one mechanical change: hoist a single descriptive constant at the top of the file and use it everywhere, including line 86.

```tsx
const TEST_ID = "card";
```

Then `data-testid={TEST_ID}` in every `render(...)` and `getByTestId(TEST_ID)` in every assertion. Confined to this one file, no behavior change, verified by the existing suite passing unchanged.

## Explicitly not included

The repeated variant strings (`"default"`, `"compact"`, `"config"`, `"error"`) were raised by the scanner but rejected. Lines 22, 31, and 51 assert that `getAttribute("data-variant")` equals those exact strings, so the literal is the contract under test. Sharing one constant between the prop and the assertion would make those tests tautological. Leave them as literals.

## Acceptance Criteria

- [ ] A single named constant holds the test id value, declared once at module scope
- [ ] Every `render` and `getByTestId` in the file uses that constant, including the "pass-through attributes" block
- [ ] Variant string literals are left inline and unchanged
- [ ] `npm run test` passes for this file with no assertion edits beyond the test id substitution

NodeJSmith:
## Summary

`card.test.tsx` used the single-character literal `"c"` as the `data-testid` in almost every test, and the "pass-through attributes" block used `"my-card"` for the same purpose with no stated reason for the difference.

This PR names both literals as module-level constants:

- `CARD_TEST_ID = "card"` — used for every `data-testid` prop and `getByTestId` lookup in the file except one.
- `PASSTHROUGH_TEST_ID = "my-card"` — used only by `passes data-testid through to the div element`.

The pass-through test keeps a distinct value on purpose. Its assertion is only meaningful if the id it renders appears nowhere else in the file; collapsing it into `CARD_TEST_ID` would make it a near-duplicate of `renders a <div> element` and it would pass whether or not pass-through actually worked. A comment on that test records the invariant. This deviates from acceptance criterion 2 in #2146 ("including the pass-through block") — the issue's objection was two values with no stated reason, and the named constant plus the comment supplies the reason.

The variant string literals (`"default"`, `"compact"`, `"config"`, `"error"`) stay inline on purpose. The tests assert that `data-variant` equals those exact strings, so the literal is the contract under test. Sharing a constant between the prop and the assertion would make those tests tautological.

The `data-testid="child"` literal on the inner `<span>` also stays inline: it identifies a different DOM node, is used once, and module-scoping it would obscure that.

This only touches tests: one file, no production code, no behavior change.

## Testing

- `npx vitest run src/components/shared/card.test.tsx`: 11 passed
- `npx tsc --noEmit`, `npx eslint`, `npx prettier --check`: clean
- `prek` pre-commit hooks: all pass

Closes #2146

### DIFF
--- a/frontend/src/components/shared/card.test.tsx
+++ b/frontend/src/components/shared/card.test.tsx
@@ -4,70 +4,71 @@ import { describe, expect, it } from "vitest";
 
 import { Card } from "@/components/ui/card";
 
-const TEST_ID = "card";
+const CARD_TEST_ID = "card";
+const PASSTHROUGH_TEST_ID = "my-card";
 
 describe("Card", () => {
   describe("renders as div", () => {
     it("renders a <div> element", () => {
-      const { getByTestId } = render(<Card data-testid={TEST_ID}>content</Card>);
-      expect(getByTestId(TEST_ID).tagName.toLowerCase()).toBe("div");
+      const { getByTestId } = render(<Card data-testid={CARD_TEST_ID}>content</Card>);
+      expect(getByTestId(CARD_TEST_ID).tagName.toLowerCase()).toBe("div");
     });
   });
 
   describe("variant prop", () => {
     it("applies base card styling when variant='default'", () => {
       const { getByTestId } = render(
-        <Card variant="default" data-testid={TEST_ID}>
+        <Card variant="default" data-testid={CARD_TEST_ID}>
           content
         </Card>,
       );
-      expect(getByTestId(TEST_ID).getAttribute("data-variant")).toBe("default");
+      expect(getByTestId(CARD_TEST_ID).getAttribute("data-variant")).toBe("default");
     });
 
     it("applies compact styling when variant='compact'", () => {
       const { getByTestId } = render(
-        <Card variant="compact" data-testid={TEST_ID}>
+        <Card variant="compact" data-testid={CARD_TEST_ID}>
           content
         </Card>,
       );
-      expect(getByTestId(TEST_ID).getAttribute("data-variant")).toBe("compact");
-      expect(getByTestId(TEST_ID).className).toMatch(/p-3/);
+      expect(getByTestId(CARD_TEST_ID).getAttribute("data-variant")).toBe("compact");
+      expect(getByTestId(CARD_TEST_ID).className).toMatch(/p-3/);
     });
 
     it("applies config styling when variant='config'", () => {
       const { getByTestId } = render(
-        <Card variant="config" data-testid={TEST_ID}>
+        <Card variant="config" data-testid={CARD_TEST_ID}>
           content
         </Card>,
       );
-      expect(getByTestId(TEST_ID).className).toMatch(/overflow-hidden/);
-      expect(getByTestId(TEST_ID).className).toMatch(/p-0/);
+      expect(getByTestId(CARD_TEST_ID).className).toMatch(/overflow-hidden/);
+      expect(getByTestId(CARD_TEST_ID).className).toMatch(/p-0/);
     });
 
     it("applies error styling when variant='error'", () => {
       const { getByTestId } = render(
-        <Card variant="error" data-testid={TEST_ID}>
+        <Card variant="error" data-testid={CARD_TEST_ID}>
           content
         </Card>,
       );
-      expect(getByTestId(TEST_ID).getAttribute("data-variant")).toBe("error");
-      expect(getByTestId(TEST_ID).className).toMatch(/text-center/);
+      expect(getByTestId(CARD_TEST_ID).getAttribute("data-variant")).toBe("error");
+      expect(getByTestId(CARD_TEST_ID).className).toMatch(/text-center/);
     });
 
     it("uses default variant when no variant is provided", () => {
-      const { getByTestId } = render(<Card data-testid={TEST_ID}>content</Card>);
-      expect(getByTestId(TEST_ID).getAttribute("data-variant")).toBe("default");
+      const { getByTestId } = render(<Card data-testid={CARD_TEST_ID}>content</Card>);
+      expect(getByTestId(CARD_TEST_ID).getAttribute("data-variant")).toBe("default");
     });
   });
 
   describe("class prop", () => {
     it("merges additional class into div className", () => {
       const { getByTestId } = render(
-        <Card className="my-layout-class" data-testid={TEST_ID}>
+        <Card className="my-layout-class" data-testid={CARD_TEST_ID}>
           content
         </Card>,
       );
-      expect(getByTestId(TEST_ID).className).toMatch(/my-layout-class/);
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
Review pass (code / integration / WTF / llm-checker / lazy-checker / nitpicker). No CodeRabbit threads — its check reported `Review rate limited`.

Two findings fixed in 080c8f19:
- **Naming (integration, WTF, nitpicker):** `TEST_ID` was the only bare `TEST_ID` constant in the frontend suite. Renamed to `CARD_TEST_ID` to match the existing `<CONTEXT>_TEST_ID` convention (`HANDLER_TEST_ID` in `registration-footer.test.tsx`, `HANDLER_LIST_TEST_ID` in `handler-list.test.tsx`).
- **Pass-through test (code-reviewer, WTF, nitpicker):** collapsing `"my-card"` into the shared constant made `passes data-testid through to the div element` a near-duplicate of `renders a <div> element` — it no longer showed that an *arbitrary* caller-supplied id reaches the DOM. Restored as a named `PASSTHROUGH_TEST_ID` with a comment stating why it differs. This deviates from acceptance criterion 2 in #2146 ("including the pass-through block"); the issue's objection was "two values, no stated reason", and the constant plus comment supplies the reason.

Skipped: nitpicker's suggestion to hoist the inner `"child"` literal — it identifies a different DOM node, is used only inside one test, and module-scoping it would obscure that.

Verified: eslint clean, `tsc --noEmit` clean, 11/11 tests pass. `duplicate-code` and `file-sizes` fail on main too (known `continue-on-error` baseline).

NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

NodeJSmith:
Second review pass on head `3dbd3cd9` (code / integration / WTF / llm-checker / lazy-checker / nitpicker).

**CodeRabbit: no threads.** Its check reports `Review skipped: automatic reviews are disabled`, and three manual `@coderabbitai review` triggers (03:13, 03:21, 03:25 UTC) all returned `Review rate limited` — the free OSS review quota is exhausted, not a short cooldown. So there was no CodeRabbit input to triage on this pass.

**Reviewer verdicts:** code-reviewer PASS (0), integration-reviewer PASS (0), lazy-checker CLEAN, wtf-reviewer 1 LOW, llm-checker 2 LOW, nitpicker 3.

**Fixed in 3dbd3cd9** — the pass-through comment. Two reviewers independently flagged that `// Deliberately not CARD_TEST_ID: ...` was phrased as a rebuttal to the shared constant rather than a statement of what the test requires. Reworded to state the invariant directly:

```
// Needs an id no other test in this file renders, so the assertion can only pass
// if an arbitrary caller-supplied id actually reached the DOM.
```

**Also fixed:** the PR title and body had drifted from the diff — both described a single `TEST_ID` constant used everywhere "including the pass-through block," which stopped being true when 080c8f19 split out `PASSTHROUGH_TEST_ID`. Since squash-merge lands the body as the commit body, both are now updated to match the actual change and to record why the variant strings and the `"child"` literal stay inline.

**Skipped:**
- *nitpicker — extract the inner `data-testid="child"` literal.* Single use, identifies a different DOM node than the card root. Module-scoping a single-use literal adds indirection with no dedup benefit and would blur the distinction between the two nodes.
- *nitpicker — rename to `DEFAULT_TEST_ID` / `CUSTOM_TEST_ID`.* `CARD_TEST_ID` matches the repo's existing `<CONTEXT>_TEST_ID` precedent (`HANDLER_TEST_ID`, `HANDLER_LIST_TEST_ID`), which integration-reviewer confirmed independently. Renaming would diverge from the established convention to fix a naming concern the accompanying comment already resolves.

**Follow-up filed:** #2198 — `badge.test.tsx` (12×) and `alert-shell.test.tsx` (5×) still inline the same `data-testid="b"` literal this PR removed from `card.test.tsx`. Verified unchanged vs `origin/main`, so pre-existing, not introduced here.

**Verified:** 11/11 tests pass, `tsc --noEmit` clean, eslint clean, prettier clean, all `prek` pre-commit hooks pass. `duplicate-code` and `file-sizes` fail as expected — both carry `continue-on-error: true` and fail on `main` too. Auto-merge (squash) is armed.

NodeJSmith:
Third review pass on head `3dbd3cd9` (code / integration / WTF / llm-checker / lazy-checker / nitpicker).

**CodeRabbit: still no threads.** Its check reports `Review rate limited` — the quota exhausted in the earlier passes has not refilled, so there was again no CodeRabbit input to triage.

**Reviewer verdicts:** code-reviewer PASS (0), integration-reviewer PASS (0), llm-checker CLEAN, lazy-checker CLEAN, nitpicker CLEAN (0), wtf-reviewer 3 LOW.

**No changes made.** All three wtf findings are LOW naming/comment nits and none warranted a churn commit:

- *`CARD_TEST_ID = "card"` is a value change, not just a rename (the original literal was `"c"`).* True, but the id is arbitrary and file-local, all 12 call sites moved together, and the assertions only compare the id to itself. Nothing outside the file reads it.
- *The comment explains why the pass-through test needs a distinct id, but not "why two constants exist" when read from the top of the file.* The comment sits on the one test that depends on the distinction, which is where a reader hits the question in practice.
- *Rename `PASSTHROUGH_TEST_ID` to something like `DISTINCT_TEST_ID`.* Already considered and declined in the second pass — `<CONTEXT>_TEST_ID` matches the repo precedent (`HANDLER_TEST_ID`, `HANDLER_LIST_TEST_ID`), independently confirmed again by integration-reviewer and llm-checker this pass.

**Verified:** 11/11 tests pass locally (`npx vitest run src/components/shared/card.test.tsx`), working tree clean. All four required checks green (`tests-complete`, `e2e-complete`, `lint-complete`, `docs/readthedocs.org:hassette`). `duplicate-code` and `file-sizes` fail as expected — both `continue-on-error`, both fail on `main`.

Auto-merge fired on `tests-complete` going green; PR is merged.

Issue #2157, opened 2026-09-09
Tighten string-value assertion and testid duplication in detail-stats.test.tsx


NodeJSmith:
## Description

Automated quality scan of `frontend/src/components/shared/detail-stats.test.tsx`. Two findings survived validity assessment: one test whose declared purpose is never actually asserted, and repeated testid string literals. Both predate the current work.

### 1. `renders string values (dashes for empty)` never asserts the string value renders

`frontend/src/components/shared/detail-stats.test.tsx:37-48`

The test name claims it verifies string values (em dashes) render, but the body only asserts that the `Min` and `Max` *labels* appear and that two cells exist:

```tsx
expect(row.textContent).toContain("Min");
expect(row.textContent).toContain("Max");
const dashCells = row.querySelectorAll("[data-testid='stats-cell']");
expect(dashCells.length).toBe(2);
```

The `"—"` value is never asserted. If `DetailStats` stopped rendering `{cell.value}` for string values entirely (`detail-stats.tsx:41-43`), this test would still pass. The cell-count assertion also duplicates what `generates per-cell testids from parent testid` (`:50-54`) already covers, so the test contributes no unique coverage today.

The variable name `dashCells` also misdescribes what it holds — it is every stat cell, not only the dash-valued ones.

**Proposed fix:** assert the rendered values, e.g. query the two `[data-tone]` value spans (or the cells) and assert their `textContent` is `"—"`. Drop the redundant cell-count assertion, and rename `dashCells` to match what it actually selects if it is kept.

### 2. `"stats"` / `"stats-cell"` testid literals repeated across every test

`frontend/src/components/shared/detail-stats.test.tsx:14,15,23,24,31,32,42,46,51,52`

`"stats"` is passed to `render()` and re-queried via `getByTestId("stats")` six times; the derived `[data-testid='stats-cell']` selector appears three times. A change to the testid convention means editing nine call sites.

**Proposed fix:** hoist a single `const TEST_ID = "stats";` next to `baseCells` and derive the cell selector from it (`` `[data-testid='${TEST_ID}-cell']` ``). This also documents the parent-to-cell testid derivation that `DetailStats` implements at `detail-stats.tsx:36`.

## Acceptance Criteria

- [ ] `renders string values (dashes for empty)` asserts that the `"—"` value actually renders in the cell values, so removing value rendering from `DetailStats` fails the test
[...]

PR #2200, opened 2026-09-11
test: assert string values render in DetailStats test and dedupe testids

NodeJSmith:
## Summary

Fixes two test-quality findings in `frontend/src/components/shared/detail-stats.test.tsx`:

1. **`renders string values (dashes for empty)` now checks the string values.** Before, it only checked that the `Min`/`Max` labels showed up and that two cells existed, so it would still pass if `DetailStats` stopped rendering `{cell.value}`. It now collects each cell's `[label, value]` text and asserts `[["Min", "—"], ["Max", "—"]]`. With value rendering removed, the value span is empty and the test fails. The duplicate cell-count assertion and the misnamed `dashCells` variable are gone.
2. **Testid literals are defined once.** `STATS_TEST_ID` and `STATS_CELL_SELECTOR` (built from `${STATS_TEST_ID}-cell`, the same way `DetailStats` builds per-cell testids) replace the nine repeated `"stats"` / `[data-testid='stats-cell']` literals.

This is a test-only change with no production or rendered UI changes.

## Review follow-ups

- Added a comment recording the DOM child-order assumption the pair extraction relies on (label span first, then value span).
- Renamed the `cells` binding in `generates per-cell testids from parent testid` to `statCells` so both tests name the same `querySelectorAll(STATS_CELL_SELECTOR)` result identically.

## Testing

- Mutation check: replacing `{cell.value}` with an empty expression in `detail-stats.tsx` makes `renders string values (dashes for empty)` fail, confirming the assertion is load-bearing.
- `npx vitest run src/components/shared/detail-stats.test.tsx`: 5/5 passed
- `npx tsc --noEmit`: clean
- `prek run --files frontend/src/components/shared/detail-stats.test.tsx`: prettier, eslint, tsc and the other applicable hooks all passed
- Not run: the backend `uv run nox -s dev` suite, because no Python files changed

Partial progress on #2157

### DIFF
--- a/frontend/src/components/shared/detail-stats.test.tsx
+++ b/frontend/src/components/shared/detail-stats.test.tsx
@@ -45,6 +45,7 @@ describe("DetailStats", () => {
     ];
     const { getByTestId } = render(<DetailStats cells={cells} data-testid={STATS_TEST_ID} />);
     const statCells = getByTestId(STATS_TEST_ID).querySelectorAll(STATS_CELL_SELECTOR);
+    // Each cell renders exactly two children, in order: the label span, then the value span.
     const labelValuePairs = Array.from(statCells, (cell) => Array.from(cell.children, (child) => child.textContent));
     expect(labelValuePairs).toEqual([
       ["Min", "—"],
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #2202, opened 2026-09-11
fix: hide decorative empty-state icon from screen readers

### LINKED ISSUE #2165: Mark decorative EmptyState icon as aria-hidden
Surfaced by an automated quality scan of `frontend/src/components/shared/empty-state.tsx` (llm-checker, lazy-checker, nitpicker). Both the LLM-bias and deferred-debt passes came back clean — the component is small, well-scoped, and has no comment bloat or accumulated shortcuts. Two hygiene items survived validity assessment.

## Finding 1 — Decorative icon glyph is not `aria-hidden`, so screen readers announce it as content

`EmptyState` renders its icon as a bare text glyph with no `aria-hidden`:

```tsx
{icon && <div className="mb-2 text-[length:var(--text-h1)] text-foreground-faint">{icon}</div>}
```
(`frontend/src/components/shared/empty-state.tsx:14`)

Every icon actually passed by a caller is purely decorative — the `title` prop always carries the meaning:

- default `"∅"` (`empty-state.tsx:11`)
- `icon="←"` — `app-detail/handlers-tab.tsx:71` ("Select a handler or job to see details.")
- `icon="◌"` — `shared/execution-table.tsx:177` ("no invocations recorded")
- `icon="🔍"` — `shared/empty-state.test.tsx:14`

A screen reader currently reads the glyph aloud ("empty set", "leftwards arrow", "magnifying glass tilted left") immediately before the real message, on every empty state in the dashboard. `EmptyState` has 15+ call sites across `pages/apps.tsx`, `pages/handlers.tsx`, `pages/logs.tsx`, `components/app-detail/*`, `components/diagnostics/services-panel.tsx`, and `components/shared/config-schema-view.tsx`, so this affects every empty view in the UI.

This is also a deviation from an established, tested convention in this frontend. Decorative glyphs are wrapped in `aria-hidden="true"` everywhere else — the closest analogue is a bare text glyph, identical in shape to this one:

```tsx
<span aria-hidden="true">{arrow}</span>
```
(`frontend/src/components/shared/sort-header.tsx:97`)

Same pattern in `app-detail/error-spotlight.tsx:30`, `app-detail/unified-handler-row.tsx:87`, `app-detail/handler-health-card.tsx:95`, `shared/status-shape.tsx:64`, `shared/mini-sparkline.tsx:26`, `shared/breadcrumbs.tsx:48`, and the icon set — which has dedicated tests asserting the attribute (`shared/icons.test.tsx:37`, `shared/filter-icon.test.tsx:33`).

**Proposed fix:** add `aria-hidden="true"` to the icon `<div>` in `empty-state.tsx:14`, and add a test in `empty-state.test.tsx` asserting it, matching the existing assertions in `icons.test.tsx` / `filter-icon.test.tsx`.

## Finding 2 — Default icon glyph is an unnamed inline literal (minor)

`icon = "∅"` sits inline in the destructuring default (`empty-state.tsx:11`). The mathematical empty-set character is not self-evident when scanning the signature, and it is visually easy to confuse with `0`, `Ø`, or `⌀` in a monospace diff.

**Proposed fix:** hoist it to a module-scope constant, e.g. `const DEFAULT_EMPTY_ICON = "∅";`, and use it as the default.

**Note on scope:** this is a readability rename only. The scanner's original framing — that the literal lacks a "shared source of truth to reuse" — does not hold: `grep -rn "∅" frontend/src` returns only `empty-state.tsx:11` and its test assertion at `empty-state.test.tsx:10`, so there is no duplication to consolidate. This is the single definition site for all callers. Fix it for legibility or decline it as unnecessary churn; nothing depends on it.

## Human judgment

Not required. Both fixes are mechanical and unambiguous, with exactly one correct form each and an existing in-repo convention to copy from. No design, product, or safety decision is involved.

## Considered and rejected

One nitpicker finding was assessed as likely invalid and is not filed:

**Claimed:** Tailwind class ordering is inconsistent across `empty-state.tsx:13-16` — "no consistent property-group order (spacing → sizing → typography → color) applied across the sibling elements."

**Actually:** every one of the four class lists already follows spacing → sizing → typography → color:
- `p-6 text-center` — spacing, typography
- `mb-2 text-[length:var(--text-h1)] text-foreground-faint` — spacing, type size, color
- `mb-1 text-sm font-medium text-foreground-secondary` — spacing, type size, weight, color
- `mx-auto max-w-[var(--size-content-narrow)] text-xs text-muted-foreground` — spacing, sizing, type size, color

**Why invalid:** the lists differ only in which property groups they populate, because the four elements set different properties. The relative order is uniform. There is also no `prettier-plugin-tailwindcss` in `frontend/package.json`, so no tool-enforced canonical order exists to be violated.

## Acceptance Criteria

- [ ] The icon element in `frontend/src/components/shared/empty-state.tsx` carries `aria-hidden="true"`.
- [ ] `frontend/src/components/shared/empty-state.test.tsx` asserts the icon is `aria-hidden="true"`.
- [ ] The default icon glyph is a named module-scope constant, or the item is explicitly declined as unnecessary churn.
- [ ] `cd frontend && npm run lint && npm test` passes.

NodeJSmith:
## Summary

`EmptyState` renders its icon as a bare text glyph, so screen readers read it aloud before the real message: "empty set", "leftwards arrow", "dotted circle". Every icon passed to `EmptyState` is decorative, since the `title` carries the meaning. The component has 15+ call sites, so the extra announcement showed up on every empty view in the dashboard.

- Added `aria-hidden="true"` to the icon `<div>` in `frontend/src/components/shared/empty-state.tsx`. Other decorative glyphs in the frontend already do this (`sort-header.tsx`, `status-shape.tsx`, `breadcrumbs.tsx`, the icon set).
- Moved the default `"∅"` glyph into a module-level `DEFAULT_EMPTY_ICON` constant so it can't be mistaken for `0`/`Ø`/`⌀` when reading the signature.
- Added a test in `empty-state.test.tsx` that checks the icon has `aria-hidden="true"`, the same kind of check `icons.test.tsx` and `filter-icon.test.tsx` already make.

Nothing changes visually. The glyph renders exactly as before and is only hidden from the accessibility tree.

## Testing

- `npx vitest run src/components/shared/empty-state.test.tsx`: 6/6 passed
- `prek -a`: all hooks passed (eslint, prettier, stylelint, tsc included)

CI runs the full backend, frontend, and e2e suites.

## Screenshots

Empty state on the app handlers tab after the change (looks the same as before):

EmptyState on app handlers tab

Closes #2165


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2203, opened 2026-09-11
test: consolidate shared setup and literals in execution-table tests

### LINKED ISSUE #2172: Clean up shared setup and duplicated literals in execution-table.test.tsx
## Description

A quality-scanner pass (llm-checker, lazy-checker, nitpicker) over
`frontend/src/components/shared/execution-table.test.tsx` surfaced a cluster of test-hygiene
findings. All predate the scan and are mechanical cleanups confined to this one file. The
llm-checker pass came back clean — the file has zero comments, correctly reuses the shared
`createExecution` / `createWouterMock` factories, and its tests are behavior-focused.

### 1. `mockNavigate.mockClear()` is called manually in 3 tests instead of a `beforeEach`

`mockNavigate` is a module-level `vi.fn()` (line 10) shared across the whole suite, but there is
no `beforeEach` anywhere in the file — only an `afterEach` for `vi.useRealTimers()`. Three tests
clear it by hand (lines 186, 204, 239), and the test at lines 101-114 renders with the same
navigation props without clearing. Any future test that asserts on `mockNavigate` and forgets the
manual call inherits call counts from earlier tests.

**Fix:** add `beforeEach(() => { mockNavigate.mockClear(); })` at suite level and drop the three
per-test calls.

### 2. The same execution-ID literal is duplicated 3x under two different names

`"abc12345-6789-abcd-ef01-234567890abc"` appears at lines 73, 187, and 240 — bound to `uuid` at
line 73 and `execId` at 187 and 240.

**Fix:** one module-scope constant (e.g. `const TEST_EXECUTION_ID = "..."`), used in all three.

### 3. The detail-route template is hardcoded twice, while the component derives it

Lines 199 and 255 both assert against `` `/apps/my_app/handlers/job/1/exec/${execId}` ``. The
component under test does not build this string inline — it calls `executionPath()` from
`../../utils/app-routes` (`execution-table.tsx:171`). If the route shape changes, one of these two
copies is easy to miss.

**Fix:** derive the expected path from `executionPath()` in the test, or hoist it into a single
shared expectation constant.

### 4. The handler-props trio is repeated verbatim 3x

`appKey="my_app"` / `handlerKind="job"` / `handlerId={1}` appears identically at lines 107-109,
193-195, and 246-248.

**Fix:** hoist into one shared props object spread into each render, or a small
`renderNavigableRow()` helper covering the render-plus-props setup these three tests share.

### 5. Two different idioms for the same row lookup

The same element is fetched as `container.querySelector("[data-testid='execution-row']")!` (lines
77, 161, 198) and as `getByTestId("execution-row")` (lines 94, 208, 252). The non-null assertion
form is the weaker of the two — it produces a null-deref rather than a descriptive
testing-library failure when the row is absent.

**Fix:** standardize on `getByTestId("execution-row")` for single-row lookups; keep
`querySelectorAll` only where a collection is genuinely needed (lines 50, 224).

### 6. Scattered epoch literals, with one implicit offset behind an assertion

The `1700000000`-family base timestamp is inlined independently at lines 45-47, 82-83, 173, 179,
and 220-221. In the timestamp test specifically, the 600-second gap between `now`
(`1_700_000_600`) and `executionStart` (`1_700_000_000`) is what produces the `"10m ago"`
assertion on line 97, but nothing states that relationship — a reader has to subtract to see it.

**Fix:** a single shared base constant, and derive `now` from it as
`BASE_TS + TEN_MINUTES_IN_SECONDS` (or an equivalent named offset) so the assertion follows from
the setup.

### 7. `tableId="t"` sits alongside descriptive table IDs

Three tests pass meaningful IDs (`"invocation-table-1"`, `"execution-table-1"`,
`"execution-table-99"` at lines 19, 24, 30) while the remaining ~17 pass a bare `"t"`. The mix
reads as accidental rather than as a deliberate "this ID is irrelevant here" convention.

**Fix:** use one shared constant for the tests where the ID is incidental, keeping the descriptive
IDs only where a test actually asserts on them (line 32).

## Acceptance Criteria

- [ ] `mockNavigate.mockClear()` runs from a suite-level `beforeEach`; the three per-test calls are removed
- [ ] The execution-ID literal is defined once at module scope and reused in all three tests
- [ ] The detail-route expectation is derived from `executionPath()` or a single shared constant, not hardcoded twice
- [ ] The `appKey`/`handlerKind`/`handlerId` trio is defined once and shared across the three navigation tests
- [ ] Single-row lookups consistently use `getByTestId("execution-row")` rather than a `querySelector(...)!` non-null assertion
- [ ] Epoch timestamps derive from one named base constant, and the `"10m ago"` test derives `now` from a named offset
- [ ] Incidental `tableId` values use one shared constant; descriptive IDs remain only where asserted on
- [ ] `cd frontend && npm run test` passes with no behavior change to the suite's assertions

NodeJSmith:
## Summary

Test-only cleanup of `frontend/src/components/shared/execution-table.test.tsx`, covering the seven findings from the quality-scanner pass in #2172. No test was added, removed, or had its assertions changed. Only the shared setup and literals moved.

- **`mockNavigate` reset:** a suite-level `beforeEach` now calls `mockNavigate.mockClear()`, and the three manual per-test calls are gone. Before this, the click-to-navigate test at the old lines 101–114 didn't clear the mock, and any new test that forgot the manual call would inherit call counts from earlier tests.
- **Execution ID:** one module constant, `TEST_EXECUTION_ID`, replaces the UUID literal that appeared three times under two names (`uuid`, `execId`).
- **Detail-route expectation:** one constant, `EXPECTED_DETAIL_PATH`, built from the shared props, replaces the two hardcoded `` `/apps/my_app/handlers/job/1/exec/${execId}` `` copies.
- **Handler props:** one `NAVIGABLE_PROPS` object replaces the `appKey`/`handlerKind`/`handlerId` trio that was repeated three times. The three navigation tests spread it in.
- **Row lookup:** single-row lookups now use `getByTestId("execution-row")` instead of `container.querySelector(...)!`, so a missing row gives a testing-library error instead of a null dereference. `querySelectorAll` is kept where the test needs a collection (row count, roving tabindex).
- **Timestamps:** all epoch literals now derive from `BASE_TS`. The relative-time test sets `now = BASE_TS + TEN_MINUTES_IN_SECONDS`, which makes it clear where the `"10m ago"` assertion comes from.
- **Table IDs:** tests where the ID doesn't matter now use `INCIDENTAL_TABLE_ID`. The descriptive `"execution-table-99"` stays only in the test that asserts on it. The two empty-state tests passed descriptive IDs without asserting on them, so they moved to the shared constant too.

## Testing

- `prek -a` passes (prettier, eslint, tsc, and the rest of the hooks).
- `npx vitest run src/components/shared/execution-table.test.tsx`: 22/22 pass.
- CI runs the full frontend and backend suites.

Closes #2172

### DIFF
--- a/frontend/src/components/shared/execution-table.test.tsx
+++ b/frontend/src/components/shared/execution-table.test.tsx
@@ -1,27 +1,38 @@
 import { render } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { afterEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { createExecution } from "../../test/factories";
 import { createWouterMock } from "../../test/mock-wouter";
 import { formatTimestamp } from "../../utils/format";
 import { ExecutionTable } from "./execution-table";
 
+const INCIDENTAL_TABLE_ID = "t";
+const TEST_EXECUTION_ID = "abc12345-6789-abcd-ef01-234567890abc";
+const BASE_TS = 1_700_000_000;
+const TEN_MINUTES_IN_SECONDS = 600;
+const NAVIGABLE_PROPS = { appKey: "my_app", handlerKind: "job", handlerId: 1 } as const;
+const EXPECTED_DETAIL_PATH = `/apps/${NAVIGABLE_PROPS.appKey}/handlers/${NAVIGABLE_PROPS.handlerKind}/${NAVIGABLE_PROPS.handlerId}/exec/${TEST_EXECUTION_ID}`;
+
 const mockNavigate = vi.fn();
 vi.mock("wouter", () => createWouterMock({ useLocation: () => ["/", mockNavigate] }));
 
 describe("ExecutionTable", () => {
+  beforeEach(() => {
+    mockNavigate.mockClear();
+  });
+
   afterEach(() => {
     vi.useRealTimers();
   });
 
   it("renders handler empty state when records are empty", () => {
-    const { getByText } = render(<ExecutionTable records={[]} kind="handler" tableId="invocation-table-1" />);
+    const { getByText } = render(<ExecutionTable records={[]} kind="handler" tableId={INCIDENTAL_TABLE_ID} />);
     expect(getByText("no invocations recorded")).toBeDefined();
   });
 
   it("renders job empty state when records are empty", () => {
-    const { getByText } = render(<ExecutionTable records={[]} kind="job" tableId="execution-table-1" />);
+    const { getByText } = render(<ExecutionTable records={[]} kind="job" tableId={INCIDENTAL_TABLE_ID} />);
     expect(getByText("no executions recorded.")).toBeDefined();
   });
 
@@ -33,7 +44,9 @@ describe("ExecutionTable", () => {
   });
 
   it("renders unified column headers", () => {
-    const { getByText } = render(<ExecutionTable records={[createExecution("job")]} kind="job" tableId="t" />);
+    const { getByText } = render(
+      <ExecutionTable records={[createExecution("job")]} kind="job" tableId={INCIDENTAL_TABLE_ID} />,
+    );
     expect(getByText("Status")).toBeDefined();
     expect(getByText("Execution")).toBeDefined();
     expect(getByText("Duration")).toBeDefined();
@@ -42,11 +55,11 @@ describe("ExecutionTable", () => {
 
   it("renders correct number of rows", () => {
     const records = [
-      createExecution("job", { execution_start_ts: 1700000001 }),
-      createExecution("job", { execution_start_ts: 1700000002 }),
-      createExecution("job", { execution_start_ts: 1700000003 }),
+      createExecution("job", { execution_start_ts: BASE_TS + 1 }),
+      createExecution("job", { execution_start_ts: BASE_TS + 2 }),
+      createExecution("job", { execution_start_ts: BASE_TS + 3 }),
     ];
-    const { container } = render(<ExecutionTable records={records} kind="job" tableId="t" />);
+    const { container } = render(<ExecutionTable records={records} kind="job" tableId={INCIDENTAL_TABLE_ID} />);
     expect(container.querySelectorAll("[data-testid='execution-row']").length).toBe(3);
   });
 
@@ -55,7 +68,7 @@ describe("ExecutionTable", () => {
       <ExecutionTable
         records={[createExecution("job", { status: "error", error_type: "ValueError", error_message: "Task failed" })]}
         kind="job"
-        tableId="t"
+        tableId={INCIDENTAL_TABLE_ID}
       />,
     );
     expect(container.textContent).toContain("failed");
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
Automated review pass: CodeRabbit (clean, 0 threads), Codex (clean), plus code-reviewer, integration-reviewer, wtf-reviewer, llm-checker, lazy-checker, and nitpicker.

**Verdict: no changes needed.** Zero CRITICAL/HIGH findings. All MEDIUM findings were verified and skipped, with reasons:

- **`EXPECTED_DETAIL_PATH` hand-builds the route instead of calling `executionPath()`** (lazy-checker, wtf-reviewer). Issue #2172's acceptance criterion allowed either form ("derived from `executionPath()` **or** a single shared constant"). The hardcoded constant is the stronger of the two here: deriving the expectation from the same helper the component calls (`execution-table.tsx:171`) would make the assertion tautological — a change to `executionPath()`'s output shape would silently keep the test green. Keeping an independent literal is what makes these two navigation tests able to fail.
- **`"execution-row"` testid repeated 8x, not extracted** (nitpicker). Checked against sibling tests: inline is the house convention for a testid the component itself renders — `log-table-view.test.tsx` repeats `"log-table"` 31x, `handlers-tab.job.test.tsx` repeats `"job-stats-row"` 30x, and 14 other `components/**/*.test.tsx` files repeat one 4x or more without extracting. The three files that *do* define a `*_TEST_ID` constant (`card.test.tsx`, `handler-list.test.tsx`, `registration-footer.test.tsx`) all pass the id *in* as a prop, which is a different case. Extracting here would make this the only diverging file.
- **`"execution-table-99"` left inline** (nitpicker). Deliberate per issue #2172 item 7 — descriptive IDs stay only where a test asserts on them. It appears twice inside one five-line test where the passed value and the asserted value being visibly identical *is* the test.

The remaining findings (llm-checker's over-extraction concerns about `BASE_TS`, `TEN_MINUTES_IN_SECONDS`, and `INCIDENTAL_TABLE_ID`) argue against extractions that issue #2172 items 6 and 7 explicitly requested; acting on them would revert the PR.

**Verification:** `npx vitest run src/components/shared/execution-table.test.tsx` → 22/22 pass. Full frontend suite → 110 files, 1610 tests pass. The `file-sizes` and `duplicate-code` CI failures are the known `main` baseline (both `continue-on-error`), not introduced here.

NodeJSmith:
Second automated review pass (re-run against the same head, `9c418b5`). No new CodeRabbit threads since the first pass — `gh-pr-threads` returns 0 threads, 0 review comments, CodeRabbit check green, Codex 👍.

**Verdict: no changes needed** (unchanged from the first pass). Re-ran all six reviewers independently:

| Reviewer | Result |
|---|---|
| code-reviewer | PASS — 0 findings. Traced every assertion against its pre-refactor counterpart; `BASE_TS + TEN_MINUTES_IN_SECONDS` resolves to exactly `1_700_000_600` as before, the suite-level `beforeEach` clear is a strict superset of the three removed per-test clears, and `getByTestId` preserves the `querySelector(...)!` failure semantics. |
| integration-reviewer | PASS — 0 findings. The module-constant and `getByTestId` patterns move this file *toward* the sibling convention (`card.test.tsx`, `badge.test.tsx`, `source-location.test.tsx`), not away from it. No shared factory exists that builds this route string. |
| llm-checker | CLEAN — every extracted constant has multiple real call sites; no dead abstractions. |
| lazy-checker | CLEAN — consolidation fully applied; `querySelectorAll` correctly retained where a NodeList is needed (row count, roving tabindex). |
| wtf-reviewer | 1 MEDIUM, 2 LOW — all skipped, see below. |
| nitpicker | 6 style findings — all skipped, see below. |

**Skipped findings, with reasons:**

- **`EXPECTED_DETAIL_PATH` moves the asserted path out of the test body** (wtf-reviewer, MEDIUM). This is the consolidation issue #2172 item 3 asked for. The constant is built from `NAVIGABLE_PROPS` + `TEST_EXECUTION_ID`, *not* from `executionPath()` — so the assertion stays independent of the production route builder (`execution-table.tsx:171`) and both navigation tests can still fail if that helper's output shape changes. Inlining it again would revert the PR.
- **`NAVIGABLE_PROPS` spread adds indirection** (wtf-reviewer, LOW) and **`TEST_EXECUTION_ID` is shared by two tests** (wtf-reviewer, LOW). Both are the requested extractions from issue #2172 items 2 and 4. Noted, not acted on.
- **`"execution-table-99"`, `"execution-id"`, and `"some-id"` left inline** (nitpicker, 3 findings). Verified each: `"execution-table-99"` appears twice inside one five-line test where the passed value and the asserted value being visibly identical *is* the test. `"execution-id"` and `"some-id"` are throwaway values in tests that never assert on them — folding them into `TEST_EXECUTION_ID` would create exactly the false coupling wtf-reviewer's own LOW finding warns about. lazy-checker independently verified all three as deliberate exceptions, not misses.
- **`BASE_TS` abbreviation, `TEST_` prefix consistency, `INCIDENTAL_TABLE_ID` naming** (nitpicker, 3 findings). `BASE_TS` matches the domain field it feeds (`execution_start_ts`); llm-checker and wtf-reviewer both read `INCIDENTAL_TABLE_ID` as a clarity win. Style preference with no functional impact.

**CI triage.** `test (3.13)` failed on `tests/unit/test_execution_mode_helpers.py::TestRunThroughGuard::test_dropped_at_drain_time_resolves_its_own_future` with a `TimeoutError`. Not attributable to this PR — the entire diff against `origin/main` is one file, `frontend/src/components/shared/execution-table.test.tsx`, with zero Python changes. The test passed on 3.11, 3.12, and 3.14 in the same run, passes locally (19/19 in that file), and is the known real-clock flake already tracked by #2178 (2.0s wall-clock waits under `-n 4`). Rerun triggered. `file-sizes` and `duplicate-code` are the known `main` baseline, both `continue-on-error`.

**Verification:** `npx vitest run src/components/shared/execution-table.test.tsx` → 22/22 pass locally.

PR #2204, opened 2026-09-11
test: extract repeated temp app-dir setup in TestValidateApps into a fixture

### LINKED ISSUE #2173: Extract the repeated temp-app-directory setup in TestValidateApps
## Description

`duplicate-code` CI (PMD CPD) flagged a block repeated across `TestValidateApps` in
`tests/unit/test_validate_apps.py`.

**The shared shape:** every test in the class opens with the same two-line temp-app-directory
preamble, then hands that directory straight to the class's own `make_config()` helper:

```python
app_dir = tmp_path / "test_apps"
app_dir.mkdir(parents=True, exist_ok=True)

config = self.make_config(
    tmp_path,
    directory=app_dir,
    apps={...},
)
```

The `apps={...}` payload is the only part that actually differs between tests — the directory
construction is pure scaffolding repeated verbatim. `make_config()` already exists to absorb
config-building boilerplate; the directory setup that always precedes it never got the same
treatment, so each new test in the class copies it again.

## Goal

Extract the temp-app-directory setup into a shared pytest fixture (e.g. an `app_dir` fixture
returning the created directory) and use it at every call site with this shape, so a test in this
class declares only the `apps={...}` payload it is actually exercising.

Any other instance of the same shape found while doing the work is in scope too — not just the
locations listed below.

**Constraint for whoever picks this up:** the directory name `test_apps` is load-bearing, not
arbitrary. `autodetect_apps()` passes `app_dir.name` as the package name, so discovered-app keys
in `test_validate_apps_merges_autodetected_and_manual_apps` and
`test_validate_apps_skips_conflicting_autodetected` are asserted as literals like
`"test_apps.validate_auto_app.ValidateAutoApp"`. The extraction must preserve the name (the file
already documents this in an inline comment). Prefer a fixture over folding `mkdir` into
`make_config()` — the latter would silently change behavior for any future test that means to pass
a directory that does not exist.

## Where

- `tests/unit/test_validate_apps.py:41-46`
- `tests/unit/test_validate_apps.py:82-88`
- `tests/unit/test_validate_apps.py:241-247`
- `tests/unit/test_validate_apps.py:274-280`
- `tests/unit/test_validate_apps.py:305-311`
- `tests/unit/test_validate_apps.py:333-339`

Found while reading around the flagged lines — same shape, also in scope:

- `tests/unit/test_validate_apps.py:62-63` (`test_validate_apps_preserves_existing_app_dir`)
- `tests/unit/test_validate_apps.py:129` (`..._merges_autodetected_and_manual_apps` — `write_app`
  creates the directory instead of `mkdir`, so this site needs the fixture's path but not its
  `mkdir`)
- `tests/unit/test_validate_apps.py:191` (`..._skips_conflicting_autodetected` — same `write_app`
  variant)

That is 9 occurrences of `app_dir = tmp_path / "test_apps"` in a single 350-line file, 7 of them
followed by the identical `mkdir` line.

## Acceptance Criteria

- [ ] Temp-app-directory setup lives in one shared fixture rather than being repeated per test.
- [ ] Every test in `TestValidateApps` uses it, including the two `write_app` variants.
- [ ] The `test_apps` directory name is preserved, and the derived-key literal assertions in
      `..._merges_autodetected_and_manual_apps` / `..._skips_conflicting_autodetected` still pass
      unchanged.
- [ ] `uv run pytest tests/unit/test_validate_apps.py` passes.

NodeJSmith:
## Summary

Every test in `TestValidateApps` (`tests/unit/test_validate_apps.py`) started with the same temp-app-directory setup: `app_dir = tmp_path / "test_apps"`, and in seven of the nine tests an identical `mkdir` line after it. PMD CPD flagged this in the `duplicate-code` CI job.

This PR moves that setup into a module-level `app_dir` fixture that creates and returns `tmp_path / "test_apps"`. All nine tests in the class now take it as a parameter. That includes the two autodetect tests that write app files with `write_app()`. Their setup never called `mkdir` itself, and the fixture's `exist_ok=True` means the directory already existing doesn't matter to them.

The `test_apps` name has to stay. `autodetect_apps()` passes `app_dir.name` as the package name, so the key literals asserted in the autodetect tests (e.g. `"test_apps.validate_auto_app.ValidateAutoApp"`) depend on it. The comment explaining this moved from the first autodetect test into the fixture's docstring, and both autodetect tests now point to the fixture instead.

I used a fixture rather than putting the `mkdir` inside `make_config()`. Otherwise a future test that means to pass a directory that doesn't exist would get one created without noticing.

This only touches tests. Framework behavior is unchanged.

## Testing

- `uv run pytest tests/unit/test_validate_apps.py`: 9 passed
- `prek -a` and `prek pyright -a --stage pre-push`: all hooks passed

I only ran the affected test module locally. CI runs the full suite.

Closes #2173

### DIFF
--- a/tests/unit/test_validate_apps.py
+++ b/tests/unit/test_validate_apps.py
@@ -18,6 +18,20 @@ from hassette.testing import make_test_config
 from tests.support.helpers import write_app
 
 
+@pytest.fixture
+def app_dir(tmp_path: Path) -> Path:
+    """Return an existing, empty app directory under `tmp_path`.
+
+    The `test_apps` name is load-bearing, not scaffolding: autodetect_apps() passes
+    `app_dir.name` as the package name, so a discovered app's key is
+    "{app_dir.name}.{module_stem}.{class_name}" — the key literals asserted in the
+    autodetect tests below change if this directory is renamed.
+    """
+    path = tmp_path / "test_apps"
+    path.mkdir(parents=True, exist_ok=True)
+    return path
+
+
 class TestValidateApps:
     """Test the validate_apps function."""
 
@@ -36,10 +50,8 @@ class TestValidateApps:
             apps_cfg["apps"] = apps
         return make_test_config(data_dir=tmp_path, apps=apps_cfg)
 
-    def test_validate_apps_sets_app_dir(self, tmp_path: Path) -> None:
+    def test_validate_apps_sets_app_dir(self, tmp_path: Path, app_dir: Path) -> None:
         """Test that validate_apps sets app_dir for apps that don't have it."""
-        app_dir = tmp_path / "test_apps"
-        app_dir.mkdir(parents=True, exist_ok=True)
         config = self.make_config(
             tmp_path,
             directory=app_dir,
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #2205, opened 2026-09-11
chore(ui): share column-width and status-label class constants in execution table

### LINKED ISSUE #2175: Clean up duplicated class-string literals in execution-table.tsx
Surfaced by an automated code-quality scan of `frontend/src/components/shared/execution-table.tsx` (llm-checker, lazy-checker, nitpicker). All findings are pre-existing and mechanical — none change rendered behavior.

Two of the three checkers returned CLEAN. The comments in this file were checked specifically for LLM comment bloat and history narration and came back clean — they explain non-obvious cross-file invariants and should be left alone.

## 1. Column width literals are duplicated between `headerClassName` and `cellClassName`

The table renders with `table-fixed`, so a column's header width and cell width must stay identical. Today each width is typed twice, as an unnamed literal, with nothing enforcing that the pair stays in sync:

- `:93` — `w-[18%] max-mobile:w-auto` in both `headerClassName` and `cellClassName` (status column)
- `:128-129` — `w-[14%] max-mobile:w-auto` / `w-[14%] whitespace-nowrap max-mobile:w-auto` (duration column)
- `:137-138` — `w-[18%] max-mobile:w-auto` twice (time column)

The status and time columns additionally share the same `w-[18%]` value across two unrelated column definitions, so a proportion change means editing four sites.

**Fix:** hoist the widths to named module constants next to `HEAD_CLASS`/`CELL_CLASS` (e.g. `const COL_W_STATUS = "w-[18%] max-mobile:w-auto";`) and reference them from both `headerClassName` and `cellClassName`. These are table-layout proportions, not design-system spacing, so local constants are the right level — do not push them into CSS custom properties.

## 2. `cn()` wrapping a single static string is a no-op

`:120` and `:138` wrap a static class string in `cn(...)`:

```ts
cellClassName: cn("font-mono text-xs [overflow-wrap:anywhere] max-mobile:hidden"),
cellClassName: cn("w-[18%] font-mono text-xs whitespace-nowrap max-mobile:w-auto"),
```

Neither string contains conflicting utilities, so `cn()` returns it unchanged, and `ExecutionCell` already runs the value through `cn(CELL_CLASS, meta?.cellClassName)` at `:209` where any real merging happens. Every other `headerClassName`/`cellClassName` in the same `columns` array is a plain string literal, so this reads as if those two entries do something special when they don't.

**Fix:** drop the `cn()` wrapper on both, leaving plain string literals consistent with the sibling entries.

## 3. `statusLabelClass` repeats the same base classes in every branch

`:67-84` — `font-mono text-xs whitespace-nowrap` is retyped in all five switch branches; only the color token varies (plus `truncate` on `err`). The same three-class run appears again at `:138`.

**Fix:** extract a `const STATUS_TEXT_BASE = "font-mono text-xs whitespace-nowrap";` and compose per-branch via `cn(STATUS_TEXT_BASE, ...)`, returning only the varying part from each branch. Keep the `switch` — `@typescript-eslint/switch-exhaustiveness-check` is on, and the exhaustiveness guarantee over `StatusKind` is worth preserving.

## 4. Minor: single-letter state updater parameter

`:346` — `setShowAll((v) => !v)`. Every other callback in the file uses a contextual name (`record`, `cell`, `row`, `header`). Rename to `prev` or `showAll`.

## Not filed (checked and rejected)

Two findings from the scan were dropped after checking `frontend/eslint.config.js`:

- **"Mixed `@/` and `../../` import styles"** — `no-restricted-imports` explicitly permits two-level relative imports and requires `@/` only at 3+ levels ("Use the @/ path alias instead of deep relative imports (3+ levels)"). This file conforms to the configured rule; it is not drift.
- **"Missing blank line between the `@tanstack` and `react`/`wouter` imports"** — all three are third-party and correctly share one group; `simple-import-sort/imports` is an `error`-level rule and the file passes it.

NodeJSmith:
## Summary

Mechanical cleanup of duplicated class-string literals in `frontend/src/components/shared/execution-table.tsx`, addressing the code-quality scan findings in #2175. The rendered class set for every element is unchanged.

- **Column widths are shared constants.** The table uses `table-fixed`, so each column's header and cell width have to match. The status, duration, and time widths are now `COL_W_STATUS` / `COL_W_DURATION` / `COL_W_TIME`, and both `headerClassName` and `cellClassName` read from the same constant instead of repeating the literal.
- **Removed `cn()` calls that did nothing.** The execution column's `cellClassName` wrapped one static string in `cn()`. It's now a plain literal like its siblings. The time column still uses `cn()`, but now it actually combines two values (the width constant and the per-cell classes).
- **Base status-label classes are declared once.** `statusLabelClass` became `statusKindClass` and returns only the classes that differ per kind (color, plus `truncate` for `err`, `undefined` for `mute`). The call site combines them with `STATUS_TEXT_BASE` (`font-mono text-xs whitespace-nowrap`) via `cn()`. The `switch` stays so `switch-exhaustiveness-check` still covers `StatusKind`.
- **Renamed the show-all updater param** from `v` to `prev`.

## Testing

- `prek -a`: all hooks pass (eslint, prettier, stylelint, tsc, ruff, the house checks).
- `npx vitest run src/components/shared/execution-table.test.tsx`: 22/22 pass.

CI runs the full backend, system, e2e, and frontend suites.

## Screenshots

These show the execution table after the change, with identical rendering.

**Failed execution row**

Execution table, failed row

**Cancelled execution row**

Execution table, cancelled row

Closes #2175


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

Issue #2163, opened 2026-09-09
Extract the shared guard-priming scaffolding in run_through_guard unit tests


NodeJSmith:
## Description

Flagged by the `duplicate-code` CI job (PMD CPD, `tools/check_duplicate_code.py`) as a block
copy-pasted 3+ times in `tests/unit/test_execution_mode_helpers.py`. None of the flagged lines
carry a `dup-ignore-*` marker.

The shared shape is the **"prime the guard with one running + one queued invocation" arrange
block** — the fixed preamble every `TestRunThroughGuard` test that needs a non-empty
`guard.pending` performs before it can assert anything:

```python
gate = asyncio.Event()
started = asyncio.Event()

async def invoke_running() -> None:
    started.set()
    await gate.wait()

async def invoke_queued() -> None:
    ...

first = asyncio.create_task(
    run_through_guard(guard, spawn, pending_done, invoke_running, MagicMock(), "r", 60.0)
)
await asyncio.wait_for(started.wait(), timeout=2.0)

second = asyncio.create_task(
    run_through_guard(guard, spawn, pending_done, invoke_queued, MagicMock(), "q", 60.0)
)
await wait_for(lambda: len(guard.pending) >= 1)
```

Everything the tests actually differ on happens *after* this block — dropping a third call at the
cap, sealing `spawn` so the drained factory's spawn is rejected, or calling
`guard.release()` + `drain_pending_done()`. The ~11 lines of gate/started plumbing and the two
deterministic waits in front of that are identical in each copy, including the "wait until the
queued factory is parked" idiom that exists specifically so these tests don't race.

Two more tests in the same class repeat the running-invocation half of the same block
(`gate`/`started` + a first `run_through_guard` task + `await asyncio.wait_for(started.wait())`)
without queueing a second call — they are part of the same cluster and should be covered by the
same helper if it fits without contorting them.

## Fix

Extract the shared guard-priming scaffolding into one reusable helper and use it at every call
site with this shape, so each test declares only the invocation bodies it cares about and the
assertion it is actually making — not the gate/started/park sequence in front of it. A natural
shape is a helper that returns the running task, the queued task, and the gate that releases the
running invocation, with the queued half optional so the two running-only tests can share it.

Per `.claude/rules/test-conventions.md`, check `tests/support/factories.py` and
`tests/support/helpers.py` for an existing helper before adding a new name, and prefer putting the
[...]

PR #2207, opened 2026-09-11
test: fix flaky real-clock hang guards in TestRunThroughGuard

### LINKED ISSUE #2178: Fix flaky real-clock timeouts in TestRunThroughGuard unit tests
## Description

`tests/unit/test_execution_mode_helpers.py::TestRunThroughGuard::test_dropped_at_drain_time_resolves_its_own_future` failed with a bare `TimeoutError` on CI (`test (3.12)`, [run 34501971595](https://github.com/NodeJSmith/hassette/actions/runs/34501971595/job/102954813543)) while `test (3.11)`, `test (3.13)`, and `test (3.14)` passed on the same commit. The test passes locally and the failure appeared on a PR (#2097) whose diff is confined to `tests/support/factories.py`, `tests/unit/test_logging_correlation.py`, and three doc/tooling files — `tests/unit/test_execution_mode_helpers.py` and `src/hassette/execution_mode.py` are byte-identical to `origin/main` on that branch. This is a pre-existing flake, not a regression.

### Mechanism

The `TestRunThroughGuard` tests bound every step of a deterministic event handshake with a **real wall-clock** budget:

```python
await asyncio.wait_for(started.wait(), timeout=2.0)
await wait_for(lambda: len(guard.pending) >= 1)   # tests.support wait_for default: 3.0s, 0.02s poll
...
await asyncio.wait_for(first, timeout=2.0)
await asyncio.wait_for(second, timeout=2.0)
```

The synchronization itself is correct — `asyncio.Event` gates, not `asyncio.sleep(0)` ticks — so these timeouts are only there as a hang guard, never as the thing under test. But the suite runs under `pytest-xdist`, so a 2-second budget is competing with worker scheduling on a shared CI runner. Under load that budget is reachable, which is exactly the real-clock-vs-CI-jitter collision `CLAUDE.md` (Regression test patterns → *Config-driven real-clock timeouts*) already documents: the deliberate hold and the timeout are two independent real-time clocks, and CI's noisier scheduling is what finally exposes the collision after the test has passed locally for months.

Every `asyncio.wait_for(...)` in the class uses `timeout=2.0`; the six `run_through_guard(...)` calls separately pass a `60.0` timeout argument, so the 2.0s values are purely the test's own hang guard and carry no assertion weight.

## Steps to Reproduce

Not reliably reproducible on demand — it is a scheduling-load flake. To force the failure signature:

```bash
# drive the hang guard low enough to lose the race on any machine
uv run pytest tests/unit/test_execution_mode_helpers.py -k TestRunThroughGuard
# after temporarily editing the asyncio.wait_for timeouts in that class from 2.0 to 0.0
```

## Expected Behavior

`TestRunThroughGuard` passes deterministically regardless of CI scheduling load. Its timeouts exist only to convert a genuine hang into a failure rather than a suite-wide stall.

## Actual Behavior

`test_dropped_at_drain_time_resolves_its_own_future` intermittently fails with `TimeoutError` on a loaded runner, blocking `tests-complete` on unrelated PRs and requiring a manual re-run.

## Acceptance Criteria

- [ ] Confirm the failure signature reproduces when the class's `asyncio.wait_for` hang guards are driven to `0.0`, ruling out a real deadlock in `ExecutionModeGuard.drain_next`.
- [ ] Widen the hang-guard timeouts in `TestRunThroughGuard` so they carry generous headroom over the deterministic handshake they guard (the `wait_for` helper's own `3.0`s default is the natural floor; do not shrink the handshake instead — the handshake is the thing under test).
- [ ] Apply the change to every `asyncio.wait_for(..., timeout=2.0)` in the class, not just the one call site that failed, so the next-loaded-runner failure doesn't just move to a sibling test.
- [ ] `uv run pytest tests/unit/test_execution_mode_helpers.py -n 4` passes.

### Note for whoever picks this up

#2163 already proposes extracting the shared "prime the guard with one running + one queued invocation" arrange block from this same class. That extraction is the natural place to land this fix — one helper owning the handshake means one timeout constant instead of six copies. Worth doing together rather than touching the same six call sites twice.

NodeJSmith:
## Summary

`TestRunThroughGuard::test_dropped_at_drain_time_resolves_its_own_future` intermittently failed on CI with a bare `TimeoutError` on a PR whose diff didn't touch this test or `execution_mode.py`. The cause was the class's own hang guards. Every `asyncio.wait_for(..., timeout=2.0)` put a 2-second real-clock budget on an event handshake that normally finishes in about 20ms. Under xdist on a loaded CI runner, that budget is sometimes reachable.

The timeouts only turn a real hang into a failure; they aren't the thing being tested. The `run_through_guard(...)` calls pass their own `60.0` timeout separately.

## Changes

- Added a module-level `HANG_GUARD_TIMEOUT = 10.0` in `tests/unit/test_execution_mode_helpers.py`.
- Replaced every `asyncio.wait_for(..., timeout=2.0)` in `TestRunThroughGuard` with it, including sibling tests that haven't flaked yet. Otherwise the next loaded runner would just move the failure to one of them.
- Passed the same constant to the three `wait_for(lambda: len(guard.pending) >= 1)` helper calls that had been using the helper's 3.0s default.

The handshakes themselves are unchanged; they're what the tests exercise.

## Testing

- Set `HANG_GUARD_TIMEOUT` to `0.0` and confirmed all five handshake tests fail with the same bare `TimeoutError` CI showed. This shows the constant controls every guard and that the CI failure came from a guard firing, not a deadlock in `ExecutionModeGuard.drain_next`.
- `uv run pytest tests/unit/test_execution_mode_helpers.py -n 4`: 19 passed.
- `prek -a` and `prek pyright -a --stage pre-push` both pass.

The full backend, system, and e2e suites run in CI.

## Notes

#2163 proposes pulling the shared "one running + one queued invocation" arrange block in this class into a helper. When that lands, the helper can own this constant, so the timeout lives in one place.

Closes #2178

### DIFF
--- a/tests/unit/test_execution_mode_helpers.py
+++ b/tests/unit/test_execution_mode_helpers.py
@@ -20,9 +20,11 @@ from hassette.execution_mode import (
 from hassette.testing import wait_for
 from hassette.types.enums import ExecutionMode
 
-#: Hang guard for TestRunThroughGuard's event handshakes, which complete in milliseconds. It only
-#: converts a genuine hang into a failure, so it needs generous headroom over xdist scheduling
-#: jitter on a loaded CI runner — well above the ``wait_for`` helper's own 3.0s default.
+#: Hang guard for this module's event handshakes, which complete in milliseconds. It only converts
+#: a genuine hang into a failure, so it needs generous headroom over xdist scheduling jitter on a
+#: loaded CI runner — well above the ``wait_for`` helper's own 3.0s default. Distinct from the
+#: ``threshold`` argument passed to ``run_with_stall_watch``/``run_through_guard``, which is the
+#: production stall-warning value under test rather than a test-side wait budget.
 HANG_GUARD_TIMEOUT = 10.0
 
 
@@ -85,10 +87,10 @@ class TestRunWithStallWatch:
 
         threshold = 0.05  # short threshold for testing
         task = asyncio.create_task(run_with_stall_watch(invoke, warn, threshold=threshold))
-        await asyncio.wait_for(started.wait(), timeout=2.0)
+        await asyncio.wait_for(started.wait(), timeout=HANG_GUARD_TIMEOUT)
         # warn.set() unblocks the gate, so awaiting the task is the deterministic
         # signal that the watchdog fired.
-        await asyncio.wait_for(task, timeout=2.0)
+        await asyncio.wait_for(task, timeout=HANG_GUARD_TIMEOUT)
 
         assert warn_calls == [threshold], f"expected warn called with {threshold}, got {warn_calls}"
 
@@ -120,7 +122,7 @@ class TestRunWithStallWatch:
             await asyncio.sleep(10)  # will be cancelled
 
         task = asyncio.create_task(run_with_stall_watch(invoke, warn, threshold=0.05))
-        await asyncio.wait_for(started.wait(), timeout=2.0)
+        await asyncio.wait_for(started.wait(), timeout=HANG_GUARD_TIMEOUT)
         task.cancel()
         with pytest.raises(asyncio.CancelledError):
             await task
--- a/tests/unit/test_execution_mode_helpers.py
+++ b/tests/unit/test_execution_mode_helpers.py
@@ -20,6 +20,11 @@ from hassette.execution_mode import (
 from hassette.testing import wait_for
 from hassette.types.enums import ExecutionMode
 
+#: Hang guard for TestRunThroughGuard's event handshakes, which complete in milliseconds. It only
+#: converts a genuine hang into a failure, so it needs generous headroom over xdist scheduling
+#: jitter on a loaded CI runner — well above the ``wait_for`` helper's own 3.0s default.
+HANG_GUARD_TIMEOUT = 10.0
+
 
 def make_spawn() -> tuple[Callable[..., asyncio.Task[None]], list[asyncio.Task[None]]]:
     """Return a spawn callable (wrapping create_task) and the list of tasks it created."""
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

NodeJSmith:
Review pass (code / integration / WTF + llm / lazy / nitpicker) — one finding landed, everything else verified clean.

**Applied in d2fc06ad:** four of the six reviewers independently flagged that `TestRunWithStallWatch`, two classes up in the same file, guards the same `asyncio.Event` handshake with the same bare `timeout=2.0` this PR just replaced everywhere else. That's the PR's own justification for converting sibling tests rather than only the one that flaked — it just stopped at a class boundary instead of a risk boundary, and left the file with two spellings of one value. Converted those three call sites, widened the constant's comment to match its now file-wide reach, and added a clause disambiguating it from the `threshold` argument these helpers take (a production stall-warning value under test, not a test wait budget). 19 passed, `prek` clean.

**Not applied — `HANG_GUARD_TIMEOUT` → `HANG_GUARD_TIMEOUT_SECONDS`:** the `_SECONDS` suffix is the convention inside `tests/support/helpers.py`, not repo-wide. Across `tests/` it's 26 bare `*_TIMEOUT` constants to 14 with the suffix, and the file-local timing constants in `tests/unit/` are mostly bare (`WORKER_READY_TIMEOUT`, `JOIN_BUDGET`, `GENEROUS_JOIN_TIMEOUT`). The current name matches its neighborhood.

**Not applied — move the constant into `tests/support/helpers.py`:** single consumer today, and #2163 is already slated to give this class a shared arrange helper that can own it.

**Filed #2208** for the one finding genuinely outside this PR's scope: the same `asyncio.wait_for(..., timeout=2.0)` hang-guard idiom appears ~32 more times across the integration suite, including `tests/integration/bus/test_execution_modes_guards.py`, which exercises the same `ExecutionModeGuard` machinery at the integration layer.

PR #2209, opened 2026-09-11
test: extract deploy-and-wait app setup in hot reload integration tests

### LINKED ISSUE #2191: Extract the repeated deploy-and-wait app setup in hot reload integration tests
## Description

The `duplicate-code` CI job (PMD CPD) flagged a block repeated across `TestBasicHotReload` in
`tests/integration/test_hot_reload.py`.

**The shared shape:** every hot-reload test first "deploys one app and waits for it to run." That
means building a manifest, writing its app module, writing it into the TOML, and announcing both
files as changed:

```python
app1 = create_app_manifest(suffix="<suffix>", app_dir=self.app_dir, enabled=True)
write_test_app_with_decorator(app_file=app1.full_path, class_name=app1.class_name)

write_app_toml(self.toml_file, app_dir=self.app_dir, apps=[app1])
await emit_change_and_wait_for_app_status(self.hassette, {self.toml_file, app1.full_path}, app1.app_key)
```

Only the suffix (and, in one test, `app_config=` / `config_fields=`) differs between copies. The
part each test actually cares about is what happens *after* the app is running: disabling it,
changing its config, or rewriting its module. `emit_change_and_wait_for_app_status` already
absorbed the emit-and-await half of this sequence. The manifest/module/TOML writes that always come
before it never got the same treatment, so every new test in the class copies them again.

## Goal

Extract the shared "deploy one app and wait until it's RUNNING" arrange step into a helper and use
it at every call site with this shape. The helper should take the suffix plus optional
`app_config` / `config_fields` and return the `AppManifest`, so each test is left with only the
reload behavior it exercises. The only consumer is this file, so the helper can live there (a
method on `TestBasicHotReload` or a module-level function). Don't add it to
`tests/support/helpers.py` unless a second consumer shows up. Avoid a `make_*`/`build_*` name
unless it goes through the `check_test_factories.py` convention.

Any other instance of the same shape found while doing the work is in scope too, not just the
locations listed below.

## Where

- `tests/integration/test_hot_reload.py:65-69` (`test_hot_reload_starts_newly_enabled_app`)
- `tests/integration/test_hot_reload.py:77-81` (`test_hot_reload_stops_newly_disabled_app`)
- `tests/integration/test_hot_reload.py:125-129` (`test_hot_reload_reimports_app_when_file_changes`)

Found while reading around the flagged lines. Same shape, also in scope:

- `tests/integration/test_hot_reload.py:97-105` (`test_hot_reload_reloads_app_with_config_change`):
  the same four steps, with `app_config={"test_value": "initial"}` on the manifest and
  `config_fields={"test_value": "str"}` on the module write. PMD missed it only because the
  keyword arguments break the token run. The helper's optional parameters should cover this site.

Adjacent duplication in the same file, fine to fold into the same change:

- `TestBasicHotReload.setup` (lines 57-61) and `TestOnlyAppsConfigFilter.setup` (lines 156-160)
  are identical autouse fixtures that unpack `hassette_and_handler` and derive `app_dir` /
  `toml_file`. They could share one base class or fixture.

`TestOnlyAppsConfigFilter.test_only_apps_starts_named_apps_and_blocks_the_rest` (lines 164-180)
deploys *three* apps with an env var set between the module writes and the TOML write, so it is a
different shape. Leave it alone unless the helper naturally generalizes to a list of manifests
without getting more complicated.

## Acceptance Criteria

- [ ] The deploy-and-wait-RUNNING arrange step lives in one helper instead of being repeated per test.
- [ ] All four `TestBasicHotReload` tests use it, including the config-change variant.
- [ ] Each test's post-deploy assertions and reload steps are unchanged.
- [ ] `uv run pytest tests/integration/test_hot_reload.py` passes.

NodeJSmith:
## Summary

Every `TestBasicHotReload` test repeated the same four-step setup before doing the thing it actually tests: build a manifest, write the app module, write the TOML entry, then announce both files and wait for the app to reach RUNNING. The PMD `duplicate-code` job flagged three copies. A fourth copy, in the config-change test, got past PMD only because its extra keyword arguments broke up the token run.

- Added `TestBasicHotReload.deploy_running_app(suffix, *, app_config=None, config_fields=None) -> AppManifest`, which runs that setup once. All four `TestBasicHotReload` tests call it now, including `test_hot_reload_reloads_app_with_config_change`, which passes `app_config` and `config_fields`.
- `TestBasicHotReload` and `TestOnlyAppsConfigFilter` had identical autouse `setup` fixtures. That fixture now lives in a shared `HotReloadTestBase` class.
- The assertions and reload steps that follow the setup in each test are unchanged. `test_only_apps_starts_named_apps_and_blocks_the_rest` sets an env var between deploying its three apps, so its setup has a different shape. The issue said to leave it alone, and it is.

The helper stays in this file because nothing else uses it. Its name isn't `make_*`/`build_*`, so the `check_test_factories.py` convention doesn't apply.

## Testing

- `prek -a`: all hooks pass
- `prek pyright --stage pre-push --files tests/integration/test_hot_reload.py`: passes
- `uv run pytest tests/integration/test_hot_reload.py -v`: 5 passed

This was a scoped local check. CI runs the full suite.

Closes #2191

### DIFF
--- a/tests/integration/test_hot_reload.py
+++ b/tests/integration/test_hot_reload.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING
 
 import pytest
 
+from hassette.config.classes import AppManifest
 from hassette.testing import HassetteHarness
 from hassette.types import ResourceStatus
 from tests.support.harness import preserve_config
@@ -46,8 +47,8 @@ def hassette_and_handler(
             shutil.rmtree(f)
 
 
-class TestBasicHotReload:
-    """Basic hot reload functionality tests."""
+class HotReloadTestBase:
+    """Shared setup for hot reload test classes: the harness, app handler, app dir, and TOML file."""
 
     hassette: HassetteHarness
     app_handler: "AppHandler"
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

PR #2212, opened 2026-09-11
test: enforce breadcrumb visible-crumb count against its CSS selectors

### LINKED ISSUE #2196: Enforce the breadcrumb visible-crumb count instead of documenting it
## Description

A post-merge review of PR #2193 (which landed `SIDEBAR_VISIBLE_CRUMBS` in `frontend/src/components/shared/breadcrumbs.tsx`) surfaced follow-up findings. These predate nothing — they describe the state of the file as merged on `main` — and are small, mechanical cleanups rather than a rework of that PR.

PR #2193 correctly named the bare `2` in the ellipsis check and documented that the two Tailwind `nth-last-child` selectors must hard-code the same value, because Tailwind extracts class names statically and cannot interpolate a JS constant. That documentation is accurate (the CSS semantics were independently verified). What's missing is anything that *enforces* the agreement, plus one remaining untracked copy of the value in prose.

## Key Items

- **`frontend/src/components/shared/breadcrumbs.tsx:6, 51-52` — nothing catches constant/selector drift.** `SIDEBAR_VISIBLE_CRUMBS = 2` is read only by the JS check at line 33. The two selectors — `max-sidebar:[&:not(:nth-last-child(-n+2))]:sr-only` and `max-sidebar:[&:nth-last-child(2)>span]:hidden` — hold independent literal `2`s. Someone changing the constant to 3 and not both selectors gets a silent visual bug at the sidebar breakpoint: the ellipsis threshold moves, but the clipping and separator-hiding don't. No test, type check, or lint rule fails. A test asserting the rendered class string contains ``​`-n+${SIDEBAR_VISIBLE_CRUMBS}`​`` and ``​`nth-last-child(${SIDEBAR_VISIBLE_CRUMBS})`​`` would turn the comment into an enforced invariant. This requires exporting the constant.

- **`frontend/src/components/shared/breadcrumbs.tsx:22` — the component JSDoc still spells the count in prose.** It reads "all but the last two crumbs are visually clipped by CSS," a fourth untracked copy of the value alongside the constant and the two selectors. It carries no comment tying it to the constant, so it goes stale silently. Reword to reference `SIDEBAR_VISIBLE_CRUMBS` by name, or drop the number ("all but the last few crumbs — see `SIDEBAR_VISIBLE_CRUMBS`").

- **(minor) `frontend/src/components/shared/breadcrumbs.tsx:5, 46-50` — the same point is made twice.** The constant's doc comment ("The nth-last-child selectors below hard-code this value") and the inline comment at the usage site both open with the hard-coding claim. If the drift test above lands, the declaration-site comment can shrink to just what it uniquely says, leaving the usage-site comment as the single explanation of the Tailwind constraint.

## Acceptance Criteria

- [ ] A test in `breadcrumbs.test.tsx` fails if `SIDEBAR_VISIBLE_CRUMBS` is changed without updating both `nth-last-child` literals.
- [ ] The `Breadcrumbs` JSDoc no longer hard-codes the visible-crumb count as a spelled-out number.
- [ ] The declaration-site and usage-site comments no longer redundantly state the same hard-coding point.
- [ ] `npx vitest run` and `prek -a` pass.

NodeJSmith:
## Summary

`SIDEBAR_VISIBLE_CRUMBS` sets the breadcrumb ellipsis threshold in JS. The two Tailwind `nth-last-child` selectors that clip crumbs and hide the leading separator below the sidebar breakpoint can't read that constant, because Tailwind extracts class names statically. They carry their own literal `2`s, and until now only a comment said they had to match. If someone changed the constant without changing both selectors, the ellipsis threshold would move while the clipping stayed put, and nothing would fail.

This PR adds a check that fails in that case:

- **Export `SIDEBAR_VISIBLE_CRUMBS`** from `breadcrumbs.tsx` so the test can import it.
- **New test in `breadcrumbs.test.tsx`**: it renders a trail and checks that the crumb's class string contains both `nth-last-child(-n+${SIDEBAR_VISIBLE_CRUMBS})` and `nth-last-child(${SIDEBAR_VISIBLE_CRUMBS})`. It fails if the constant or either selector changes on its own.
- **Comment cleanup**: the component JSDoc now refers to `SIDEBAR_VISIBLE_CRUMBS` instead of saying "the last two crumbs". The declaration-site comment keeps a short pointer that the selectors hard-code the value, and the usage-site comment carries the full explanation of the Tailwind constraint and points at the test.

Rendering and behavior are unchanged.

## Visual evidence

None — carries the `no-visual-change` label. The diff on `breadcrumbs.tsx` is limited to comments and an added `export` keyword; no JSX, class string, or style changes, so rendered output is byte-identical to `main`. Earlier revisions of this PR committed two `docs/pr-evidence/` screenshots that documented no change (and duplicated the existing `issue-2141-breadcrumbs-*.png` evidence for this same component); they have been removed.

## Testing

- `npx vitest run src/components/shared/breadcrumbs.test.tsx`: 8/8 pass, including the new drift test.
- Full frontend suite: 1613 tests across 110 files pass.
- Drift verified by mutation, not inspection: bumping the constant alone, bumping only the `-n+2` range selector, and bumping only the `nth-last-child(2)` position selector each fail the new test.
- `prek -a`: all hooks pass (eslint, prettier, tsc, stylelint, frontend CSS guards, etc.).

CI runs the full backend, system and frontend suites.

Closes #2196


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2217, opened 2026-09-11
test: name repeated literals and derive the selected key in LogTableWithDrawer tests

### LINKED ISSUE #2216: Clean up duplicated grid-class and testid literals in log-table-with-drawer.test.tsx
Surfaced by an automated quality scan during the autofix review of #2215. Pre-existing; not introduced by that branch (which only moved two import statements into the import block).

## Description

`frontend/src/components/shared/log-table/log-table-with-drawer.test.tsx` repeats three literals across paired assertions, and uses a redundant non-null assertion inconsistently with the rest of the file.

The highest-value item is the Tailwind class string. Lines 58 and 63 are a matched positive/negative pair:

```ts
expect(getByTestId("log-table-with-drawer").className).toContain("grid-cols-[1fr_var(--size-drawer)]");
// ...
expect(getByTestId("log-table-with-drawer").className).not.toContain("grid-cols-[1fr_var(--size-drawer)]");
```

If the open-state grid class in `log-table-with-drawer.tsx` ever changes and only the positive assertion is updated, the negative one silently becomes vacuous — it would pass against any class string. A shared constant makes the two assertions move together.

## Key Items

- `log-table-with-drawer.test.tsx:58,63` — `"grid-cols-[1fr_var(--size-drawer)]"` duplicated across a `toContain` / `not.toContain` pair. Drift here turns the negative assertion vacuous rather than failing.
- `log-table-with-drawer.test.tsx:36,58,63` — the `"log-table-with-drawer"` testid literal repeated three times.
- `log-table-with-drawer.test.tsx:43,49` — the `"log-table-drawer-table-area"` testid literal repeated twice.
- `log-table-with-drawer.test.tsx:15,57,69` — the `"1001-1"` selectedKey literal is derived from `makeEntry(1)`'s `timestamp: 1000 + seq` / `seq` pair, but that `{timestamp}-{seq}` relationship is implicit. A reader cannot tell from the call site that `"1001-1"` is meant to name the entry `makeEntry(1)` produces.
- `log-table-with-drawer.test.tsx:44,50` — `tableArea!.contains(...)` / `tableArea!.textContent` use a non-null assertion on a `getByTestId` result. `getByTestId` throws when the element is missing, so the `!` is dead noise, and it is inconsistent with lines 36, 58, 63, 70, and 75, which call straight through on the same return type.

## Acceptance Criteria

- [ ] The open-state grid class string is a single file-level constant referenced by both the `toContain` and `not.toContain` assertions.
- [ ] The `"log-table-with-drawer"` and `"log-table-drawer-table-area"` testids are file-level constants rather than inline literals at each call site.
- [ ] The `"1001-1"` selectedKey is derived from `makeEntry` (or a small helper) so its relationship to the entry's `timestamp`/`seq` is explicit rather than implied.
- [ ] The redundant `!` non-null assertions on `getByTestId` results are removed.
- [ ] `npx vitest run src/components/shared/log-table/log-table-with-drawer.test.tsx` passes in `frontend/`.

NodeJSmith:
## Summary

Cleans up `frontend/src/components/shared/log-table/log-table-with-drawer.test.tsx`. This is test-only and changes no runtime behavior.

- **Shared grid-class constant.** The open-drawer Tailwind class `grid-cols-[1fr_var(--size-drawer)]` is now `DRAWER_OPEN_GRID_CLASS`. Both the `toContain` and `not.toContain` assertions use it. Before, the two assertions were separate literals. If only the positive one got updated after a class change, the negative one would keep passing against any class string.
- **Testid constants.** `"log-table-with-drawer"` and `"log-table-drawer-table-area"` are now the file-level constants `WRAPPER_TEST_ID` and `TABLE_AREA_TEST_ID`.
- **Derived selected key.** The hardcoded `"1001-1"` selectedKey is replaced by a `makeSelectedDrawerProps()` helper that builds the entry with `makeEntry(1)` and gets its key from the real `rowKey()` in `./types`. The key can no longer drift from the entry's `timestamp`/`seq` or from how production builds keys.
- **Removed redundant `!`.** Dropped the non-null assertions on `getByTestId(...)` results. `getByTestId` throws when the element is missing, so the `!` did nothing, and the rest of the file already called straight through.

## Testing

- `prek -a`: all hooks pass (ruff, eslint, prettier, tsc, test-factory shadowing, etc.)
- `npx vitest run src/components/shared/log-table/log-table-with-drawer.test.tsx` in `frontend/`: 7/7 tests pass

CI runs the full backend and frontend suites.

Closes #2216

### DIFF
--- a/frontend/src/components/shared/log-table/log-table-with-drawer.test.tsx
+++ b/frontend/src/components/shared/log-table/log-table-with-drawer.test.tsx
@@ -4,12 +4,17 @@ import { describe, expect, it, vi } from "vitest";
 import { createLogEntry } from "@/test/factories";
 
 import { LogTableWithDrawer } from "./log-table-with-drawer";
+import { rowKey } from "./types";
 import type { LogDrawerProps } from "./use-log-table";
 
 vi.mock("./log-detail-drawer", () => ({
   LogDetailDrawer: (props: { selectedKey: string | null }) =>
     props.selectedKey ? <aside data-testid="drawer" role="complementary" /> : null,
 }));
+
+const WRAPPER_TEST_ID = "log-table-with-drawer";
+const TABLE_AREA_TEST_ID = "log-table-drawer-table-area";
+const DRAWER_OPEN_GRID_CLASS = "grid-cols-[1fr_var(--size-drawer)]";
 
 function makeEntry(seq: number) {
   return createLogEntry({ seq, timestamp: 1000 + seq, message: `msg-${seq}`, app_key: "app", source_tier: "app" });
@@ -25,6 +30,11 @@ function makeDrawerProps(overrides: Partial<LogDrawerProps> = {}): LogDrawerProp
   };
 }
 
+function makeSelectedDrawerProps(): LogDrawerProps {
+  const entry = makeEntry(1);
+  return makeDrawerProps({ selectedKey: rowKey(entry), entries: [entry] });
+}
+
 function renderWithDrawer(drawerProps: LogDrawerProps, children = <div data-testid="table-content" />) {
   return render(<LogTableWithDrawer drawerProps={drawerProps}>{children}</LogTableWithDrawer>);
 }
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
## Automated review pass

CodeRabbit had no unresolved threads on this PR (its review came back rate-limited), so the full reviewer battery was run against `origin/main...HEAD` instead.

| Reviewer | Verdict |
|---|---|
| integration-reviewer | PASS — 0 findings |
| wtf-reviewer | 2 LOW |
| llm-checker | 2 LOW |
| lazy-checker | 3 LOW |
| nitpicker | 7 style nits |

No CRITICAL/HIGH/MEDIUM findings, so nothing was changed. Verified independently:

- `rowKey(makeEntry(1))` derives exactly `"1001-1"` (`seq=1`, `timestamp=1000+1`), matching the literal it replaced — the helper is equivalent, not a silent coverage change.
- `DRAWER_OPEN_GRID_CLASS` matches `log-table-with-drawer.tsx:19` verbatim; the trailing `max-tablet:grid-cols-1` doesn't break the `toContain` match.
- The removed `!` assertions were redundant — RTL's `getByTestId` throws rather than returning `null`.
- `npx vitest run src/components/shared/log-table/log-table-with-drawer.test.tsx` → 7/7 pass.

The remaining LOW findings cancel out and were deliberately not acted on: llm-checker argues `TABLE_AREA_TEST_ID` and `DRAWER_OPEN_GRID_CLASS` are over-extracted for two uses apiece, while lazy-checker and nitpicker argue the opposite — that `"table-content"` and `"drawer"` should have been extracted too. Both directions are defensible; churning the file again to satisfy one at the other's expense isn't worth a round trip. (`"drawer"` in particular can't be extracted cleanly — it's referenced inside the hoisted `vi.mock` factory, which can't close over a module-level const.)

`file-sizes` and `duplicate-code` are the known pre-existing `main` baseline failures (`continue-on-error`), not introduced here. Auto-merge enabled.

PR #2222, opened 2026-09-12
test: assert filter-icon active dot with queryByTestId null check

### LINKED ISSUE #2218: Clean up redundant active-dot assertion in filter-icon.test.tsx
## Description

A quality scan of `frontend/src/components/shared/filter-icon.test.tsx` (llm-checker, lazy-checker, and nitpicker) turned up one small hygiene problem in the positive "active dot" test. It was already in the file before the scan.

## Key Items

- **`filter-icon.test.tsx:40-42`: the active-dot assertion is redundant and uses a different idiom from the rest of the file.**
  The test does:
  ```tsx
  const dot = screen.getByTestId("filter-icon-dot");
  expect(dot).toBeTruthy();
  ```
  `getByTestId` throws when the element is missing, so `toBeTruthy()` can never fail. The throwing getter is what actually tests the dot. The line also checks for existence with `toBeTruthy()`, but everywhere else the file uses `null` checks: `.not.toBeNull()` on lines 10 and 16, and `queryByTestId(...)` with `.toBeNull()` on lines 47 and 52.

  **Recommended fix:** replace lines 41-42 with one assertion that mirrors the two negative tests right below it:
  ```tsx
  expect(screen.queryByTestId("filter-icon-dot")).not.toBeNull();
  ```
  **Why this one:** the positive and negative dot tests then use the same query, so they differ only in `.not`. The assertion becomes the real check instead of riding on a throwing getter. It also matches the `.not.toBeNull()` idiom the file already uses on lines 10 and 16, which is the more common idiom across `frontend/src` tests. The unused `dot` local goes away. This doesn't change what the test covers.

## Acceptance Criteria

- [ ] `filter-icon.test.tsx`'s "shows active dot when active=true" test uses `expect(screen.queryByTestId("filter-icon-dot")).not.toBeNull()` instead of `getByTestId` + `toBeTruthy()`
- [ ] `cd frontend && npx vitest run src/components/shared/filter-icon.test.tsx` passes

NodeJSmith:
## Summary

Replaces the redundant active-dot assertion in `filter-icon.test.tsx` with one that
actually does the checking.

The "shows active dot when active=true" test was written as:

```tsx
const dot = screen.getByTestId("filter-icon-dot");
expect(dot).toBeTruthy();
```

`getByTestId` throws when the element is missing, so the throwing getter was the real
test and `toBeTruthy()` could never fail — it asserted on a value that was guaranteed
non-null by the line above it. Now:

```tsx
expect(screen.queryByTestId("filter-icon-dot")).not.toBeNull();
```

## Why

- The assertion itself is now the check, rather than riding on a side effect of the query helper.
- The positive and the two negative dot tests now use the same `queryByTestId` query and differ
  only by `.not`, so they read as a set.
- `.not.toBeNull()` is the idiom already used elsewhere in this file (lines 10 and 16) and is the
  more common one across `frontend/src` tests.
- The unused `dot` local goes away.

Test coverage is unchanged — this is a hygiene fix, not a behavior change.

## Testing

- `cd frontend && npx vitest run src/components/shared/filter-icon.test.tsx` — 8 passed
- `prek -a` — all hooks pass (ruff, eslint, prettier, stylelint, tsc, and the repo's custom checks)

CI runs the full backend, system, and frontend suites.

No screenshots: the change is confined to a `*.test.tsx` file, which
`tools/frontend/check_pr_screenshots.py` explicitly excludes from the visual-evidence
requirement, and nothing rendered was touched.

Closes #2218

### DIFF
--- a/frontend/src/components/shared/filter-icon.test.tsx
+++ b/frontend/src/components/shared/filter-icon.test.tsx
@@ -38,8 +38,7 @@ describe("FilterIcon", () => {
 
   it("shows active dot when active=true", () => {
     render(<FilterIcon active={true} />);
-    const dot = screen.getByTestId("filter-icon-dot");
-    expect(dot).toBeTruthy();
+    expect(screen.queryByTestId("filter-icon-dot")).not.toBeNull();
   });
 
   it("does not show active dot when active=false", () => {


NodeJSmith:
@coderabbitai review

NodeJSmith:
## Autofix review — clean, no changes needed

**CodeRabbit:** review completed with zero findings (0 threads, 0 review comments).

**Review battery** (6 reviewers on `git diff origin/main...HEAD`):

| Reviewer | Verdict |
|---|---|
| llm-checker | CLEAN |
| lazy-checker | CLEAN |
| wtf-reviewer | 1 LOW (see below) |
| nitpicker | 2 LOW, both pre-existing |
| integration-reviewer | 1 MEDIUM, pre-existing (see below) |

**Verification:** `filter-icon.test.tsx` 8/8 pass; full `src/components/shared/` suite 500/500 pass; `eslint` and `tsc --noEmit` clean. All four required CI checks green (`tests-complete`, `e2e-complete`, `lint-complete`, `docs/readthedocs.org`). `file-sizes` and `duplicate-code` fail as known `continue-on-error` baseline noise.

### Findings not actioned in this PR

- **MEDIUM (integration-reviewer) — parallel drift, pre-existing.** `sort-header.test.tsx:425` and `table-footer.test.tsx:118` assert presence of the *same* `filter-icon-dot` test id using the exact `getByTestId(...)` + `toBeTruthy()` shape this PR removes. Both predate this branch and sit outside the issue's `scanned-path` scope, so they're tracked in #2223 rather than expanding this PR.
- **LOW (wtf-reviewer, llm-checker) — idiom tradeoff, accepted.** `getByTestId` throws with a DOM snapshot on failure, so it yields a better error message than `queryByTestId` + `not.toBeNull()`. Keeping the change as-is: it's what #2218's acceptance criteria specify, the old `toBeTruthy()` was a no-op riding on the throwing getter, and all three active-dot tests in this file now differ only by polarity. Deliberately **not** proposing a repo-wide rewrite of the ~19 other `getByTestId(...).toBeTruthy()` sites — that would trade away the better failure messages for no real gain.

PR #2224, opened 2026-09-12
refactor: replace FilterIcon inline styles with Tailwind utilities

### LINKED ISSUE #2219: Replace inline style objects with Tailwind utilities in FilterIcon
Surfaced by an automated quality scan of `frontend/src/components/shared/filter-icon.tsx`. One finding survived validity assessment.

## Finding: component bypasses the repo's Tailwind styling convention with raw inline `style` objects

`FilterIcon` styles both of its elements with inline `style` objects carrying entirely static values:

- `filter-icon.tsx:10` — wrapper `<span style={{ position: "relative", display: "inline-flex", alignItems: "center" }}>`
- `filter-icon.tsx:17-25` — active-dot `<span>` with a 7-property inline style object (`position`, `top`, `right`, `width`, `height`, `borderRadius`, `background`)

CLAUDE.md (CSS Architecture → Component Styling) states the convention directly: "Use Tailwind utilities directly in JSX. Compose conditional class names with `cn()`." Every value here is static and has a direct Tailwind equivalent, so there is no computed-style justification. This is the largest static inline-style block in `components/shared/` — the two sibling files that also use inline styles have a reason to: `table-card.tsx:29` sets a CSS custom property from a prop (Tailwind can't express that), and `status-shape.tsx:66` is a single `flexShrink: 0`.

### Recommended fix

Replace both `style` objects with `className` strings:

```tsx
<span className="relative inline-flex items-center">
  <svg width={size} height={size} viewBox="0 0 12 12" fill="none" aria-hidden="true">
    <path d="M1 2h10L7.5 6.5V10L4.5 9V6.5L1 2z" stroke="currentColor" strokeWidth="1.2" strokeLinejoin="round" />
  </svg>
  {active && (
    <span
      data-testid="filter-icon-dot"
      className="absolute top-0 right-0 size-[5px] rounded-full bg-[var(--accent)]"
    />
  )}
</span>
```

Rationale for this direction over leaving the inline styles or introducing a CSS rule: it matches the documented repo convention exactly, keeps the diff to two attributes in one file, produces identical rendered output, and uses `size-*` the same way `icons.tsx` already does (`size-4` on `StrokeIcon`). This also incidentally resolves the magic `5px` / `50%` literals the scan flagged separately — they become idiomatic Tailwind values rather than untokenized inline numbers, so no new constant needs to be introduced.

### Important: do not write `bg-accent`

The dot's color must stay the arbitrary-value form `bg-[var(--accent)]`. `global.css:31` maps Tailwind's `--color-accent` to `var(--highlight-bg)` (shadcn's subtle hover-background role), which is a *different color* from Hassette's brand `--accent` at `global.css:70`. Writing `bg-accent` would silently change the active-filter dot from the brand accent to a pale surface color. CLAUDE.md calls out this exact aliasing collision.

### Acceptance criteria

- [ ] `frontend/src/components/shared/filter-icon.tsx` contains no `style={{ ... }}` attributes
- [ ] Wrapper span uses `className="relative inline-flex items-center"`
- [ ] Active dot uses `bg-[var(--accent)]` (not `bg-accent`) and retains its `data-testid="filter-icon-dot"`
- [ ] `frontend/src/components/shared/filter-icon.test.tsx` still passes unchanged
- [ ] Rendered appearance is unchanged (no visual diff)

## Not filed

The scan also flagged `viewBox="0 0 12 12"` and `strokeWidth="1.2"` as magic numbers needing named constants. Assessed as invalid: the repo's canonical icon module `icons.tsx` inlines `viewBox="0 0 24 24"` and `strokeWidth="2"` in its own `StrokeIcon` wrapper, and `status-shape.tsx` does the same. SVG geometry is coupled to its path data; extracting it would diverge from the established pattern and reduce readability.

NodeJSmith:
## Summary

`FilterIcon` styled both of its elements with inline `style` objects containing entirely static values, bypassing the repo's Tailwind convention documented in CLAUDE.md (Component Styling). Every value had a direct utility equivalent — there was no computed-style justification for the inline objects.

- Outer wrapper: `style={{ position, display, alignItems }}` → `className="relative inline-flex items-center"`
- Active dot: `style={{ position, top, right, width, height, borderRadius, background }}` → `className="absolute top-0 right-0 size-[5px] rounded-full bg-[var(--accent)]"`

Net: 10 lines removed, 2 added. No behavior or API change — the component's props, the `data-testid="filter-icon-dot"` hook, and the SVG markup are untouched.

## Why `bg-[var(--accent)]` and not `bg-accent`

The active dot deliberately keeps the arbitrary-value form. Tailwind's `--color-accent` is aliased to `var(--highlight-bg)` — shadcn's subtle hover-background role — which is a *different color* from Hassette's brand `--accent`. This is the exact aliasing collision called out in CLAUDE.md's CSS Architecture section, so `bg-accent` would silently change the dot's color.

## Testing

Frontend suite scoped to the changed component and both of its consumers — CI runs the full suite:

```
npx vitest run src/components/shared/filter-icon.test.tsx \
                src/components/shared/sort-header.test.tsx \
                src/components/shared/table-footer.test.tsx
→ 3 files passed, 43 tests passed
```

`prek -a` passes clean (ruff, prettier, eslint, stylelint, tsc, plus all repo-specific hooks).

Verified against the live demo stack: the dot's computed styles match the previous inline values exactly — position, offsets, 5×5 size, and background `oklch(0.5 0.14 255)`. The one computed-value difference is `rounded-full` resolving to `calc(infinity * 1px)` instead of `50%`, which clamps to the same circle on a 5×5 box.

## Screenshots

Active filter icon, zoomed:

Active filter icon zoom

Inactive filter icon, zoomed:

Inactive filter icon zoom

Mobile table-footer consumer, active:

Mobile table footer active

Handlers view with filter active (sort-header consumer, in context):

Handlers filter active

Closes #2219


NodeJSmith:
@coderabbitai review

NodeJSmith:
Post-merge review pass (this PR auto-merged before the scheduled review reached it).

**Code:** clean. Six reviewers (code, integration, WTF, llm-checker, lazy-checker, nitpicker) found nothing blocking. Every replaced inline property maps 1:1 to its utility, and the `--accent` / `--color-accent` aliasing trap is handled correctly — `bg-accent` would have resolved to `--highlight-bg` and silently changed the dot's color. Checked the `bg-primary` alternative two reviewers raised: `--primary` is byte-identical to `--accent` in both the light (`global.css:70` / `:215`) and dark (`:378` / `:412`) blocks, so it would have worked, but the arbitrary-value form matches the nine existing `*-[var(--accent*)]` sibling usages and is what #2219 prescribed. No change warranted.

The only residual findings were the un-tokenized `size-[5px]` dot (carried over verbatim from the pre-refactor inline style, not introduced here) and `status-shape.tsx:66`'s `flexShrink: 0` — out of scope for an issue scoped to `FilterIcon`, and #2219 already accounted for it.

**One real issue found, unrelated to the code:** the Screenshots section above was rendering as four broken images. The links were branch-scoped (`.../hassette/autofix/issue-2219/...`) and the branch was deleted on merge, so all four 404'd. I rewrote them to the merge-commit SHA and confirmed all four return 200. The same pattern is used across the merged autofix PR history, so it's systemic — filed as #2225, which also proposes tightening `check_pr_screenshots.py` so the #1063 guard actually guarantees durable evidence rather than evidence that evaporates at merge.

PR #2226, opened 2026-09-12
test: assert real Tailwind classes and deduplicate icons smoke tests

### LINKED ISSUE #2221: Clean up stale CSS-modules assertions and duplication in icons smoke tests
Quality scan of `frontend/src/components/shared/icons.test.tsx` (45 lines, 6 tests). Three findings, all test-only. Pre-existing — not introduced by any in-flight branch.

## 1. Stale CSS-modules references — the test name and comment describe a system this repo no longer uses

`icons.test.tsx:27` names the test `"all icons have the iconSvg module class"`, and `icons.test.tsx:32` comments `// CSS modules produce a scoped class name — just verify the svg has a class attribute`.

Neither is true anymore:
- `grep -rn iconSvg src` matches only this test's own name — no such class exists anywhere in the codebase.
- `find src -name '*.module.css'` returns nothing. CLAUDE.md (CSS Architecture) states outright: "There are no CSS Modules, no `ht-*` global utility classes, and no separate `tokens.css` file."
- `icons.tsx:11` applies plain Tailwind utilities: `className="size-4 shrink-0 align-middle"` — unhashed, directly assertable.

The consequence is that `expect(svg?.getAttribute("class")).toBeTruthy()` is now a near-tautology whose stated justification (hashed class names can't be checked exactly) no longer holds. Any non-empty class passes, so the test cannot catch the shared wrapper losing its sizing/alignment utilities.

**Recommended fix:** rename the test to `"all stroke icons carry the shared wrapper classes"`, delete the stale CSS-modules comment, and assert the concrete classes the `StrokeIcon` wrapper actually applies (`size-4`, `shrink-0`, `align-middle`) rather than mere truthiness. Rationale: the only reason the assertion was weak was class-name hashing, and that reason is gone — with Tailwind the class list is stable source text, so the strong assertion is now both possible and cheap, and it turns a vacuous check into one that would actually fail if `StrokeIcon` lost its sizing.

## 2. Four hand-unrolled tests sit next to two loop-based tests solving the same problem

`icons.test.tsx:7-25` are four near-identical `it` blocks (`IconPlay`, `IconSquare`, `IconRefresh`, `IconWarning`), each with the same three-line body and only the component name varying. The very next two tests (`:27-35`, `:37-44`) already loop over an inline icon array — and that array literal `[IconPlay, IconSquare, IconRefresh, IconWarning]` is itself duplicated verbatim at `:28` and `:38`. Adding or removing an icon today means editing six places in a 45-line file.

**Recommended fix:** hoist one module-level `const STROKE_ICONS = [...]` and convert all six tests to `it.each`, naming cases via `STROKE_ICONS.map((Icon) => [Icon.name, Icon] as const)` so failures still report which icon broke. Rationale: `it.each` is already the established convention in this suite — 10 usages across 8 frontend test files, including the hoisted-constant form (`it.each(SCHEDULE_STATUSES)` in `handlers-tab.job.test.tsx:357`, `it.each(conditionalCellCases)` in `stat-cell-builders.test.ts:77`) — so this matches sibling code rather than inventing a pattern, and it collapses the duplicated array to one edit site.

## 3. Two of the module's six exported icons are never tested

`icons.tsx` exports `IconPlay`, `IconSquare`, `IconRefresh`, `IconArrowRight`, `IconWarning`, and `IconChevron`. The test file covers four. `IconArrowRight` (`icons.tsx:47`) is a zero-prop `StrokeIcon` structurally identical to the four that are covered — it is simply missing from the list, which is exactly the drift the duplicated inline arrays in finding 2 invite. `IconChevron` (`icons.tsx:63`) is untested entirely; it takes a required `open` prop, renders its own bare `<svg>` rather than the `StrokeIcon` wrapper, and swaps its `points` on `open`.

**Recommended fix:** add `IconArrowRight` to the `STROKE_ICONS` constant from finding 2 (it needs no other change — it satisfies every existing assertion), and add one separate `it` for `IconChevron` asserting `aria-hidden="true"` plus different `points` for `open: true` vs `open: false`. Rationale: `IconChevron` deliberately does not use the shared wrapper, so folding it into the shared loop would force the loop's assertions to weaken back toward the vacuous state finding 1 removes; a dedicated test keeps the shared-wrapper invariant strict while still covering the one icon with real conditional rendering logic.

## Acceptance Criteria

- [ ] No reference to `iconSvg` or CSS modules remains in `icons.test.tsx`.
- [ ] The shared-wrapper test asserts the concrete `StrokeIcon` utility classes, not just a truthy `class` attribute.
- [ ] A single module-level icon constant exists; no icon array literal is duplicated.
- [ ] The four unrolled per-icon tests are replaced by `it.each`, with per-icon names still visible in failure output.
- [ ] `IconArrowRight` is covered by the shared-icon assertions.
- [ ] `IconChevron` has its own test covering `aria-hidden` and both `open` states.
- [ ] `cd frontend && npm test -- icons.test` passes.

## Considered and rejected

The nitpicker also flagged the literal `"svg"` repeated 7 times as an uncentralized magic string. Rejecting it: `"svg"` is an HTML tag name inside `container.querySelector`, not a project value that can drift, and no sibling test in `frontend/src` extracts DOM selectors to constants. Naming it would add indirection for zero protection.

NodeJSmith:
## Summary

Rewrites `frontend/src/components/shared/icons.test.tsx` so its assertions match how the icons are actually built today, and collapses the hand-unrolled duplication.

The file was written against a CSS-modules setup this repo no longer uses. The old test was named `"all icons have the iconSvg module class"` and asserted only `expect(svg?.getAttribute("class")).toBeTruthy()`, justified by a comment claiming class names are hashed and therefore unassertable. Neither holds: there are no `*.module.css` files, `iconSvg` existed nowhere but in that test's own name, and `StrokeIcon` applies plain Tailwind utilities (`size-4 shrink-0 align-middle`) that are stable source text. The assertion had decayed into a near-tautology that would still pass if the shared wrapper lost all of its sizing and alignment classes.

## What changed

- **Assert the real classes.** The shared-wrapper test now checks for the concrete `size-4`, `shrink-0`, and `align-middle` utilities `StrokeIcon` applies, instead of mere truthiness. The stale CSS-modules test name and comment are gone.
- **One icon list, `it.each` everywhere.** Four near-identical per-icon `it` blocks and two separate inline `[IconPlay, IconSquare, IconRefresh, IconWarning]` array literals collapse into a single module-level `STROKE_ICONS` constant driving `it.each`. Cases are named via `Icon.name`, so a failure still reports which icon broke. Adding an icon is now one edit site instead of six.
- **Cover the two missing icons.** `IconArrowRight` is a zero-prop `StrokeIcon` that was simply absent from the list — it joins `STROKE_ICONS` and needs no other change. `IconChevron` gets its own test, since it deliberately renders a bare `<svg>` rather than the shared wrapper; folding it into the shared loop would force those assertions back toward the vacuous state this PR removes. Its test covers `aria-hidden` plus the `points` swap between `open: true` and `open: false`.

Test-only change — no source, behavior, or rendered output is touched.

## Testing

- `npx vitest run src/components/shared/icons.test.tsx` — 16 passed.
- `prek -a` — all hooks pass, including `tsc`, `eslint`, `prettier`, and `stylelint` for the frontend.

CI runs the full backend, system, and frontend suites across every supported Python version.

No screenshots: this touches only `*.test.tsx`, which `tools/frontend/check_pr_screenshots.py` excludes from the visual-evidence requirement, and nothing rendered changes.

Closes #2221

### DIFF
--- a/frontend/src/components/shared/icons.test.tsx
+++ b/frontend/src/components/shared/icons.test.tsx
@@ -1,45 +1,45 @@
 import { render } from "@testing-library/react";
 import { describe, expect, it } from "vitest";
 
-import { IconPlay, IconRefresh, IconSquare, IconWarning } from "./icons";
+import { IconArrowRight, IconChevron, IconPlay, IconRefresh, IconSquare, IconWarning } from "./icons";
+
+/** Every icon built on the shared `StrokeIcon` wrapper. `IconChevron` renders its own svg, so it is tested separately. */
+const STROKE_ICONS = [IconPlay, IconSquare, IconRefresh, IconArrowRight, IconWarning];
+
+const STROKE_ICON_CASES = STROKE_ICONS.map((Icon) => [Icon.name, Icon] as const);
 
 describe("Icons smoke tests", () => {
-  it("IconPlay renders an SVG element", () => {
-    const { container } = render(<IconPlay />);
+  it.each(STROKE_ICON_CASES)("%s renders an SVG element", (_name, Icon) => {
+    const { container } = render(<Icon />);
     expect(container.querySelector("svg")).not.toBeNull();
   });
 
-  it("IconSquare renders an SVG element", () => {
-    const { container } = render(<IconSquare />);
-    expect(container.querySelector("svg")).not.toBeNull();
+  it.each(STROKE_ICON_CASES)("%s carries the shared wrapper classes", (_name, Icon) => {
+    const { container } = render(<Icon />);
+    const svg = container.querySelector("svg");
+    expect(svg?.getAttribute("class")?.split(/\s+/)).toEqual(
+      expect.arrayContaining(["size-4", "shrink-0", "align-middle"]),
+    );
   });
 
-  it("IconRefresh renders an SVG element", () => {
-    const { container } = render(<IconRefresh />);
-    expect(container.querySelector("svg")).not.toBeNull();
+  it.each(STROKE_ICON_CASES)("%s has aria-hidden='true'", (_name, Icon) => {
+    const { container } = render(<Icon />);
+    expect(container.querySelector("svg")?.getAttribute("aria-hidden")).toBe("true");
   });
 
-  it("IconWarning renders an SVG element", () => {
-    const { container } = render(<IconWarning />);
-    expect(container.querySelector("svg")).not.toBeNull();
-  });
+  it("IconChevron is aria-hidden and swaps its points on open", () => {
+    const closed = render(<IconChevron open={false} />);
+    const closedSvg = closed.container.querySelector("svg");
+    expect(closedSvg?.getAttribute("aria-hidden")).toBe("true");
+    const closedPoints = closedSvg?.querySelector("polyline")?.getAttribute("points");
 
-  it("all icons have the iconSvg module class", () => {
-    const icons = [IconPlay, IconSquare, IconRefresh, IconWarning];
-    for (const Icon of icons) {
-      const { container } = render(<Icon />);
-      const svg = container.querySelector("svg");
-      // CSS modules produce a scoped class name — just verify the svg has a class attribute
-      expect(svg?.getAttribute("class")).toBeTruthy();
-    }
-  });
+    const opened = render(<IconChevron open={true} />);
+    const openedSvg = opened.container.querySelector("svg");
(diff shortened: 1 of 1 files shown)


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2231, opened 2026-09-13
chore: drop unused IconChevron size prop and correct the shared icons header

### LINKED ISSUE #2228: Clean up unused size prop and incorrect header comment in shared icons
## Description

A quality scan of `frontend/src/components/shared/icons.tsx` surfaced two pre-existing findings: an unused speculative prop and a factually incorrect file header comment. Both predate the current work and are independently actionable.

Note: the file is otherwise clean. Two of the three checkers returned no findings, and the third's remaining findings were assessed as invalid (see Assessed and rejected below).

### 1. `IconChevron`'s `size` prop is never used by any caller (`icons.tsx:63-66`)

`IconChevron` declares `size?: number` with a default of `10`, but no call site anywhere in the frontend passes it:

- `components/app-detail/registration-footer.tsx:74` — `<IconChevron open={registrationExpanded} />`
- `pages/apps-table-row.tsx:68` — `<IconChevron open={isExpanded} />`
- `components/shared/config-schema-view.tsx:254` — `<IconChevron open={expanded} />`
- `components/shared/log-table/log-table-view.tsx:229` — `<IconChevron open={isSelected} />`
- `components/shared/icons.test.tsx:31,36` — `<IconChevron open={false} />` / `<IconChevron open={true} />`

`grep -rn 'IconChevron[^)]*size'` matches only the definition itself. There is no barrel/index re-export in `components/shared/`, and every importer is a direct in-repo import, so the consumer list above is exhaustive — this is not a public API surface with unknown users.

**Recommended fix (committed direction):** delete the `size` prop and inline the constant, so the signature becomes `({ open }: { open: boolean })` and the svg carries `width="10" height="10"`.

Rationale for this direction over keeping the prop "for flexibility": every existing consumer renders at the default, so the parameter is speculative surface with no observed demand. This matches the sibling convention in the same file — the five `StrokeIcon`-based icons take no props at all and fix their size via the `size-4` utility class. It is also the smallest diff, since no call site changes. Removing it is not an observable behavior change: rendering is byte-identical for all current callers.

### 2. File header comment is factually wrong and describes icons that are not in this file (`icons.tsx:1-6`)

```
// Shared SVG icon components
// Sidebar icons use fill-based paths (no stroke attributes).
// All other icons use stroke-based attributes.
...
// --- Stroke-based icons ---
```

Two problems:

- **No sidebar icons live in this file.** The sidebar's chevron is `SidebarChevron`, defined locally in `components/layout/sidebar.tsx:28`.
- **The claim about them is also backwards.** `SidebarChevron` renders `fill="none" stroke="currentColor" strokeWidth="1.5"` — stroke-based, not "fill-based paths (no stroke attributes)".

The `// --- Stroke-based icons ---` divider at line 6 compounds it by labelling the file's only section, implying a fill-based section that does not exist.

This comment actively misleads: two independent reviewers in this scan read it and drew different wrong conclusions from it — one treated it as documenting a real fill-vs-stroke split in this file, the other as documenting an `IconChevron` convention break it never mentions.

**Recommended fix (committed direction):** delete lines 2-3 and the line 6 divider, keeping only `// Shared SVG icon components`. Leave the accurate per-component docstrings at `:8` and `:62` untouched.

Rationale for deleting rather than rewriting the comment to describe the file correctly: the fill-vs-stroke distinction it is reaching for is already conveyed by the `StrokeIcon` wrapper's own docstring at `:8`, so a corrected version would restate what the code shows. The divider removal additionally matches the house "No Section Divider Comments" convention.

## Assessed and rejected

The nitpicker pass flagged the SVG geometry literals (`viewBox="0 0 24 24"` at `:13`, `strokeWidth="2"` at `:16`, `viewBox="0 0 12 12"` at `:64`, `strokeWidth="1.5"` at `:65`, and the polyline `points` strings at `:65`) as magic numbers needing named constants. Not filed:

- Each literal occurs exactly once in the file, so naming it adds indirection without removing duplication or a branch.
- The `points` strings at `:65` were singled out while the ~10 sibling coordinate literals (`polygon points` at `:28`, `rect` dims at `:34`, path `d` at `:40-43`, `:49-50`, `:56-58`) were not. Extracting only the chevron's would create the inconsistency the finding claims to fix.
- The stated rationale was speculative ("if a second icon size/weight is ever introduced"), and the `strokeWidth` 2-vs-1.5 split is deliberate, not drift: the values sit on different coordinate grids (2 on a 24-unit grid, 1.5 on a 12-unit grid), and `SidebarChevron` independently uses the same 1.5-on-12 pairing.

The `size = 10` default was also flagged as a magic number; it is subsumed by item 1, which removes it outright.

## Acceptance Criteria

- [ ] `IconChevron`'s signature is `({ open }: { open: boolean })` — the `size` prop and its default are gone, and the svg uses literal `width="10" height="10"`.
- [ ] No call site required changes (all five existing consumers still compile untouched).
- [ ] `icons.tsx` header is reduced to the single accurate line `// Shared SVG icon components`; the sidebar/fill-based claim and the `// --- Stroke-based icons ---` divider are removed.
- [ ] Docstrings at `:8` (`StrokeIcon`) and `:62` (`IconChevron`) are unchanged.
- [ ] `frontend/src/components/shared/icons.test.tsx` still passes, and `npm run build` / lint are clean.

NodeJSmith:
## Summary

Two unrelated-but-adjacent cleanups in `frontend/src/components/shared/icons.tsx`, both no-ops at render time.

**Dead prop on `IconChevron`.** The component declared `size?: number` with a default of `10`, but none of its four call sites (`pages/apps-table-row.tsx`, `components/app-detail/registration-footer.tsx`, `components/shared/log-table/log-table-view.tsx`, `components/shared/config-schema-view.tsx`) ever passed it. The signature is now `({ open })` and the svg carries literal `width="10" height="10"`, so every existing consumer renders byte-identical output.

**Inaccurate file header.** The header claimed that sidebar icons live in this file and use fill-based paths. Neither half is true — the sidebar's chevron is `SidebarChevron`, defined locally in `components/layout/sidebar.tsx`, and it is stroke-based (`fill="none" stroke="currentColor"`). That claim is dropped, along with the `// --- Stroke-based icons ---` divider that labelled the file's only section. The `StrokeIcon` docstring already carries the stroke convention, so nothing is lost.

## Testing

CI runs the full suite; locally I ran the scoped checks that cover this change:

- `prek -a` — all hooks passed, including `tsc (frontend)`, `eslint (frontend)`, `prettier (frontend)`, and `stylelint (frontend)`.
- `npx vitest run src/components/shared/icons.test.tsx` — 16/16 passed. This file covers `IconChevron` directly, asserting it is `aria-hidden` and that it swaps its `points` on `open`.
- `npx vitest run src/pages src/components/app-detail src/components/shared/log-table src/components/shared/config-schema-view.test.tsx` — 42 files, 757 tests passed. This is every consumer of `IconChevron`, run because the component's signature changed.

## Screenshots

Chevron in both states, confirming the inlined dimensions render identically:

Chevron open and closed

The chevron in context on the apps table:

Apps table chevron

Closes #2228


NodeJSmith:
@coderabbitai review

PR #2232, opened 2026-09-13
refactor: deduplicate column-picker button styling and share its props type

### LINKED ISSUE #2230: Clean up styling duplication and prop-type duplication in log-table column picker
Surfaced by the automated quality scanner (llm-checker, lazy-checker, nitpicker) against `frontend/src/components/shared/log-table/column-picker.tsx`. All findings predate any current work. Two of the three checkers returned CLEAN — the file has no comment bloat, no dead code, and no deferred debt. What remains is styling/typing hygiene.

## 1. Duplicated hover/focus affordance classes across the two buttons

The trigger button (line 26) and the "Reset to defaults" button (line 63) both carry the same verbatim affordance string:

```
hover:text-foreground focus-visible:outline focus-visible:outline-2 focus-visible:outline-primary
```

plus the same `cursor-pointer border-none bg-transparent transition-colors` fragments. The two are also authored inconsistently — the trigger wraps its classes in `cn(...)` with two segments, the reset button is a single flat string with no `cn()`. A future change to the focus ring has to be made twice, by hand, with nothing to catch drift.

**Recommended fix (committed direction):** hoist the shared affordance classes into one module-level constant at the top of `column-picker.tsx` and route both buttons' `className` through `cn()` for consistency.

Rationale for choosing this over converting both to `<Button variant="ghost">` from `@/components/ui/button`: the shadcn route matches the CLAUDE.md "use the primitives instead of raw markup" convention on paper, but `ghost` is a background-hover, bordered, fixed-height variant, so adopting it would visibly change both controls and drag PR screenshot evidence along with it. The sibling components in this same directory (`log-detail-drawer.tsx`, `use-log-table.tsx`) also use raw `<button>` for these bare inline affordances, so the constant-hoist is both the lowest-risk and the most consistent-with-neighbors option, and it's a pure-mechanical zero-visual-change diff.

## 2. `cursor-pointer` stays active on disabled column rows

`column-picker.tsx:47` — the `<label>` carries an unconditional `cursor-pointer` while already conditionally muting its text via `has-[:disabled]:text-muted-foreground`. For required columns (`level`, `message`) and viewport-hidden columns the checkbox is `disabled`, so the row cannot be toggled, but the cursor still promises it can. The disabled affordance is half-applied.

**Recommended fix:** add `has-[:disabled]:cursor-default` alongside the existing `has-[:disabled]:text-muted-foreground` on the same element, reusing the pattern already present one utility over.

## 3. `py-0` is a no-op utility

`column-picker.tsx:47` — Tailwind Preflight applies no vertical padding to `<label>`, and nothing in `PopoverContent` or `global.css` adds any, so `py-0` overrides nothing.

**Recommended fix:** delete `py-0` from the label's class list. Zero visual change.

## 4. `ColumnPickerProps` is a byte-identical duplicate of this component's own `Props`

`use-log-table.tsx:44-49` declares:

```ts
export interface ColumnPickerProps {
  selectedColumns: ColumnId[];
  viewportHidden: ReadonlySet<ColumnId>;
  onToggle: (id: ColumnId) => void;
  onReset: () => void;
}
```

which is field-for-field identical to the local `Props` interface at `column-picker.tsx:9-14`. Two declarations of one contract that must stay in sync manually — adding a prop to the picker silently leaves the hook's exported type stale.

**Recommended fix:** rename the local interface in `column-picker.tsx` to `ColumnPickerProps`, export it, and have `use-log-table.tsx` import it instead of redeclaring. The component owns its own prop contract, so it should be the single source; the hook already imports from this directory. Optionally collapse `pages/logs.tsx:63-68`'s four one-by-one prop passes into `{...log.columnPickerProps}` in the same edit.

## Acceptance Criteria

- [ ] Shared hover/focus affordance classes live in one module-level constant in `column-picker.tsx`; both buttons use `cn()` consistently
- [ ] Disabled column rows no longer show `cursor-pointer`
- [ ] `py-0` removed from the label
- [ ] `ColumnPickerProps` declared and exported once, in `column-picker.tsx`, and imported by `use-log-table.tsx`
- [ ] `column-picker.test.tsx` still passes; no visual change to the picker beyond the disabled-row cursor

NodeJSmith:
## Summary

Removes the styling and prop-type duplication in the log-table column picker surfaced by the automated quality scanner. Almost entirely mechanical — the only rendered change is that disabled column rows stop showing a pointer cursor.

## What changed

**`frontend/src/components/shared/log-table/column-picker.tsx`**

- The popover trigger and the "Reset to defaults" button carried the same affordance classes verbatim (`cursor-pointer border-none bg-transparent`, muted-until-hover text, and the `focus-visible` ring), authored inconsistently — the trigger wrapped them in `cn()`, the reset button was one flat string. Both now compose a single module-level `BARE_BUTTON_CLASSES` constant through `cn()`, so changing the focus ring is one edit instead of two hand-synced strings with nothing to catch drift.
- Added `has-[:disabled]:cursor-default` to the column row `<label>`. Required columns (`level`, `message`) and viewport-hidden columns render a disabled checkbox, so the row could not be toggled, but the cursor still advertised that it could. This is the one user-visible change in the PR.
- Dropped `py-0` from that same label — Preflight gives `<label>` no vertical padding and nothing in `PopoverContent` or `global.css` adds any, so it overrode nothing.

**`frontend/src/components/shared/log-table/use-log-table.tsx`**

- `ColumnPickerProps` was declared field-for-field twice: once as a local `Props` in `column-picker.tsx`, once exported from `use-log-table.tsx`. Adding a prop to the picker would silently leave the hook's exported type stale. The component now owns and exports the interface, and the hook imports it.

### Why not `<Button variant="ghost">`

CLAUDE.md prefers the shadcn primitives over raw markup, but `ghost` is a background-hover, bordered, fixed-height variant — adopting it would visibly change both controls and turn a zero-visual-change cleanup into a redesign. The sibling components in this directory (`log-detail-drawer.tsx`, `use-log-table.tsx`) also use raw `<button>` for these bare inline affordances, so hoisting the constant is both the lower-risk option and the one consistent with its neighbors.

## Not included

The issue listed one **optional** follow-on: collapsing `pages/logs.tsx`'s four one-by-one prop passes into `{...log.columnPickerProps}`. Left as-is to keep this diff to the four required acceptance criteria.

## Testing

- `prek -a` — all hooks pass, including `tsc`, `eslint`, `prettier`, and `stylelint` on the frontend.
- `npx vitest run src/components/shared/log-table/` — 9 files, 210 tests, all passing. This covers both changed files, including the dedicated `column-picker.test.tsx`.

Scoped deliberately to the affected area; CI runs the full backend, system, and frontend suites across every supported Python version.

## Screenshots

Column picker open, dark and light — visually unchanged apart from the disabled-row cursor:

Column picker, dark

Column picker, light

Logs page with the picker in context:

Logs page after

Closes #2230


NodeJSmith:
@coderabbitai review

NodeJSmith:
## Automated review pass

Ran the full review battery (code, integration, WTF, llm-checker, lazy-checker, nitpicker) against `origin/main`. CodeRabbit and Codex both returned zero findings.

**Independently verified the two substantive claims in the description:**

- **`has-[:disabled]:cursor-default` actually wins.** Built the CSS and inspected the output: `.has-\[\:disabled\]\:cursor-default:has(:disabled)` is emitted at specificity (0,2,0) versus `.cursor-pointer` at (0,1,0), and lands later in the stylesheet — so it wins on both specificity and source order. The label's `className` is a plain string literal, so tailwind-merge never gets a chance to strip either class.
- **`py-0` was genuinely a no-op.** Tailwind v4 Preflight zeroes padding via the universal selector, which covers `<label>`; nothing in `global.css` or `PopoverContent` re-adds any.

**Fixed in e1caaee:**

- Renamed `BARE_BUTTON_CLASSES` → `BARE_BUTTON_CLASS`. Every other single-string class constant in the frontend uses the singular suffix; the only `_CLASSES` constant (`TONE_CLASSES`) is a `Record`, a different shape.
- Reworded the constant's comment to state the invariant it protects instead of re-narrating its own utility classes.
- Added a regression test for the one user-visible change. Verified RED (fails when the class is removed) then GREEN. Full frontend suite: 1624 tests passing.

**Deliberately not changed — flagging for a human call:**

- **`ColumnPickerProps` ownership direction.** Two reviewers noted this type now flows component → hook, while its siblings `LogTableViewProps` and `LogDrawerProps` flow hook → component. The inconsistency is real, but the PR's direction (component owns its own props) is the more idiomatic React placement and the more discoverable one, and it's what #2230 asked for. Flipping it to match the siblings would move a props type away from its component and re-open the question for the other two. Left alone rather than churned in an autofix PR — worth settling for all three types at once if it matters.
- **Utility ordering across the `cn()` arguments.** Hoisting a shared constant necessarily puts the shared classes first, so per-site layout utilities now trail the constant's interactive-state ones. Cosmetic only: there are no tailwind-merge conflicts between the two argument groups, and the repo has no Tailwind class-sorting Prettier plugin to enforce an order either way.

NodeJSmith:
## Automated review pass (post-`e1caaee`)

Re-ran the full battery — code-reviewer, integration-reviewer, wtf-reviewer, llm-checker, lazy-checker, nitpicker — against the current head rather than the state the earlier pass reviewed. No unresolved CodeRabbit threads; Codex left a 👍 with no findings.

**code-reviewer: PASS** — 0 critical/high/medium. I independently re-verified the class-output equivalence by hand rather than taking the earlier pass's word for it: both `cn(BARE_BUTTON_CLASS, ...)` call sites produce a utility set *identical* to the pre-refactor inline strings (13 utilities for the trigger, 12 for the reset button), and `BARE_BUTTON_CLASS` contributes no display/padding/rounded utility, so tailwind-merge has nothing to drop in either call.

**Fixed in `65362f7a`** — the one finding with a concrete, unambiguous fix. The new test's comment claimed the assertion proved `has-[:disabled]` "outranks the base cursor-pointer on specificity." jsdom applies no Tailwind CSS, so the assertion pins the utility's *presence*, not the computed cursor — the comment overclaimed what the test establishes. Reworded it to the real claim, and renamed the test from `disabled rows drop the pointer cursor via has-[:disabled]:cursor-default` to `disabled rows do not advertise a pointer cursor`, matching the behavior-describing style of every sibling test in the same block. Re-verified RED (removing the class fails the test) then GREEN; 15/15 passing, `tsc --noEmit` and eslint clean.

**Not changed, with reasons:**

- **Test asserts a class string** (flagged by 4 of 6 reviewers). Kept. jsdom genuinely cannot compute Tailwind CSS, so there is no stronger assertion available here, and the pattern has strong precedent — 14 frontend test files assert on class names, including siblings in this same directory. The fix was the misleading comment, not the assertion.
- **`ColumnPickerProps` ownership direction.** Unchanged and still worth a human call, as flagged in the description. Reviewers split the same way: the inconsistency with `LogTableViewProps`/`LogDrawerProps` is real, but component-owns-its-props is the better placement in isolation.
- **`BARE_BUTTON_CLASS` comment.** llm-checker called it narration; wtf-reviewer praised it as exactly the self-contained-comment style this repo asks for. Sided with the latter — "so the pair cannot drift apart" states a durable invariant, not an incident.

**Filed #2233** for the one finding that is real but genuinely out of scope here. Three reviewers noted this PR's dedup stops at the file boundary — I verified it directly: the `focus-visible:outline focus-visible:outline-2 focus-visible:outline-primary` fragment is hand-inlined at **18 sites across 13 files**, including four in `log-detail-drawer.tsx` in this very directory. That predates this PR and #2230 was scoped to the column picker, so consolidating it here would have been scope creep.

NodeJSmith:
@coderabbitai review

PR #2246, opened 2026-09-15
fix: orphan-mark crashed sessions that never finalized

### LINKED ISSUE #1806: Orphan-mark crashed sessions that never finalized
## Description

A session that records a crash and then dies before `finalize_session` runs is invisible to every cleanup path afterward:

- `_do_on_service_crashed` (`session_manager.py:167-187`) sets `status='failure'` but leaves `stopped_at` NULL.
- `_do_mark_orphaned_sessions` (`session_manager.py:143-152`) only matches `status='running'`, so the next boot never orphan-marks the crashed session.
- The once=True listener cleanup's liveness join (`session_manager.py:222-227`) treats any session with `stopped_at IS NULL` as live, so that session's once-listener rows are never eligible for cleanup — permanently.

## Steps to Reproduce

1. Run hassette; let a service crash (CRASHED event recorded, `status='failure'`, `stopped_at` NULL).
2. Kill the process before graceful teardown (`finalize_session` never runs).
3. Restart. `mark_orphaned_sessions` skips the row (status is 'failure', not 'running'); `stopped_at` stays NULL forever.
4. once=True listeners with executions in that session are never cleaned up.

## Decision (2026-09-14)

Widen `_do_mark_orphaned_sessions()` to match `stopped_at IS NULL` (any status) — not have the crash-recording UPDATE set `stopped_at` itself.

**Why not set `stopped_at` at crash time:** a service crash does not mean the session (process) ends — `ServiceWatcher` can restart a crashed service while the session keeps running (the restart-supervision design this framework is built around). Stamping `stopped_at` at crash time would make a still-alive session with one recovered service look "stopped" to every other consumer of that field, including the once-listener cleanup's own liveness check for *fired* listeners — which would then start reaping listeners still actively firing in a live session. That's a new, worse bug than the one being fixed.

Widening the orphan-marking match is also consistent with what `_do_mark_orphaned_sessions()` already does for the "we don't know exactly when it died" problem: it backfills `stopped_at = last_heartbeat_at` as the best available estimate — applies equally well to a crashed-then-killed session as to a stuck-`'running'` one.

**Refinement to the acceptance criteria's literal wording:** the current query unconditionally overwrites `status` to `'unknown'`. A crashed session already recorded real diagnostic info (`error_type`/`error_message`/`error_traceback`, status `'failure'`) — overwriting that with `'unknown'` throws away more specific information for no reason. Only force `status → 'unknown'` for rows that were still `'running'` (truly unknown outcome at time of death); leave an already-`'failure'` status alone while still backfilling its `stopped_at`.

## Fix Shape

**`session_manager.py`** — `_do_mark_orphaned_sessions()`:

```python
async def _do_mark_orphaned_sessions(self) -> None:
    """Execute the orphan-session UPDATE; called by the write-queue worker.

    Matches any session left with stopped_at IS NULL from a previous process run --
    not just status='running'. A session that recorded a crash (status='failure') but never
    reached finalize_session() before the process died is just as orphaned as one still
    marked 'running'; both need stopped_at backfilled so downstream liveness checks (the
    once=True listener cleanup's session join, in particular) treat them as ended. Only rows
    still 'running' get their status overwritten to 'unknown' -- a 'failure' row already
    recorded a real outcome and keeps it, gaining only the backfilled stopped_at.
    """
    db = self._database_service.db
    cursor = await db.execute(
        """
        UPDATE sessions
        SET
            status = CASE WHEN status = :old_status THEN :new_status ELSE status END,
            stopped_at = last_heartbeat_at
        WHERE stopped_at IS NULL
        """,
        {"new_status": SESSION_STATUS_UNKNOWN, "old_status": SESSION_STATUS_RUNNING},
    )
    if cursor.rowcount and cursor.rowcount > 0:
        self.logger.warning("Marked %d orphaned session(s) as ended", cursor.rowcount)
    await db.commit()
```

(Log message updated from "Marked N orphaned session(s) as 'unknown'" since not every matched row's status changes to 'unknown' anymore.)

No change needed to `_do_on_service_crashed()` or the once-listener cleanup's `sessions.stopped_at IS NULL` check itself — once orphan-marking backfills `stopped_at` for these rows on the next boot, the existing liveness join correctly treats them as ended without any further change.

## Key Items

- `src/hassette/core/session_manager.py:143-152` — `_do_mark_orphaned_sessions()` query and log message.

## Acceptance Criteria

- [ ] Orphan-marking matches `stopped_at IS NULL` regardless of status.
- [ ] A session already recorded as `status='failure'` keeps that status after orphan-marking — only `stopped_at` is backfilled, not overwritten to `'unknown'`.
- [ ] A session still `status='running'` is overwritten to `'unknown'` on orphan-marking, unchanged from today.
- [ ] once=True cleanup becomes eligible for rows belonging to crashed-then-killed sessions on the next boot.
- [ ] Test: session with `status='failure'`, `stopped_at` NULL from a "previous run" → orphan-marked on startup (status unchanged, `stopped_at` backfilled) and its once-listener rows (that have fired) become cleanup-eligible.
- [ ] Test: session with `status='running'`, `stopped_at` NULL from a "previous run" → orphan-marked to `status='unknown'`, `stopped_at` backfilled (existing behavior, regression-proof this still works).
- [ ] `uv run pytest tests/unit/core tests/integration -n 4` passes.

NodeJSmith:
## Summary

Widens `_do_mark_orphaned_sessions()` to match any session with `stopped_at IS NULL`, not just those with `status='running'`. This fixes a bug where a session that recorded a crash (`status='failure'`) but died before `finalize_session()` ran was invisible to orphan cleanup — its `stopped_at` stayed NULL forever, permanently blocking `once=True` listener cleanup for that session's rows.

## What changed

**`src/hassette/core/session_manager.py`** — The orphan-marking query now uses `WHERE stopped_at IS NULL` instead of `WHERE status = :old_status`. A `CASE WHEN` conditionally sets status: only rows still `'running'` (truly unknown outcome) get overwritten to `'unknown'`; `'failure'` rows keep their real diagnostic info (`error_type`, `error_message`, `error_traceback`) and gain only the backfilled `stopped_at`. The log message is updated from "as 'unknown'" to "as ended" to reflect that not all marked rows change status.

**`tests/integration/database/test_session_manager.py`** — Two new integration tests:
- `test_mark_orphaned_sessions_backfills_crashed_session` — verifies a crashed session gets `stopped_at` backfilled while preserving its `'failure'` status and error fields
- `test_crashed_session_once_listeners_eligible_after_orphan_mark` — end-to-end: inserts a crashed session with a fired `once=True` listener, orphan-marks it, then runs cleanup and verifies the listener is deleted

## Why not set `stopped_at` at crash time

A service crash doesn't end the session — `ServiceWatcher` can restart a crashed service while the session continues. Stamping `stopped_at` at crash time would make a still-alive session look "stopped" to the once-listener cleanup, causing it to reap listeners still actively firing. Widening the orphan-marking match avoids that and is consistent with how it already handles the "unknown death time" problem: backfilling `stopped_at = last_heartbeat_at`.

## Testing

- `prek -a` + `prek pyright -a --stage pre-push`: all checks pass
- `pytest tests/integration/database/test_session_manager.py -v -n 4`: 16/16 passed (3.28s), including the 2 new tests
- CI runs the full backend + system + e2e suites on every PR

Closes #1806


NodeJSmith:
@coderabbitai review

PR #2247, opened 2026-09-15
fix: prevent logger filter accumulation across resource reconstruction

### LINKED ISSUE #1808: Prevent logger filter accumulation across resource reconstruction
## Description

`Resource._setup_logger` (`resources/base.py:198-212`) runs on every Resource construction and does:

```python
self.logger = getLogger(self._get_logger_name())
...
self.logger.addFilter(_ResourceContextFilter(self.source_tier))
```

Loggers are process-global singletons keyed by name. `_ResourceContextFilter` defines no `__eq__`, so stdlib `Filterer.addFilter`'s `if not (filter in self.filters)` dedupe never matches — every construction appends a fresh filter instance. App instances recreated on hot-reload derive the same logger name (parent-based `unique_name`), so a hot-reload-heavy dev session accumulates one filter per reload, unboundedly: a slow memory leak plus an O(n) filter chain on every log record through that logger. (`setLevel` on the shared logger is also re-applied per construction — harmless, but same shape.)

## Decision (2026-09-14)

Value-based `__eq__` (and `__hash__`) on `_ResourceContextFilter` — not a type-based dedupe check in `_setup_logger()`. Checked `source_tier`: it's derived from the resource's role/identity, so every reconstruction sharing a logger name (a hot-reloaded app keeps the same `unique_name` → same logger name → same `source_tier`) always constructs the filter with the same value. That makes the two options behaviorally equivalent here, and the `__eq__` approach is simpler: stdlib's `logging.Filterer.addFilter()` already does `if filter not in self.filters: self.filters.append(filter)` — it just never matches today because `_ResourceContextFilter` has no `__eq__` (falls back to identity). Adding value-based `__eq__` makes the existing stdlib dedup work with zero changes to `_setup_logger()` itself.

`__hash__` needs to be defined alongside `__eq__` — Python sets `__hash__ = None` automatically on a class that defines `__eq__` without it, which would make `_ResourceContextFilter` instances unhashable. Nothing currently hashes a `Filter` (stdlib's `Filterer.filters` is a plain list), but leaving instances hashable is cheap and avoids a surprise if that ever changes.

## Fix Shape

```python
class _ResourceContextFilter(Filter):
    """Stamps source_tier on every LogRecord so downstream handlers and structlog processors can read it."""

    def __init__(self, source_tier: str) -> None:
        super().__init__()
        self.source_tier = source_tier

    def filter(self, record: LogRecord) -> bool:
        record.source_tier = self.source_tier  # pyright: ignore[reportAttributeAccessIssue]
        return True

    def __eq__(self, other: object) -> bool:
        return isinstance(other, _ResourceContextFilter) and self.source_tier == other.source_tier

    def __hash__(self) -> int:
        return hash((_ResourceContextFilter, self.source_tier))
```

No change needed to `_setup_logger()` — `self.logger.addFilter(_ResourceContextFilter(self.source_tier))` already relies on stdlib's own `filter not in self.filters` check, which now correctly matches.

## Key Items

- `src/hassette/resources/base.py:88-97` — add `__eq__`/`__hash__` to `_ResourceContextFilter`.

## Acceptance Criteria

- [ ] `_setup_logger` does not accumulate duplicate filters across repeated constructions with the same logger name (achieved via `_ResourceContextFilter.__eq__`, not a change to `_setup_logger` itself).
- [ ] Test: constructing two Resources sharing a logger name leaves exactly one context filter on the logger.
- [ ] Test: two `_ResourceContextFilter` instances with different `source_tier` values are not considered equal (guards against over-collapsing if `source_tier` ever legitimately differs).
- [ ] `uv run pytest tests/unit/core -n 4` passes.

NodeJSmith:
## Summary

Add value-based `__eq__` and `__hash__` to `_ResourceContextFilter` so that stdlib
`Filterer.addFilter()`'s existing dedup (`filter not in self.filters`) correctly
collapses identical filters when a Resource sharing a process-global logger name is
reconstructed (e.g. on hot-reload).

Previously, `_ResourceContextFilter` had no `__eq__`, so every `Resource` construction
appended a fresh filter instance to the shared logger — unbounded accumulation across
hot-reloads (slow memory leak + O(n) filter chain per log record).

## What changed

**`src/hassette/resources/base.py`** — added `__eq__` (checks `source_tier` equality)
and `__hash__` (tuple of class + `source_tier`) to `_ResourceContextFilter`. No change
to `_setup_logger()` itself — the stdlib dedup that was always there now works.

**`tests/unit/resources/test_resource_context_filter.py`** (new) — 4 tests:
- Value equality: same `source_tier` → equal, same hash
- Value inequality: different `source_tier` → not equal
- Type safety: not equal to non-`_ResourceContextFilter` objects
- Accumulation guard: two `ConcreteResource` instances sharing a parent (same logger
  name) leave exactly one context filter on the logger

## Testing

Ran locally:
- `pytest tests/unit/resources/test_resource_context_filter.py` — 4/4 passed
- `pytest tests/unit/resources/` — 209/209 passed

CI runs the full backend + system + e2e suite on every PR.

Closes #1808


NodeJSmith:
@coderabbitai review

PR #2254, opened 2026-09-16
refactor(frontend): rename lastLabel to lastValue and guard fixed stat-cell count

### LINKED ISSUE #1905: Clean up naming and duplicated logic in stat-cell-builders.ts
Automated quality scan of `frontend/src/components/app-detail/stat-cell-builders.ts`. Re-verified against current code — findings 2, 4, and 6 are already resolved by PR #2142 (`refactor(frontend): decouple stat cell labels from lookup literals`) and PR #2189 (`fix(frontend): export COMMON_STAT_CELL_COUNT`). Findings 1, 3, and 5 remain.

## 1. Inverted naming: `lastLabel` holds a value, `lastFieldLabel` holds a label (still live, mechanical)

`CommonStatInput.lastLabel` (line ~12) holds the *value* displayed in the "last" cell, while `lastFieldLabel` (line ~15) holds the column label. Rename `lastLabel` → `lastValue`. Touches `stat-cell-builders.ts`, `stat-cell-builders.test.ts`, `job-detail.tsx`, `listener-detail.tsx`.

## 2. Duplicated magic strings (RESOLVED — PR #2142)

`LABEL_CANCELLED`/`LABEL_TIMED_OUT` constants now exist (lines 6-7) and are referenced at both `resolveExtraCellIndex` and `buildCommonStatCells`. No action needed.

## 3. `COMMON_STAT_CELL_COUNT` can drift from the fixed-cell array — decided: keep the constant, add an assertion, don't derive-and-remove

`COMMON_STAT_CELL_COUNT = 5` (line 4-5) is exported and documented (PR #2189), but is still an independent literal from the 5-element array literal in `buildCommonStatCells` (lines 41-47) — nothing ties the two together, so an edit adding or removing a fixed cell can silently desync them.

**Checked why "derive from `cells.length` and drop the constant" isn't safe:** `COMMON_STAT_CELL_COUNT` is imported directly by `stat-cell-builders.test.ts` (lines 4, 102, 117, 133, 146) to slice off the conditional cells in test assertions — it is not just an implementation detail consumed only within this file, it's part of this module's tested surface. Deriving it away and deleting the export would break that test file, not just the sibling #1904 issue that also imports it.

**Decision: keep the exported constant, add a runtime assertion tying it to the actual array.** Immediately after constructing the fixed `cells` array in `buildCommonStatCells` (after line 47), add an assertion (e.g. `if (cells.length !== COMMON_STAT_CELL_COUNT) throw new Error(...)`, or an equivalent dev-time check) so a future edit that changes the fixed-cell count without updating the constant fails loudly instead of drifting silently. `resolveExtraCellIndex`'s existing fallback to `COMMON_STAT_CELL_COUNT` (line 38) stays as-is — the assertion is what closes the drift risk, not a structural change to how the index is computed.

## 4. Duplicated ternary for `failed` tone (RESOLVED — PR #2142)

`failedTone` is now extracted once (line 41) and referenced at both cell definitions (lines 43-44). No action needed.

## 5. Inconsistent `Count` suffix on quantity fields (DECIDED: keep as-is, mirrors backend)

`suppressedCount`/`droppedCount` use a `Count` suffix; `failed`/`timedOut`/`cancelled`/`threadLeaked` don't. Traced both groups to `src/hassette/web/models.py:357-386`: the backend fields are `failed`, `cancelled`, `timed_out`, `thread_leaked` (no `_count` suffix) vs. `suppressed_count`, `dropped_count` (with it) — the frontend interface is a direct camelCase mirror of the backend field names. Unifying the suffix locally would break the 1:1 backend-to-frontend field correspondence that makes it possible to grep from a UI field straight to its backend source. No change.

## 6. Oversized doc comment (RESOLVED — already condensed)

The `extraCell` JSDoc (lines ~21-26) is now 5 focused lines stating the priority order and the builder's ownership of the insertion point — not the 12-line, editorial-aside version originally flagged. No further action needed.

## Acceptance Criteria

- [ ] `lastLabel` renamed to `lastValue` across `stat-cell-builders.ts`, `stat-cell-builders.test.ts`, `job-detail.tsx`, `listener-detail.tsx`
- [x] Finding 3 decided: keep `COMMON_STAT_CELL_COUNT` exported (it's a tested-surface constant, not derivable-and-removable), add a runtime assertion tying it to the actual fixed-cell array length instead.
- [ ] Assertion added in `buildCommonStatCells` immediately after the fixed `cells` array is built.
- [x] Duplicated magic strings — resolved by PR #2142
- [x] Duplicated `failedTone` ternary — resolved by PR #2142
- [x] `Count`-suffix naming — declined; fields intentionally mirror backend schema names
- [x] Oversized `extraCell` doc comment — already condensed, no further action

NodeJSmith:
## Summary

Two mechanical cleanups in `frontend/src/components/app-detail/stat-cell-builders.ts`, both from the quality scan in #1905.

### 1. Rename `lastLabel` → `lastValue`

`CommonStatInput` had two adjacent fields whose names read backwards:

- `lastLabel` held the **value** rendered in the "last" cell (`"2m ago"`, `"next in 5m"`)
- `lastFieldLabel` held that cell's **column label** (`"Last"`, `"Next"`)

So the field named `...Label` was the value, and the field named `...FieldLabel` was the label. Renaming `lastLabel` to `lastValue` makes the pair self-describing and matches how the builder actually consumes them:

```ts
{ label: input.lastFieldLabel ?? "Last", value: input.lastValue }
```

Updated at all four call sites: `stat-cell-builders.ts`, `stat-cell-builders.test.ts`, `job-detail.tsx`, `listener-detail.tsx`. No behavior change — the rendered label and value are identical.

### 2. Pin `COMMON_STAT_CELL_COUNT` against drift

`COMMON_STAT_CELL_COUNT = 5` is a hand-maintained literal describing the fixed-cell array built in `buildCommonStatCells`. Nothing tied the two together, so adding or removing a fixed cell without updating the constant would desync `resolveExtraCellIndex`'s fallback offset — the `extraCell` would land in the wrong slot.

This pins the two together in a test:

```ts
it("keeps COMMON_STAT_CELL_COUNT in sync with the fixed cells actually built", () => {
  expect(buildCommonStatCells(baseInput())).toHaveLength(COMMON_STAT_CELL_COUNT);
});
```

An earlier revision enforced this with a runtime `throw` inside `buildCommonStatCells`. Review replaced it: the condition could only ever fire from a future edit to the cell literal directly above it, never from runtime data, so it left a permanently-dead branch in a render path (patch coverage 33%) and was the only non-boundary `throw` in `frontend/src` — every other one validates external HTTP/WS data. Verified that drift already fails five existing tests without it; the new test just makes that failure name the constant rather than surfacing as unrelated assertion mismatches.

Per the decision recorded in #1905, the constant stays exported rather than being derived away: `stat-cell-builders.test.ts` imports it directly (5 call sites) to slice conditional cells off in assertions, so it is part of this module's tested surface.

### 3. Finish the value/label rename upstream

`LastCellDisplay` in `job-detail.tsx` still named its display-text field `label` alongside a sibling `fieldLabel` — the same conflation this PR removes from `CommonStatInput`, which left `lastValue: lastCell.label` reading as a mismatch. Renamed to `value` (interface is local to that file, 5 references). Also added a doc comment to `lastValue`, matching its paired `lastFieldLabel`.

The remaining scanner findings were already resolved or explicitly declined upstream (see the issue): duplicated label strings and the `failedTone` ternary landed in #2142, the `Count`-suffix naming is a deliberate 1:1 mirror of the backend field names in `web/models.py`, and the `extraCell` doc comment was already condensed.

## Testing

- `prek -a` — all hooks pass (ruff, eslint, stylelint, prettier, **tsc**, breakpoint-drift and dead-token checks)
- `npx vitest run src/components/app-detail/` — **18 files, 293 tests, all passing**, covering `stat-cell-builders.test.ts` plus the `job-detail` / `listener-detail` consumers via the `handlers-tab` suites

Scoped locally to the touched area; CI runs the full backend + system + frontend suite across every supported Python version.

## Screenshots

No visual change — this is a rename plus an internal invariant check. Captured to confirm both detail views render their stat rows identically after the rename:

**Listener detail stats**

Listener detail stats

**Job detail stats** (exercises the `lastFieldLabel: "Next"` override path)

Job detail stats

Closes #1905


NodeJSmith:
@coderabbitai review

PR #2271, opened 2026-09-17
refactor: drop unreachable severity handling in boot issues panel

### LINKED ISSUE #1934: Clean up speculative severity handling and accessibility in boot-issues-panel
## Description

Quality scanner surfaced findings in the boot issues panel. `BootIssueResponse.severity` (generated-types.ts:949) is the closed union `"err" | "warn"`, so the dead-code findings are still valid. Two findings needed investigation before a shape could be picked; both are resolved below with evidence.

## Key Items

- **Dead code in severity map** (line 6-7): `SEVERITY_ORDER` includes an unreachable `info: 2` entry and `UNKNOWN_SEVERITY_SORT_ORDER` provides a fallback for a value that can never occur. Remove both, and the `??` fallback on lines 16-17.

- **Loose typing of `SEVERITY_ORDER`**: type it against `BootIssue["severity"]` instead of `Record<string, number>`, so a future dead entry is caught at compile time.

- **No-op ternary** (line 29): `issue.severity === "err" ? "err" : "warn"` always equals `issue.severity`. Replace with `kind={issue.severity}`.

- **Index in list key (DECIDED: keep the index)**: traced `label` back to `runtime_query_service.py:385,393` — it's built as `f"App blocked: {manifest.display_name}"` / `f"App failed: {manifest.display_name}"`, and `manifest.display_name` (`config/classes.py:130,168-169`) has no cross-app uniqueness constraint (it defaults to `class_name`/`app_key` but can be explicitly set by user config, with no dedup check). `severity+label` collisions are genuinely possible if two apps share a display name. **Keep the array-index prefix in the key** (`` `${i}-${issue.severity}-${issue.label}` ``) as the collision tie-breaker — no change here.

- **Redundant aria-label (DECIDED: use a distinct sublabel, don't remove)**: checked every sibling diagnostics panel — `logging-panel.tsx` ("Logging health" / "Log drop counters"), `services-panel.tsx` ("Internal services" / "Service list"), `telemetry-panel.tsx` ("Telemetry health" / "Drop counters") — all use a **distinct** sublabel on the inner list, never repeating the outer `Panel`'s `ariaLabel`. `boot-issues-panel.tsx` is the outlier with both set to "Boot issues". Fix by renaming the `<ul>`'s `aria-label` to a distinct value (e.g. `"Boot issue list"`), matching the established sibling pattern — do not remove it.

## Acceptance Criteria

- [ ] `SEVERITY_ORDER` typed against `BootIssue["severity"]`; `info` entry removed
- [ ] `UNKNOWN_SEVERITY_SORT_ORDER` constant and `??` fallback removed
- [ ] Ternary on line 29 replaced with `kind={issue.severity}`
- [ ] List `key` unchanged (index prefix retained — `label` is not guaranteed unique across apps)
- [ ] `<ul>`'s `aria-label` renamed to a distinct sublabel (e.g. "Boot issue list"), matching the pattern in `logging-panel.tsx`/`services-panel.tsx`/`telemetry-panel.tsx` — not removed

NodeJSmith:
## Summary

`BootIssueResponse.severity` is the closed union `"err" | "warn"` (see `severity: "err" | "warn"` in `frontend/src/api/generated-types.ts`), but `boot-issues-panel.tsx` carried three pieces of handling for severities that can never arrive:

- an `info: 2` tier in `SEVERITY_ORDER`
- an `UNKNOWN_SEVERITY_SORT_ORDER = 99` fallback behind `??`
- a `issue.severity === "err" ? "err" : "warn"` ternary that returns its own input

All three are removed. `SEVERITY_ORDER` is now typed as `Record<BootIssue["severity"], number>` rather than `Record<string, number>`, so if a future schema change adds a severity, the missing entry fails `tsc` instead of silently sorting to the end — and a dead entry fails too.

The panel's two tests that existed only to exercise the unreachable branches (both needed `@ts-expect-error` to construct their inputs) are replaced with one test that asserts the real behavior: errors sort ahead of warnings.

Also renames the inner `<ul>`'s `aria-label` from `"Boot issues"` to `"Boot issue list"`. It previously duplicated the outer `Panel`'s `ariaLabel`, which made the list and its containing region indistinguishable to a screen reader navigating by label. The sibling panels already use distinct sublabels — `logging-panel` (`"Log drop counters"`), `services-panel` (`"Service list"`), `telemetry-panel` (`"Drop counters"`).

### Deliberately unchanged

The list `key` keeps its array-index prefix (`` `${i}-${issue.severity}-${issue.label}` ``). Labels embed `manifest.display_name`, which carries no cross-app uniqueness constraint, so two apps can produce the same severity+label pair — the index is what keeps the key unique.

## Testing

No behavior change is intended, so verification focused on the type-level claim and the panel's consumers:

- `prek -a` — all hooks pass, including `tsc (frontend)`, the type checker that enforces the new `Record<BootIssue["severity"], number>` constraint. No Python files are touched, so `pyright` is not implicated.
- `npx vitest run src/components/diagnostics/` — 3 files, 10 tests passed (includes the rewritten panel test).
- `npx vitest run src/pages/diagnostics.test.tsx` — 26 tests passed (the page-level consumer of this panel).
- Confirmed by grep that nothing selects on the old `aria-label="Boot issues"` — every test targets `data-testid` (`diag-boot-*`), all of which are unchanged.

CI runs the full backend, system, and frontend suites across every supported Python version.

## Screenshots

Visually this is a no-op — the removed ternary returned its own input, and the `aria-label` is not rendered. Evidence that the panel still renders correctly:

Boot issues panel, 1280 light

Diagnostics page, 1280 dark

Closes #1934


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2275, opened 2026-09-17
fix: stop the code tab gutter from numbering a phantom trailing line

### LINKED ISSUE #1994: Fix gutter/header line-count mismatch in CodeTab, and clean up code-tab.test.tsx
Findings from an automated quality scan of `frontend/src/components/app-detail/code-tab.test.tsx` (llm-checker, lazy-checker, nitpicker). The file is generally in good shape — one shared render helper, no dead code, no TODO/FIXME debt, no commented-out blocks, clean imports. The items below are the surviving findings.

## 1. Weak assertion in "renders line numbers in gutter" — decided: real bug in `code-tab.tsx`, not just a weak test

`code-tab.test.tsx:87-92` asserts `expect(gutterLines.length).toBeGreaterThanOrEqual(1)` while the comment above it claims "Line numbers 1, 2, 3 should appear in gutter". The assertion would pass if the gutter rendered a single line, so it does not actually verify what the test name and comment promise.

**Root cause traced:** `defaultSource.content` ends in a trailing newline. The test's mocked `codeToHtml` (`code-tab.test.tsx:13-22`) does a naive `code.split("\n")` and wraps *every* resulting piece — including the trailing empty string produced by the trailing `\n` — in a numbered `<span class="line">`. So the gutter renders and numbers a phantom line 4. Meanwhile `code-tab.tsx:132-133` computes the displayed `lineCount` by explicitly stripping that trailing empty split entry, correctly showing "3 lines" in the header (`code-tab.tsx:154`). Gutter and header actively disagree on the same source.

This is `code-tab.tsx` production code, not a test-fixture-only artifact — `injectLineNumbers` (`code-tab.tsx:39-52`) numbers however many `<span class="line">` elements the highlighter output contains, with no relationship to `lineCount`. Whether real Shiki also emits a trailing empty line span for trailing-newline source (matching the mock) or drops it (making this a mock-only quirk) wasn't confirmed by running a real highlighter — but the fix below is correct either way.

**Decision: fix `injectLineNumbers` to cap at `lineCount`, not the raw span count from the highlighter.** This makes the gutter agree with the header regardless of whether the trailing-empty-span behavior originates in Shiki or only in the test's mock — worst case (if Shiki never emits a trailing empty span) this change is a no-op.

**Fix:**
- `code-tab.tsx`: compute `lineCount` once (already exists at lines 132-133) and pass it into `injectLineNumbers`, which should stop numbering/rendering a line once it exceeds `lineCount` (drop or leave un-numbered any trailing empty line span beyond that count).
- `code-tab.test.tsx:87-92`: replace `toBeGreaterThanOrEqual(1)` with an exact assertion — `toBe(3)` — now that the gutter is expected to match the header's line count exactly.
- This expands the issue's scope beyond the originally-scanned test file into `code-tab.tsx` itself, since the real defect lives there.

## 2. Route literal duplicated across three MSW handlers

`code-tab.test.tsx:61`, `:96`, and `:154` each inline `"/api/apps/:app_key/source"`. An endpoint rename requires three synchronized edits.

**Fix:** extract a single file-level constant (e.g. `const SOURCE_ENDPOINT = "/api/apps/:app_key/source";`) and reference it at all three `http.get(...)` call sites. The three handlers themselves should stay separate — they cover genuinely distinct cases (happy path, 404, delayed in-flight) and unifying them behind a builder would obscure rather than clarify.

## 3. Comments that restate the following line

Three one-line comments narrate code that already says the same thing:

- `:89` — `// Line numbers 1, 2, 3 should appear in gutter` (also inaccurate; see finding 1)
- `:136` — `// Wait for the focus effect to run`, above a `waitFor` whose assertion is self-explanatory
- `:164` — `// Wait for the request to be initiated`, above `waitFor(() => expect(requestSignal).toBeDefined())`

**Fix:** delete all three. The rationale-bearing comments in this file (`:10`, `:27`, `:36`, `:39`, `:132`) explain *why* and should be kept.

## 4. `body` is a misleading variable name

`code-tab.test.tsx:78` names `screen.getByTestId("code-tab-content").lastElementChild` as `body`. In a test file that also deals with HTTP responses, `body` reads as a false cognate for a response body.

**Fix:** rename to something that says what it is, e.g. `contentWrapper` or `shikiWrapper`.

## Acceptance Criteria

- [x] Finding 1 decided: fix `injectLineNumbers` in `code-tab.tsx` to cap at `lineCount`, not the raw highlighter span count. This is a production-code fix, not just a test change.
- [ ] `injectLineNumbers` updated to stop numbering/rendering beyond `lineCount`.
- [ ] `code-tab.test.tsx:87-92` asserts the exact gutter-line count (`toBe(3)`).
- [ ] `"/api/apps/:app_key/source"` appears exactly once in the file, as a named constant
- [ ] The three restating comments at `:89`, `:136`, `:164` are removed; rationale comments retained
- [ ] `body` renamed to a name that describes the element
- [ ] `cd frontend && npx vitest run src/components/app-detail/code-tab.test.tsx src/components/app-detail/code-tab.tsx` passes (note: also verify no other `code-tab.tsx` consumers/tests regress, since the fix touches production code)

NodeJSmith:
## Summary

The app detail **Code** tab rendered one more line number in the gutter than the header claimed the file had. Any source file ending in a trailing newline — which is nearly all of them — showed a phantom final line number with no code next to it, disagreeing with the "N lines" count in the header directly above it.

## Cause

Shiki emits one extra, empty `<span class="line">` when the source it highlights ends in a newline. (Confirmed against the real highlighter, not just the test mock — the mock did not reproduce this.) `injectLineNumbers` numbered every span it found, so it numbered that trailing empty one too. The header computes its count separately and strips the trailing empty entry, so the two never agreed.

## Fix

`CodeTab` already computes the correct `lineCount` for the header. That value is now passed into `injectLineNumbers`, which leaves any span past it un-numbered — the gutter stops exactly where the header says it does. One authoritative count, used by both.

## Also in this change

Small cleanups in the touched files, no behavior impact:

- Tightened the gutter test to assert the exact line count. It previously asserted `>= 1`, which is why it never caught this.
- Extracted the thrice-inlined `/api/apps/:app_key/source` endpoint into a `SOURCE_ENDPOINT` constant.
- Dropped three comments that just restated the line below them.
- Renamed `body` to `contentWrapper` in the theme-utilities test for clarity.

## Testing

Scoped to the affected area — CI runs the full suite:

- `prek -a` — all hooks pass (ruff, eslint, stylelint, prettier, tsc, and the repo's custom checks).
- `npx vitest run src/components/app-detail/code-tab.test.tsx` — 12/12 pass.
- `npx vitest run src/components/app-detail/` — 18 files, 296/296 pass.

The tightened assertion (`toBe(defaultSource.line_count)`) fails against the old `injectLineNumbers` and passes with the fix, so the test is real evidence rather than a restatement.

## Screenshots

**Before** — gutter runs one line past the header's count:

Code tab gutter before the fix

**After** — gutter and header agree:

Code tab gutter after the fix

Closes #1994


NodeJSmith:
@coderabbitai review

PR #2334, opened 2026-09-21
refactor: remove doubled label margin and share the field-label class in error-display components

### LINKED ISSUE #2169: Clean up error-display.tsx: double label spacing, mixed import styles, and duplicated label classes
## Description

A quality scan of `frontend/src/components/shared/error-display.tsx` surfaced three styling and import-hygiene findings. All predate the current work.

### 1. Double horizontal spacing between label and message — decided: drop `mr-2`

`error-display.tsx:59-60` — the flex container declares `gap-2`, and the label `<span>` additionally sets `mr-2`, producing an effective `1rem` gap instead of the `0.5rem` `gap-2` advertises. No other component in `frontend/src` combines `mr-2` with `flex ... gap-2` on the same axis — the codebase norm is to let `gap-2` alone own the spacing. **Decision: drop `mr-2` from the label span**, accepting the tighter `0.5rem` gap that `gap-2` already declares. Needs a screenshot in the PR per `.claude/rules/design-completeness.md` since this changes rendered spacing.

### 2. Mixed import styles within a single file

`error-display.tsx:1,3-4` — line 1 imports via the `@/` alias (`@/lib/utils`), while lines 3-4 use two-level relative paths (`../../api/generated-types`, `../../utils/format`) for the same kind of cross-directory import. The `@/*` alias is the majority convention in `components/shared` (37 files use it vs 22 still on `../../`).

Fix: switch lines 3-4 to `@/api/generated-types` and `@/utils/format` so the file uses one style.

### 3. Duplicated label styling

The field-label class string `font-mono text-xs uppercase tracking-[var(--text-label-tracking)] text-foreground-faint` appears verbatim in `error-display.tsx:60` and `traceback-viewer.tsx:75` (differing only in their layout-specific prefixes). Within `error-display.tsx` itself, `font-mono text-xs` is also repeated between the label span (line 60) and the message span (line 63).

Fix: hoist the shared label class fragment into a named constant that both components reference, keeping the per-site layout classes inline.

### Considered and not filed

The nitpicker also flagged the bare `Props` interface name (line 8) as non-descriptive. Not filed: bare `Props` is the dominant convention in `components/shared` (16 files use it), so renaming just this one would introduce inconsistency rather than remove it.

## Acceptance Criteria

- [ ] `error-display.tsx:60`'s label span no longer has `mr-2`; the flex container's `gap-2` is the sole horizontal spacing mechanism, confirmed by a screenshot in the PR
- [ ] `error-display.tsx` uses the `@/` alias for all cross-directory imports
- [ ] The shared field-label class string is defined once and referenced by both `error-display.tsx` and `traceback-viewer.tsx`
- [ ] `cd frontend && npm run build` and the existing `error-display.test.tsx` suite pass

NodeJSmith:
## Summary

Three small hygiene fixes in the shared error-display components, all confined to `frontend/src/components/shared/`.

## What changed and why

- **Dropped the redundant `mr-2`** on `ErrorDisplay`'s label span. The wrapping `div` is already `flex items-baseline gap-2`, so the margin stacked on top of the flex gap and made the label/message spacing wider than every other label pairing in the UI. The flex `gap-2` is now the sole horizontal spacing mechanism.
- **Switched the file's remaining two-level relative imports** (`../../api/generated-types`, `../../utils/format`) to the `@/` alias, matching the convention already used for `@/lib/utils` in the same file and across the rest of the frontend.
- **Hoisted the duplicated field-label class string** into a new `field-label.ts` as `FIELD_LABEL_CLASS`. `ErrorDisplay`, `TracebackViewer`, and `ExecutionLogs` all carried byte-identical copies of the same four-utility label styling. Per-site layout classes stay inline at each call site (`TracebackViewer` composes `"mb-2 block"` and `ExecutionLogs` composes `"mb-1 block"` via `cn()`), so only the genuinely shared styling moved.

No behavior change beyond the intended label spacing fix, and no API or schema surface is touched. The `ExecutionLogs` migration emits the identical class list it had before (`cn()` only concatenates here — there are no conflicting utilities for tailwind-merge to drop), so it renders unchanged.

## How it was tested

- `prek -a` — all hooks pass, including `eslint`, `prettier`, `stylelint`, and `tsc` (frontend).
- Scoped vitest run on the three affected components: `npx vitest run src/components/shared/error-display.test.tsx src/components/shared/traceback-viewer.test.tsx src/components/shared/execution-logs.test.tsx` — 3 files, 16 tests, all passing.
- CI runs the full backend + system + frontend suites across every supported Python version.

## Screenshots

`ErrorDisplay` label/message spacing after removing the doubled margin:

ErrorDisplay label gap

`TracebackViewer` label rendering unchanged after the class hoist — the `mb-2 block` layout classes still apply:

TracebackViewer label unchanged

Closes #2169


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2337, opened 2026-09-21
test: consolidate duplicate closing-spawn task bucket stubs onto shared factory

### LINKED ISSUE #2180: Consolidate duplicate closing-spawn task bucket stubs in test factories
## Description

Surfaced by the integration reviewer during the review pass on #2150 ("test: extract shared bus/scheduler service setup into test factories"). This predates that PR — `_default_spawn` is present on `origin/main` — but #2150 makes it visible by creating the canonical version of the same idiom a few dozen lines away in the same file.

`tests/support/factories.py` now contains two independent implementations of the "closing spawn" test-double idiom (close any coroutine handed to `spawn()` so it is never reported as *never awaited*, and hand back a mock stand-in for the task):

- `make_closing_task_bucket()` — the new shared, documented, allowlisted factory.
- `make_scheduler()`'s inline `_default_spawn` closure — a second, slightly different copy.

They are not equivalent, which is what makes the drift risk real:

| | `make_closing_task_bucket()` | `make_scheduler`'s `_default_spawn` |
|---|---|---|
| coroutine detection | `asyncio.iscoroutine(coro)` | `hasattr(coro, "close")` |
| returned task | one shared `MagicMock` | a fresh `Mock()` per call |
| `task.done()` | wired to `True` | not wired |
| `task.add_done_callback` | fires the callback synchronously | not wired |

A future fix to one (e.g. a bug in the coroutine-closing condition) is easy to apply to only one copy.

## Decision (2026-09-14)

Consolidate all three copies onto `make_closing_task_bucket()`. Investigation below found no consumer of `make_scheduler` (11 files) or `make_scheduler_service` (`tests/unit/core/_fixtures_bus_scheduler.py`) depends on the divergent behavior, so this is safe as a straight swap — not a behavior-preserving-with-parameter change.

**Investigation:**
- Of the 11 files importing `make_scheduler`, only `tests/unit/test_scheduler_resource.py` touches `scheduler.scheduler_service.task_bucket.spawn` at all — and it fully overrides `spawn` itself (`MagicMock(side_effect=lambda coro, **_: coro.close() or None)`) rather than relying on the factory's default. The other 10 files never reference `task_bucket`, `.spawn(`, `.done(`, or `add_done_callback` for a scheduler built via `make_scheduler`.
- `Scheduler`'s own methods (`remove_job`, `_on_job_removed`, the `remove` path) never call `self.scheduler_service.task_bucket.spawn(...)` — that only happens inside the real `SchedulerService` implementation (`src/hassette/core/scheduler_service.py`), which unit tests using `make_scheduler` don't exercise. The mocked `mock_service.remove_job`/`dequeue_job` side effects in the factory call `mark_job_removed` directly (awaited), not via `task_bucket.spawn`. So the "fresh `Mock()` per call" / unwired `done()`/`add_done_callback` behavior in the current `_default_spawn` is not observably exercised by any of the 11 consumers today.
- `tests/unit/core/_fixtures_bus_scheduler.py`'s `make_scheduler_service()` has the identical `hasattr(coro, "close")` closure (lines 84-90) — same conclusion applies; it's used to bypass `SchedulerService.__init__`, and nothing in its consuming tests asserts on the spawned-task mock's identity or wiring.
- `asyncio.iscoroutine(coro)` (the shared factory's check) still correctly detects the `AsyncMock()`-produced coroutine from `mark_job_removed(...)` that motivated the original `hasattr` check — `asyncio.iscoroutine` returns `True` for coroutine objects created by calling an `AsyncMock`, so the closing behavior is preserved, not just the mock identity behavior.

## Key Items

- `tests/support/factories.py:330-339` — `make_scheduler`'s inline `_default_spawn` duplicates `make_closing_task_bucket()`. Replace with `mock_service.task_bucket = make_closing_task_bucket()`.
- `tests/support/factories.py:388-414` — `make_closing_task_bucket()`, the canonical version. No change needed.
- `tests/unit/core/_fixtures_bus_scheduler.py:81-90` — `make_scheduler_service()`'s inline `_spawn`, a third copy. Replace with `svc.task_bucket = make_closing_task_bucket()` (import from `tests.support.factories`).
- `tests/unit/test_scheduler_resource.py` — already overrides `task_bucket.spawn` locally in each test that needs it; unaffected by the factory default changing, no edits required here.

## Acceptance Criteria

- [x] Confirmed no `make_scheduler` or `make_scheduler_service` consumer depends on the current per-call fresh `Mock()` return or on `done()`/`add_done_callback` being unwired — see investigation above.
- [ ] `make_scheduler` builds its `mock_service.task_bucket` from `make_closing_task_bucket()` instead of the inline `_default_spawn` closure.
- [ ] `tests/unit/core/_fixtures_bus_scheduler.py`'s `make_scheduler_service()` inline `_spawn` is replaced with `make_closing_task_bucket()`.
- [ ] `tests/support/helpers.py`'s `make_task_bucket()` is left as-is (real-task spawning is a deliberate contrast), with the distinction still documented.
- [ ] `uv run pytest tests/unit tests/integration -n 4` passes.

NodeJSmith:
## Summary

`tests/support/factories.py` carried three independent copies of the "closing spawn" task-bucket
test-double idiom — close any coroutine handed to `spawn()` so it is never reported as *never
awaited*, and hand back a mock stand-in for the task. This consolidates the two ad hoc copies onto
the canonical `make_closing_task_bucket()` factory.

The copies were not equivalent, which is what made the drift risk real:

| | `make_closing_task_bucket()` | the two inline copies |
|---|---|---|
| coroutine detection | `asyncio.iscoroutine(coro)` | `hasattr(coro, "close")` |
| returned task | one shared `MagicMock` | a fresh `Mock()` per call |
| `task.done()` | wired to `True` | not wired |
| `task.add_done_callback` | fires the callback synchronously | not wired |

A future fix to the coroutine-closing condition in one copy would have been easy to apply to only
one of the three.

## What changed

- `tests/support/factories.py` — `make_scheduler()` now builds `mock_service.task_bucket` from
  `make_closing_task_bucket()` instead of its inline `_default_spawn` closure. The comment
  explaining *why* the bucket must close coroutines (an `AsyncMock` `mark_job_removed()` produces a
  real coroutine object that an unconfigured mock never closes, leaking a "coroutine was never
  awaited" warning into some later, unrelated test's GC) is retained at the call site.
- `tests/unit/core/_fixtures_bus_scheduler.py` — `make_scheduler_service()`'s inline `_spawn`
  closure replaced with `make_closing_task_bucket()`. Its `make_async_adapter` override is still
  applied on top, since the shared factory returns a `MagicMock`.

Net −13 lines. `tests/support/helpers.py`'s `make_task_bucket()` is deliberately left alone —
real-task spawning is the documented contrast to the closing stub.

This is a straight swap rather than a behavior-preserving parameterization: no consumer of
`make_scheduler` (11 files) or `make_scheduler_service` depended on the divergent behavior. Only
`tests/unit/test_scheduler_resource.py` touches `task_bucket.spawn` at all, and it fully overrides
`spawn` itself rather than relying on the factory default. `asyncio.iscoroutine()` still correctly
detects the `AsyncMock`-produced coroutine that motivated the original `hasattr` check, so the
closing behavior is preserved.

## Testing

Test-only change; no src, docs, or frontend surface touched. CI runs the full suite across every
supported Python version — locally I ran the scoped set covering both changed factories and all
their consumers:

- `prek -a` — all Python hooks pass (ruff, ruff format, house-lint, the test-factory shadowing
  check, and the rest). The three frontend hooks (eslint, stylelint, tsc) fail only because
  worktrees don't share `frontend/node_modules`; no frontend file is touched by this branch.
- `prek pyright -a --stage pre-push` — clean.
- `uv run pytest tests/unit/scheduler/ tests/unit/core/ tests/unit/test_scheduler_job_names.py
  tests/unit/test_scheduler_resource.py tests/unit/test_source_tier_propagation.py
  tests/unit/test_forgotten_await_completeness.py --ignore=tests/unit/core/test_log_records_retention.py -n 4`
  — **1597 passed**. No "coroutine was never awaited" warnings, which is the regression this idiom
  exists to prevent.

`tests/unit/core/test_log_records_retention.py` is excluded above because it is flaky under xdist
independently of this branch: its `caplog.text` assertions come back empty depending on worker
placement. Verified against the default branch rather than assumed — two runs of the identical
broad command on a clean `main` checkout (at `5bd6c22`, this branch's merge-base) produced 6
failures and then 16 failures from that same file on unchanged code. The file passes 34/34 in
isolation both on `main` and on this branch. Pre-existing flakiness, not introduced here.

Closes #2180


NodeJSmith:
@coderabbitai review

NodeJSmith:
@coderabbitai review

PR #2339, opened 2026-09-21
refactor: read execution-table status colors from the shared STATUS_TONE_CLASSES palette

### LINKED ISSUE #2206: Consolidate execution-table status colors onto STATUS_TONE_CLASSES
## Description

Surfaced by the review battery on #2205 (integration-reviewer, HIGH). `frontend/src/utils/status.ts` already exports `STATUS_TONE_CLASSES: Record<StatusKind, string>`, documented as the "text-color utility for each StatusKind, for callers that tint plain text rather than render a badge or shape" — exactly what the execution table's status label does. Two other components already consume it (`shared/detail-stats.tsx`, `shared/stats-strip.tsx`).

`execution-table.tsx`'s `statusKindClass()` hand-rolls a third, parallel `StatusKind` → text-color mapping with the same CSS-variable values. This duplication predates #2205 (the old `statusLabelClass` had the same hardcoded colors); #2205 only made it more visible by narrowing the function to return exactly a color class. A future change to the status text palette has to be made in two places, and nothing fails if only one is updated.

This was deliberately **not** folded into #2205, because it is not rendering-neutral (see below) and that PR's contract — stated in its body and backed by its screenshots — is that the rendered class set is unchanged. Per `refactoring-discipline.md` ("No Smuggled Behavior Changes"), the behavior-changing consolidation belongs in its own PR.

## Decision (2026-09-14)

Adopt `STATUS_TONE_CLASSES`, including the `mute`/skipped rendering change. Muted grey for a skipped execution is more semantically correct than inheriting normal foreground — skipped isn't a normal-weight outcome — and it's a small enough visual delta to confirm with a single screenshot.

## The one real divergence

The two mappings agree on `ok`, `warn`, `err`, and `cancel`. They differ on `mute`:

| kind | `STATUS_TONE_CLASSES` | `statusKindClass` (execution table) |
|---|---|---|
| `mute` | `text-muted-foreground` | `undefined` (inherits the cell's default foreground) |

`mute` is the kind for `skipped` executions (`EXECUTION_STATUS_KIND` in `status.ts`). Adopting `STATUS_TONE_CLASSES` renders "skipped" labels in muted grey instead of normal foreground — accepted per the decision above.

`err` also carries `truncate` alongside its color, which is a layout concern rather than a tone concern and has to be composed separately.

## Implementation shape

```ts
// execution-table.tsx
const STATUS_TRUNCATE: Partial<Record<StatusKind, string>> = { err: "truncate" };

<span className={cn(MONO_TEXT_CLASS, STATUS_TONE_CLASSES[statusKind], STATUS_TRUNCATE[statusKind])}>
```

Dropping the `switch` is safe here: `STATUS_TONE_CLASSES` is typed `Record<StatusKind, string>`, so exhaustiveness over `StatusKind` is enforced by the type rather than by `@typescript-eslint/switch-exhaustiveness-check`.

## Acceptance Criteria

- [ ] `execution-table.tsx` no longer hardcodes `StatusKind` → text-color class strings; it reads `STATUS_TONE_CLASSES`.
- [ ] The `truncate` behaviour on `err` is preserved.
- [ ] Skipped-execution rows render `text-muted-foreground` (the new, accepted behavior). Include a screenshot in the PR showing a skipped-execution row.
- [ ] `src/components/shared/execution-table.test.tsx` passes; add a case covering the `skipped`/`mute` label classes if none exists.

NodeJSmith:
## Summary

`execution-table.tsx` hand-rolled a `StatusKind -> text-color` mapping in a local `statusKindClass()` switch, duplicating the CSS-variable values already exported as `STATUS_TONE_CLASSES` in `utils/status.ts` (also consumed by `detail-stats.tsx` and `stats-strip.tsx`). It was the third copy of the same palette, and a color change had to be made in two places with nothing failing if only one was updated.

This drops the switch and reads `STATUS_TONE_CLASSES` directly. Exhaustiveness over `StatusKind` is still enforced — by the `Record<StatusKind, string>` type rather than by the eslint switch-exhaustiveness check.

The `truncate` class that was bundled onto `err` is a layout concern, not a tone concern, so it is composed alongside the tone class at the call site rather than being smuggled into the shared palette.

## Not rendering-neutral (intended)

One visual delta ships with this: `mute` (skipped executions) previously returned `undefined` and inherited the cell's default foreground. It now renders `text-muted-foreground`. That change is deliberate — a skipped execution is not a normal-weight outcome — and is precisely why this was kept out of the rendering-neutral cleanup that preceded it. Every other kind (`ok`, `warn`, `err`, `cancel`) maps to a byte-identical class string, so those rows are unchanged.

## Review

Reviewed by the full battery (code, integration, WTF, llm-checker, lazy-checker, nitpicker).

- Three reviewers independently flagged the original `STATUS_TRUNCATE: Partial<Record<StatusKind, string>>` as ceremony: with a single populated key, the `Partial` type makes the exhaustiveness check inert, so the map bought nothing over a direct comparison. Collapsed to `statusKind === "err" && "truncate"` at the composition site. Rendered classes are unchanged and the existing tests (which assert on rendered class names) still pass.
- Integration review surfaced a pre-existing fourth copy of the same palette in `shared/error-display.tsx`, verified unchanged against `origin/main` and out of scope for #2206. Filed as #2340.

## Testing

- `prek -a` — all hooks pass (ruff, eslint, stylelint, prettier, tsc, and the frontend CSS/breakpoint guards).
- Scoped vitest on the changed component and the shared constant's other consumers:
  - `src/components/shared/execution-table.test.tsx` — 26 passed, including two new assertions that the skipped label carries `STATUS_TONE_CLASSES.mute` without `truncate`, and the failed label carries `STATUS_TONE_CLASSES.err` with it.
  - `src/utils/status.test.ts` and `src/components/shared/detail-stats.test.tsx` — 23 passed, confirming the sibling consumers are unaffected.
- CI runs the full backend + system + frontend suite across every supported Python version.
- Verified against the live demo stack in both themes: skipped rows on `predicate_demo`'s `motion_zone_check` render muted grey, and failed rows on `demo_stimulator`'s `sensor_health_check` still render destructive red with truncation.

## Screenshots

Evidence lives under `docs/pr-evidence/` and is excluded from the built docs site (`mkdocs.yml`).

Skipped executions, light theme — the muted-grey delta this PR introduces:

Skipped executions, light theme

Skipped executions, dark theme:

Skipped executions, dark theme

Failed executions, light theme — unchanged destructive red, truncation intact:

Failed executions, light theme

Closes #2206


NodeJSmith:
@coderabbitai review

PR #2341, opened 2026-09-21
test: extract shared WS envelope validation helper in test_ws_models

### LINKED ISSUE #2220: Extract the repeated WS envelope build-and-validate scaffolding in test_ws_models.py
Surfaced by the `duplicate-code` CI job (PMD CPD via `tools/check_duplicate_code.py`). Not introduced by any in-flight branch.

## What the shared shape is

Every discrimination test in `tests/unit/test_ws_models.py` does the same three things by hand:

1. Builds a raw WS envelope dict — a `"type"` discriminator, a `"data"` payload, and the same hardcoded `"timestamp": 1234567890.0` literal.
2. Feeds it through the class-level `TypeAdapter(WsServerMessage)` with `self.adapter.validate_python(raw)`.
3. Asserts the union narrowed to the expected concrete message class via `assert isinstance(msg, <X>WsMessage)`, and (for the batch cases) that `len(msg.data)` matches.

Steps 2 and 3 plus the envelope's trailing `"timestamp"` line are byte-for-byte identical across the file; only the payload body and the expected class differ. The `1234567890.0` timestamp literal alone appears 11 times, and `TypeAdapter(WsServerMessage)` is re-declared as a class attribute in two separate test classes.

## Decision

Extract the envelope-build-and-validate step into one module-level helper — `validate_envelope(msg_type, data)` returning the narrowed message — backed by a single module-level timestamp constant and a single shared `TypeAdapter(WsServerMessage)`. Payload dicts stay as inline literals (they are the thing under test). This matches the direction already used for equivalent test-scaffolding extractions elsewhere in this backlog; the counter-argument about protocol tests restating the wire format applies to the payload bodies, which this extraction deliberately leaves inline, not to the wrapper mechanics being removed here.

## Where

PMD flagged these three as one cluster:

- `tests/unit/test_ws_models.py:173-178`
- `tests/unit/test_ws_models.py:198-203`
- `tests/unit/test_ws_models.py:237-242`

Reading the file turned up the same shape at six more call sites beyond PMD's sample, all in scope:

- `tests/unit/test_ws_models.py:95-103` (`test_app_status_changed`)
- `tests/unit/test_ws_models.py:105-121` (`test_log_message`)
- `tests/unit/test_ws_models.py:123-132` (`test_connected`)
- `tests/unit/test_ws_models.py:134-138` (`test_connectivity`)
- `tests/unit/test_ws_models.py:140-147` (`test_service_status`)
- `tests/unit/test_ws_models.py:208-213` (`test_execution_completed_empty_batch_valid`)

`test_invalid_type_raises` (`:149-152`) builds the same envelope shape but asserts a raised `ValueError` instead of narrowing — worth covering by the same helper only if that reads cleanly; forcing it through is not required.

## Acceptance criteria

- [ ] The `1234567890.0` timestamp literal appears once in the file as a named module-level constant.
- [ ] `TypeAdapter(WsServerMessage)` is constructed once rather than per test class.
- [ ] The envelope-build + `validate_python` + narrowing-assert sequence is expressed through one shared helper (`validate_envelope(msg_type, data)`) at every call site with that shape (payload dicts remain inline literals).
- [ ] `uv run pytest tests/unit/test_ws_models.py` passes.

NodeJSmith:
## Summary

Removes the repeated WS envelope build-and-validate scaffolding from `tests/unit/test_ws_models.py`, surfaced by the `duplicate-code` CI job (PMD CPD). Pre-existing duplication — not introduced by any in-flight branch.

Every discrimination test in this file hand-built the same three-step sequence: a raw envelope dict (`type` / `data` / the same hardcoded `timestamp` literal), a `self.adapter.validate_python(raw)` call, and a narrowing `isinstance` assert. Only the payload body and the expected message class actually varied.

## What changed

- Added a module-level `validate_envelope(msg_type, data)` helper that builds the envelope, validates it, and returns the message the discriminated union narrowed it to.
- Hoisted the `1234567890.0` timestamp to a single `TEST_TIMESTAMP` constant — it appeared 11 times before.
- Hoisted `TypeAdapter(WsServerMessage)` to one module-level `MESSAGE_ADAPTER`; it had been re-declared as a class attribute in two separate test classes.
- Applied the helper at all 11 call sites with that shape, including `test_invalid_type_raises`, which reads cleanly through it despite asserting a raise rather than narrowing.

Payload dicts stay inline as literals. That is deliberate: they are the thing under test, and the counter-argument about protocol tests restating the wire format applies to the payload bodies, not to the wrapper mechanics removed here.

Net: 35 insertions, 54 deletions in a single test file. No source or behavior changes.

## Testing

- `uv run pytest tests/unit/test_ws_models.py` — 25 passed. Test-method count is unchanged at 21 before and after, so the refactor dropped no coverage.
- `prek -a` — all Python hooks pass (ruff check, ruff format, flake8-async, typos, house-lint, and the repo's custom test/import-boundary checks, including `Check test factory shadowing`).
- The three frontend hooks (eslint, stylelint, tsc) failed locally with package-resolution errors because this worktree has no `node_modules` installed; this branch touches zero frontend files. CI runs the full backend, system, and frontend suites across every supported Python version.

Closes #2220


NodeJSmith:
@coderabbitai review

